PR #47, opened 2022-07-13
Add AMBER FF. Remove redundant files.

SSchott:
Hey Brady,
Thanks for the cool addon. Seems very functional :D. I have no clue about Serpent, or if we can even mod the addon structure itself from our side, but I am guessing you are using the dicts in import_structure.py and the resid to make the per residue name selection node. I added some Amber residue names that could get handy for rendering trajectories. I also added type names, though it seems you are not using them right now? Should aa_type_no be just a dict of aa_type? I also took the liberty to delete files that, given that you are tracking with tags, I guess shouldn't be included in the repo.
Cheers,


BradyAJohnston:
Hello! Thanks for the contribution. Serpens has made the initial prototyping phase certainly a lot simpler and reduce the amount of boilerplate code that I have to write, but it make it harder for other people to contribute so I need to at some point convert it all to a more pure-code implementation.

You are right about the previous files that can be removed, I was keeping them around out of an abundance of caution but that's the whole point of git anyway!

The AA type names etc I haven't implemented yet (and hadn't properly assigned, so thanks for doing that!) but I will look into potentially using them as well.

Thanks for the amber residues also, I will have a poke around and make sure nothing is breaking unexpectedly and can merge it after that. Need to sort out a simpler solution for this whole process :)

BradyAJohnston (approved):
Thanks for adding the definitions! I will maybe put something together about how others can input other definitions to add support as well.

SSchott:
cool 😄 there could be defs generated according to residues that are in the PDB but not previously defined in the script. Dunno if the addon can generate such a flexible node.

Issue #109, opened 2022-12-13
Issues in `assembly.py`


padix-key:
Hi, I had a look at the code for building biological assemblies. Since I just implemented reading assemblies from PDB and MMTF myself in *Biotite* (see https://github.com/biotite-dev/biotite/pull/450 for reference), I think I spotted a few errors in `assembly.py` that I would like to report here.

1. As far as I understand the code, it does not parse the chains the transformations should be applied on. In my experience this is not an issue for most structures, but some structures use one set of transformations for some chains and another set of transformations for other chains (example: `5ZNG`).

2. The PDB parser parses assembly symmetry operations from `SMTRY` records in [Remark 290](https://www.wwpdb.org/documentation/file-format-content/format32/remarks1.html#REMARK%20290).
However this remark contains transformations to recreate atoms in the crystallographic unit cell, which do not necessarily represent the functional unit of the biomolecule. The transformations to create the assembly are stored in [REMARK 350](https://www.wwpdb.org/documentation/file-format-content/format33/remarks2.html#REMARK%20350) instead.

4. The PDBx parser assumes that each entry in the `pdbx_struct_oper_list` category gives the transformation for one copy in the assembly, as it is the case for PDB and MMTF. Unfortunately, the behavior for PDBx files is somewhat different: Which of these transformations must be applied is defined in the "oper_expression" category. Although for most structures this makes no difference in my experience, there are exceptions: Some assemblies may omit certain transformations, while other assemblies require even a Cartesian product of multiple transformations.

As far as I understand, the reason `assembly.py` is parsing the transformations instead of using `get_assembly()` from *Biotite* directly is for performance reasons in the rendering process, correct? If you specify what kind of information is required, I could contribute a PR by copy-pasting and adapting the code I wrote for *Biotite* for building assemblies.

I think the current `transform_dict` that the functions in `assembly.py` return are not able solve point 1 form the list above. So maybe another return value is required here? Alternatively, one could just say that cases, where point 1 applies are simply not covered.

BradyAJohnston:
Hi mate, thanks for looking into it! The assemblies were high on my list of things to rework as I had only managed to get it reliably working for when importing from the RCSB via `mmtf` format, and yes even then it currently lumps all chains into the same transformations and only does it for the first biological assembly in the file rather than supporting later ones (laziness on my part when getting it set up).

I likely think the whole approach needs an overhaul, and would be very excited to have contributions on it!

[...]

PR #121, opened 2022-12-21
Biological Assembly from mmCIF

corredD:
Attempt for a General Support for mmCIF biological assembly. Some code extract from ePMV to create the instance mesh.
Handle cellpack cif files. I dont know if there is a performance difference between geometrynode and Instancing on Faces using the regular interface. 
Two examples : 2PLV and a cellpack model.
image
image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       max-parallel: 4
       fail-fast: false
       matrix:
-        blender-version: ["3.5.0"]
+        blender-version: ["3.5.0", "3.6.0"]
         os: [ubuntu-latest, windows-latest]
         # os: [ubuntu-latest]
     steps:
@@ -24,7 +24,7 @@ jobs:
         with:
           python-version: '3.10'
           cache: 'pip'
-      # - run: pip install bpy==${{ matrix.blender-version }}
+      - run: pip install bpy==${{ matrix.blender-version }}
       # - run: pip install -r requirements.txt
       - name: Install MolecularNodes + Dependencies
         run: pip install .
--- a/MolecularNodes/assembly/cif.py
+++ b/MolecularNodes/assembly/cif.py
@@ -66,9 +66,9 @@ class CIFAssemblyParser(AssemblyParser):
                         rotations, translations
                     )
                     transformations.append((
-                        np.array(affected_chain_ids, dtype="U4"),
-                        total_rotation,
-                        total_translation
+                        np.array(affected_chain_ids, dtype="U4").tolist(),
+                        total_rotation.tolist(),
+                        total_translation.tolist()
                     ))
         
         return transformations
--- a/MolecularNodes/assembly/mmtf.py
+++ b/MolecularNodes/assembly/mmtf.py
@@ -41,7 +41,7 @@ class MMTFAssemblyParser(AssemblyParser):
             chain_ids = np.array(self._file["chainNameList"], dtype="U4")
             affected_chain_ids = chain_ids[transform["chainIndexList"]]
             transformations.append((
-                affected_chain_ids,
+                affected_chain_ids.tolist(),
                 matrix[:3, :3].tolist(),
                 matrix[:3, 3].tolist()
             ))
--- a/MolecularNodes/assembly/pdb.py
+++ b/MolecularNodes/assembly/pdb.py
@@ -81,9 +81,9 @@ class PDBAssemblyParser(AssemblyParser):
             )
             for rotation, translation in zip(rotations, translations):
                 transformations.append((
-                    np.array(affected_chain_ids, dtype="U4"),
-                    rotation,
-                    translation
+                    np.array(affected_chain_ids, dtype="U4").tolist(),
+                    rotation.tolist(),
+                    translation.tolist()
                 ))
 
         return transformations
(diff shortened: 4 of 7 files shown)


corredD:
TODO: maybe change the quad mesh to triangle.

BradyAJohnston:
Thanks so much for wanting to work on and improve the project! I'm excited to have you contributing. 

I won't have time to look through in detail until after Christmas, but I downloaded the fork and played around with it with some success.

## How to Handle Assemblies
This is something I have been unsure of how to handle in the past. Whether to create a new object that will become the biological assembly or change the current object (which is the asymettric unit) and then the object now becomes the biological assembly.

Certainly what I want to try and do, is keep everything inside of Geometry Nodes as much as possible. I'm unsure if there are particular performance improvements instancing inside of GN instead of on meshes. Regardless, Geometry Nodes is the current development focus of the Blender foundation and so it will be getting the biggest development focus and likely become more performant if it isn't already. 

Having the instancing occur within Geometry Nodes also enables easier manipulation of instances & geometry all without having to move between different methods of GN & regular instancing / mesh manipulation.

To handle biological assemblies for mmtf structures, I have been creating lots of nodes, each of which output a point that contains rotation and translation information. I then use those points to instance the required geometry. My implementation is lacking, as I haven't properly got it set up to handle symmetries which operate only on particular selections of chains (discussed in #109). I'm also not a huge fan of creating dozens of nodes, each of which output a single point, it seems kind of messy. @padix-key I believe is working on ensuring a common set of transformations that can be extracted via Biotite, which should also streamline biological assembly creation hopefully.

Potentially what could work is a combination of your current implementation and the current MolecularNodes implementation.

Creating an object which contains all of the information for instancing, either in the faces as you have currently done or just with points. We could use the `add_attribute()` to add the required information for chain selection, rotation matrices etc inside of attributes that are accessible inside of Geometry Nodes.
https://github.com/BradyAJohnston/MolecularNodes/blob/1c550d834734db620323eea38c0c65b6d1d550e8/MolecularNodes/load.py#L142-L151

There could then be an object which contains the transformation information for different biological assemblies, and on that object, using a Geometry Nodes node tree, we can instance the biological assembly, or this could be done inside the node tree of the original object as well.

## Parsing CellPack Files
Are you able to give more information around the parsing of CellPack files? I couldn't immediately figure out what changes you have done that enables the parsing of the files. I have a function which 'fixes' the `.cif` file so that it can be ready by Biotite, but being able to do this without having to read, write, then read the file again would be great.
https://github.com/BradyAJohnston/MolecularNodes/blob/1c550d834734db620323eea38c0c65b6d1d550e8/MolecularNodes/cellpack.py#L10-L28

Again, thanks for PR so far and excited to have your contributions, so would love your feedback on design principles and how this should best be approached.

padix-key:
> @padix-key I believe is working on ensuring a common set of transformations that can be extracted via Biotite, which should also streamline biological assembly creation hopefully.

Yes, I am still working on it. Since I am currently also occupied with other stuff, I will probably still need some weeks until I create a PR.

corredD:
Thanks for you comments Brady, and thanks for this great plugin.

I will try to continue to look into it in my spare time !

> Potentially what could work is a combination of your current implementation and the current MolecularNodes implementation.
> Creating an object which contains all of the information for instancing, either in the faces as you have currently done or just with points. We could use the `add_attribute()` to add the required information for chain selection, rotation matrices etc inside of attributes that are accessible inside of Geometry Nodes.
Can the attribute be a mesh object ? I would see the BU node as the "Distribute Points on Face" node but instead of using rand om or poisson disk placement it use the same approach as DupliFace. The other input would be the BU index and a selection.
 
> ## Parsing CellPack Files
> 
> Are you able to give more information around the parsing of CellPack files? I couldn't immediately figure out what changes you have done that enables the parsing of the files. I have a function which 'fixes' the `.cif` file so that it can be ready by Biotite, but being able to do this without having to read, write, then read the file again would be great.

I didn't change the cellpack files, only changed the biological unit interpretation. The problem was in the difference between the label_asym_id and auth_asym_id. label_asym_id is used in the transformation operator not the auth_asym_id which was used by default to build the molecular node. By toggling of the use_author_fields biotite will use the label_asym_id  as chain_id. Also biotite currently has a parsing error you can reproduce by trying to parse the cif file for 2PLV the operation with pdbx.convert._parse_operation_expression which I think correspond to this line : 
6 '(X0)(1-5,11-15,21-30,41-50)' A,B,C,D,E,F,G,H,I,J 

`
mol  = pdbx.get_structure(file, extra_fields = ['b_factor', 'charge'], use_author_fields=False)

def get_transformations_pdbx(file_pdbx):
    import biotite.structure.io.pdbx as pdbx
    assembly_gen_category = file_pdbx.get_category(
        "pdbx_struct_assembly_gen", expect_looped=True
    )
    struct_oper_category = file_pdbx.get_category(
        "pdbx_struct_oper_list", expect_looped=True
    )
    transformations = pdbx.convert._get_transformations(struct_oper_category)
    transform_dict = {}
    for aid, op_expr, asym_id_expr in zip(
                    assembly_gen_category["assembly_id"],
                    assembly_gen_category["oper_expression"],
                    assembly_gen_category["asym_id_list"]):
        # skip until biotite fix the issue
        if len(op_expr.split("-")) > 2:
            continue
        operations = pdbx.convert._parse_operation_expression(op_expr)
        asym_ids = asym_id_expr.split(",")
        if aid not in transform_dict:
            transform_dict[aid] = {}
        for asym in asym_ids:
            if asym not in transform_dict[aid]:
                transform_dict[aid][asym] = []
            for i, operation in enumerate(operations):
                for op_step in operation:
                    transform_dict[aid][asym].append(transformations[op_step])
    return transform_dict
`

PR #132, opened 2023-01-03
include_bonds was ignored

DrDomenicoMarson:
Hi! Thanks again for the great work!
Just a quick fix as the include_bonds UI option was not honored when loading an MD trajectory!

### DIFF
--- a/MolecularNodes/ui.py
+++ b/MolecularNodes/ui.py
@@ -94,7 +94,7 @@ class MOL_OT_Import_Protein_MD(bpy.types.Operator):
         md_step =  bpy.context.scene.mol_import_md_frame_step
         md_end =   bpy.context.scene.mol_import_md_frame_end
         del_solvent = bpy.context.scene.mol_import_del_solvent
-        
+        include_bonds = bpy.context.scene.mol_import_include_bonds
         
         mol_object, coll_frames = md.load_trajectory(
             file_top    = file_top, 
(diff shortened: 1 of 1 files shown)


DrDomenicoMarson:
Just a quick note also:

when importing a protein of about 270 aa, with solvent, on my system (32GB ram, i9, and RTX3080) I experienced crashes if I tried to visualize all the atoms in balls-and-stick representation. 

To avoid crashes I just load the trajectory as atoms, then I selected just the protein residues before switching to a balls-and-stick representation. Strangely enough, sometimes If I re-visualize everything (remove the selection) in balls and sticks after having visualized only the protein, it works without problems, I don't know why!

BradyAJohnston (approved):

BradyAJohnston:
Interesting that it crashed once but then didn't crash when already some of the scene had the ball-and-stick representation. 

In comparison to Atoms Cycles which uses an _extremely efficient_ point-cloud rendering system, Atoms EEVEE and Ball and Stick utilise instancing of geometry. Quite quickly, this can become problematic for scenes with lots of atoms. 270 aa should not have been an issue. May have been processing something else in the background and the spike in computation caused it to crash. 

If it becomes more common can you take note of what causes it and open an issue? 

Thanks for the pull request, and for the features and bugs you have picked up so far (with solutions provided)! I really appreciate it. And in the future feel free to open a pull request with fixes if you manage to figure them out already as well.

BradyAJohnston:
Hi @DrDomenicoMarson. Discussed in #109 but I accidentally deleted the license file for this project when I went through a refactor. It was originally licensed MIT. This pull request is the only contribution so far from someone other than myself from when the license was deleted. 

It leaves it ambiguous as to what happens with the license, but if you are happy with your contributions being released under the MIT license? If so I'll  add it back in to the project.

DrDomenicoMarson:
Hi! Yeah, there is no problem for me to have this licensed as you think is appropriate!

Issue #63, opened 2022-08-17
Non-English Blender Breaks Node Setup on Import


BradyAJohnston:
When Blender is set to non-English languages, the initial node setup fails.

Documented cases: #62, #55, #41, #29. 

Closing all of those issues and documenting here.

YaoYinYing:
Hi Brady, 

FYI, I changed my language setting and inserted a debug line `raise ValueError(node_mod.node_group.nodes.keys())` in `/Users/yyy/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/nodes.py:115` and restart Blender for reloading this addon, then load a PDB and it reported the following:

```python
Python: Traceback (most recent call last):
  File "/Users/yyy/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/ui.py", line 41, in invoke
    return self.execute(context)
  File "/Users/yyy/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/ui.py", line 27, in execute
    mol_object = load.molecule_rcsb(
  File "/Users/yyy/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/load.py", line 27, in molecule_rcsb
    nodes.create_starting_node_tree(
  File "/Users/yyy/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/nodes.py", line 115, in create_starting_node_tree
    raise ValueError(node_mod.node_group.nodes.keys())
ValueError: ['组输入', '组输出']
```
it looks like blender has already translated the keys into simplified Chinese.

YaoYinYing:
I was thinking that if someone prefers to use their native language in Blender, we could launch a translation project to replace all keys in the MolecularNodes addon that have already been defined by Blender in different languages, and create a toggle option or automatically detect the language setting to select the correct translation.

BradyAJohnston:
Thanks for looking into this. 

Looking at the error, Blender is translating the names for the default nodes, but I am trying to request them via their English translation keys. 

https://github.com/BradyAJohnston/MolecularNodes/blob/9e77b1e304d9bb4210254c50a1167faa24625bcc/MolecularNodes/nodes.py#L115-L118

This should only be a problem for the default nodes potentially. Having to create our own translation of each of the nodes would potentially be a lot of work. What might be a better approach is to find a different way to reference these particular nodes, like the underlying RNA values of the nodes that shouldn't change regardless of the language.

[...]

PR #139, opened 2023-01-16
Support non-English setting in Blender #63 via Blender build-in API

YaoYinYing:
The rest of texts stay untouched.

### DIFF
--- a/MolecularNodes/nodes.py
+++ b/MolecularNodes/nodes.py
@@ -112,9 +112,9 @@ def create_starting_node_tree(obj, coll_frames, starting_style = "atoms"):
     #     mol_append_node(node_group)
     
     # move the input and output nodes for the group
-    node_input = node_mod.node_group.nodes['Group Input']
+    node_input = node_mod.node_group.nodes[bpy.app.translations.pgettext("Group Input",)]
     node_input.location = [0, 0]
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
Amazing! Thanks so much for figuring this out. I'll run some tests this evening but I can't see anything being a problem!

BradyAJohnston:
I've done some testing by changing the language to other languages (I tested English, French and Japanese) and the non-english language settings still didn't set up the node groups correctly.

Is the changing of languages working for you?

BradyAJohnston:
Ah OK, in testing it some more, I had to change the language, then restart Blender, and now the fix works.

BradyAJohnston:
If you could just add the changes to these lines also (the only other usage of the 'Group Input') then I'll merge.

https://github.com/YaoYinYing/MolecularNodes/blob/20cb63c9a710827a28387df68ef7726bf537ab0a/MolecularNodes/assembly.py#L122-L123

YaoYinYing:
Sure! I forgot to replace all these keys in the entire repo. Thanks for your kind reminding! I think this should be fully working now. 

Just let me know if something weird occurs.

BradyAJohnston (approved):

BradyAJohnston:
All looks great. I really appreciate you looking into this and contributing the PR!

Issue #146, opened 2023-01-29
non-english language issue in 2.3.0


Mangkun-Chen:
![image](https://user-images.githubusercontent.com/108992267/215299976-588475d8-915a-435b-98b3-f3e5ddcad085.png)
![image](https://user-images.githubusercontent.com/108992267/215300013-62697a2d-0037-49a5-a6a0-e34fdd8c7664.png)

Hi Brady

I installed version 2.3.0. 
Problems still exist in the Chinese environment. Can you invite @ YaoYinYing to view this issue？I think he should be Chinese.

Mangkun-Chen:
@YaoYinYing

YaoYinYing:
@Mangkun-Chen did you ever restart Blender after switching language setting? This is actually essential for MN to apply a complete reloading.

YaoYinYing:
okay maybe that is the point. can you make the “新建数据(New Data)” checked and take another try?
<img width="286" alt="Snipaste_2023-01-29_12-45-14" src="https://user-images.githubusercontent.com/33014714/215305541-4a628431-7542-47d7-978b-10ba9400be26.png">

Mangkun-Chen:
![image](https://user-images.githubusercontent.com/108992267/215306096-0e701899-2a80-4383-9a1b-5a8bdf5da618.png)
@YaoYinYing I restart Blender. and open the “新建数据(New Data)”.  it is still reported error

Mangkun-Chen:
![image](https://user-images.githubusercontent.com/108992267/215306186-dd81f492-a329-4107-8c11-13c1591bf7e0.png)

YaoYinYing:
This looks wired. I cannot reproduce this error on either my macbook or retired windows laptop. Can you try to: 

1. change Language to Chinese,
2. keep this option of new data translation checked,
3. do a restart of Blender,
4. and finally load the PDB?

Mangkun-Chen:
@YaoYinYing it works now. 
不过很多插件开启“新建数据”会有bug，我一般关闭

YaoYinYing:
@Mangkun-Chen Thanks for reporting this issue. I have carefully read the official documents and submit a new PR #147 . It should be fixed soon.

BradyAJohnston:
Thanks for the bug report @Mangkun-Chen and for the fix @YaoYinYing, merged into master!

Mangkun-Chen:
Thank you！

PR #147, opened 2023-01-30
Fix non-English error when New Data translation is disabled.

YaoYinYing:
Blender has provided a series of translation APIs that allow addon designers and users to translate the texts on UI, Tools, Tips and data. 

As @Mangkun-Chen metioned in #146 , Blender does not recommend using the "New Data" translation option (Edit-Preferences-Interface-Translation-Affect-New Data) as it may cause conflicts with many addons. 

So let me just make it clear: 
- `bpy.app.translations.pgettext()` always performs the translation, regardless of translation options, which should be rarely used, 
- `bpy.app.translations.pgettext_tip()` translates the tool and tip messages only if `Tooltips` is checked,
- `bpy.app.translations.pgettext_iface()` handles the main user interface and is controlled by the `Interface` option,
- `bpy.app.translations.pgettext_data()` manages the "New Data" and is controlled by the "New Data" option.

In this case, the `Group Input`/`Group Output` should be always handled by `pgettext_data()` and the translation is determined by the user.


BradyAJohnston:
Thanks again for looking into this and fixing it. Really appreciate the help!

PR #148, opened 2023-01-30
Reading biological assemblies from structure files

padix-key:
This PR delivers the groundwork for reading the transformations (rotation/translation) necessary for building biological assemblies from structure file, as discussed in #109.
The implemented API is slightly different from the one proposed in #109.

The PR adds the `assembly` subpackage to `MolecularNodes` containing

- The abstract `AssemblyParser` base class, that is able to list available biological assemblies for a *Biotite* file object and parse the transformations for a given assembly ID
- Concrete subclasses of `AssemblyParser` for each supported file format:
  - [x] PDB
  - [x] PDBx/mmCIF
  - [x] MMTF

Furthermore, a `tests` directory was created for running tests using `pytest`. The `MolecularNodes` package is automatically added to the `PYTHONPATH` via preparations in `tests/conftest.py`.


padix-key:
At the moment the PR only supports assembly parsing from MMTF. @BradyAJohnston is the proposed API and directory structure appropriate? If this is the case I would proceed implementing assembly parsing for PDB and PDBx.

BradyAJohnston:
Hi @padix-key thanks so much for opening up the PR, and for offering to contribute so much to the project I really appreciate it. 

From looking at the code so far, the API should work well. I shouldn't have any trouble with integrating that on the Blender side of things.

For the module structure, you certainly have more development experience with this kind of thing and how you think it should be structured. So far I've kept it a relatively flat structure for simplicity. I'm open to suggestions on the direction to structure the project in a more robust (and importantly testable) way. Having more MolecularNodes-specific classes might also be a way to go about this. I've thought about it, but I don't have enough experience with python classes to know.

padix-key:
I added now an `AssemblyParser` for PDB, CIF and MMTF, so I think I have finished my part of this feature. I would leave the appropriate import into Blender up to @BradyAJohnston. Of course I am willing to support this effort, if questions arise where I can help. Let me know what you think about my additions and whether you want to implement the remaining part inside this PR or in a new one.

Finally I would like to note a potential caveat here: CIF files provide two values for chain IDs: the `label_asym_id` and `auth_asym_id`. The former ones is the 'official' value in CIF files, that is also referred to by the output in the `CIFAssemblyParser`. The latter ones are consistent with the chain IDs from the PDB format.
`PDBxFile.get_structure()` parses by default the `auth_asym_id`. To be consistent with the chain IDs used by `CIFAssemblyParser` you need to give `use_author_fields=False` in `PDBxFile.get_structure()`.

Most of the time `label_asym_id` and `auth_asym_id` will be the same - at least for the macromolecular part of a structure. However, sometimes thay may differ.

BradyAJohnston:
Thanks @padix-key for putting this all together so comprehensively. I'm at a conference over the coming week & this week, so I won't have a chance to review in depth until afterwards. 

When I get a chance to review it all, I'll make the required changes to your PR and then we can merge just the one I think that will be best.

PR #149, opened 2023-01-31
Customized Residue ID selection

YaoYinYing:
Introducing a new and improved feature for custom residue id selection in MolecularNodes! With this update, selecting residues has never been easier. The residue selection string can now be inputted in a user-friendly format, combining dash-separated continuous residue IDs and comma-separated individual residue IDs.

For instance, to select residues 1 to 10, 20 to 25 and 30, simply input the string `1-10,20-25,30`. Happy selecting!

MN_residue_selection_input

### DIFF
--- a/MolecularNodes/ui.py
+++ b/MolecularNodes/ui.py
@@ -452,6 +452,15 @@ class MOL_OT_Assembly_Bio(bpy.types.Operator):
         
         return {"FINISHED"}
 
+def menu_residues_selection_custom(layout_function):
+    obj = bpy.context.view_layer.objects.active
+    label = 'Residue selection'
+    op = layout_function.operator(
+        'mol.residues_selection_custom', 
+        text = label, 
+        emboss = True, 
+        depress = True
+    )
 
 def menu_item_surface_custom(layout_function, label):
     op = layout_function.operator('mol.style_surface_custom', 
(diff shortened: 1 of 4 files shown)


MolecularNodes/ui.py
```diff
@@ -452,6 +452,15 @@ def execute(self, context):
         
         return {"FINISHED"}
 
+def menu_residues_selection_custom(layout_function):
+    obj = bpy.context.view_layer.objects.active
+    label = 'Residue selection'
```
BradyAJohnston, line 457: I think change to 'Res ID Custom' to match `Res ID` and `Res ID Range`.

BradyAJohnston (changes requested):
I think this is really really great! Thanks for putting it together so quickly. 

I like the menu / UI that pops up when you click the selection.

It would require a bit more code, but I think if a user inputs a range `10,20-100, 120-150`, then the parts where a range is used we should use the `MOL_sel_res_id_range` node rather than creating an input for each number. For larger ranges, the resulting node can get very large, as opposed to if it is constructed from using the range nodes.

image

Overall the PR looks great and it's ready to be merged with some minor changes.

MolecularNodes/ui.py
YaoYinYing: Yeah you are correct that the combination of `MOL_sel_res_id` and `MOL_sel_res_id_range` would be nicer for GN editing. That is why I consider it as a `phenotype`.  I will get it improved soon.

BTW, the input box is absolutely cool. If you want to make a programatic enhancement of MN, I believe that is a good way.

BradyAJohnston: Yep I'll definitely see how I can use it in the future!

YaoYinYing:
Okay here are the latest and exciting updates:
1. fix long but continuous `MOL_sel_res_id` with `MOL_sel_res_id_range`
2. allow fuzzy input string. e.g. `19;94+1:16` equals to `19,94,1-16`
3. hide `MOL_sel_res_id` and `MOL_sel_res_id_range` entries from menu.
4. Update comments
5. rename confusing variable `input_list` of OT function to `input_resid_string`

BradyAJohnston:
Thanks for doing all of this! It looks great. I'm at a conference over the next week so I won't have the chance to review it properly until then, but I don't see any immediate issues. Thanks again for all of your contributions so far :)

BradyAJohnston:
I've had a look through and a play around and it all looks good. 

I made some minor changes to refactor some of the names. I think it's also best to still have the `Res ID Single` and `Res ID Range` available to the user if they want them, but the `Res ID` now becomes the custom input that you have created, which also moves it to the top of the menu. 

If you are happy with the changes that I have made, then I'll merge into main :)

BradyAJohnston (approved):

BradyAJohnston (approved):

YaoYinYing (commented):
Sure! It looks great! I agree that it will be more confusing for one to select only one point or one range but get such a hierarchical selection node group. 

I think future cleanup/refactoring should get this improved while providing just one selecting entry. 

I'm glad you like this PR!

BradyAJohnston:
I'm hoping once Geometry Nodes adds support for string inputs this will be much cleaner.

PR #160, opened 2023-02-12
Attribute secondary structure of proteins

OLaprevote:
Only computed on the first model of a biotite molecular array. See https://twitter.com/PestoFasta/status/1613441786147639296.


BradyAJohnston:
Thanks for opening the PR! Exciting to soon have some secondary structure information inside of MN.

I think we should try and do this with the new `annotate_sse()` function that was introduced with Biotite [`0.36`](https://github.com/biotite-dev/biotite/releases). It will require me to bump the version number for biotite inside of MN, but Patrick has added a bunch of quality of life improvements to Biotite specifically for MN so we should bump the version and use the most up to date code going forward.

OLaprevote:
Damn it, I was making sure to use the most recent version one month ago, didn't see this new one.
That's great, though, it should definitely make the whole code cleaner! Will do!

BradyAJohnston:
Other than that, the implementation looks solid. Once we have access to the secondary structure information, then it's just a matter of building the required nodes for creating nice cartoons.

OLaprevote:
Took me a while and several merges from the main, but it's now much cleaner using biotite 3.6.
Would you rather that I re-fork MolecularNodes to make this in a single, cleaner commit (removing all the merges), or is it fine as is?

For the record: Had to hack my way through for updating biotite by importing MolecularNodes inside the blender scripting interface and run `MolecularNodes.pkg.install(pypi_mirror=bpy.context.scene.pypi_mirror)` manually. Hopefully this will change after PR #172 (or equivalent) to enable easy dependency update.

YaoYinYing:
Well you don't have to struggle with the installation stuff. That PR was initially inspired by this thread and hoped to be helpful to this DSS PR.

BradyAJohnston (approved):

BradyAJohnston:
Looks good! Yes I've had a couple of cracks at trying to make nice secondary structure cartoons, and it's quite difficult, but definitely something that is possible. 

In the meantime the secondary structure can certainly be used for width & colouring so I'll merge into main and deal with trying to build the cartoons later.

Issue #108, opened 2022-12-13
Apple silicon/ Blender 3.4.0 / MolNodes 2.0.2 package installation failure


PlethoraChutney:
Hi Brady! Excited to play with the new features, and maybe finally dip my toes into contributing!

Unfortunately, trying to install the required packages using this button in blender:

<img width="248" alt="image" src="https://user-images.githubusercontent.com/27385979/207387948-e2f0a682-32b7-4292-9241-c5069a5903fb.png">

results in an error message reading:

> Failed to install required packages. Please check /Users/posert/Library/Application Support/Blender/3.4/scripts/addons/MolecularNodes/logs/side-packages-install.log

The contents of that log file are:

```
$ cat ~/Library/Application\ Support/Blender/3.4/scripts/addons/MolecularNodes/logs/side-packages-install.log
-----------------------------------
Installer Started: 2022-12-13 08:18:36.297157
-----------------------------------
###################################
Installer finished: 2022-12-13 08:19:05.242502
###################################
```

# System Info
* M2 macbook air
* Monterey 12.6.1
* Blender 3.4.0
* MolNodes 2.0.2

# Steps already taken
* Uninstall molnodes, restart blender, re-install molnodes

BradyAJohnston:
How incredibly unhelpful of a log file haha. (also envoius of the M2). 

Can you try launching Blender via the terminal `blender` or `/Applications/Blender` should launch it. Then when you try to install the packages again it should hopefully give a more helpful error message printed to the terminal.

PlethoraChutney:
Let me tell ya, if I'd known how nice these machines are I would have "accidentally" destroyed my old Dell ages ago.

Ah, good call. It seems Blender is failing to build a wheel for `numpy` during the `biotite` install. I'll poke around a bit! Thanks!

BradyAJohnston:
Are you getting the same compiler error as #102 where it's lacking `gcc`?

PlethoraChutney:
No, I'm missing `python.h`. Trying to figure out why, since I should have it. I'm now just failing to install MDAnalysis, I got numpy working.

BradyAJohnston:
Okay this is a good test case. This seemed to be an issue with earlier versions of Blender when I was originally developing the addon which is why I had the complicated installation od MDAnalsysis in an external python and you had to then link it to Blender's python. 

It might be that it's the ARM build of Blender doesn't include the `python.h` file but it's now included in other versions?

[...]

Issue #111, opened 2022-12-17
Installation of Molecular Nodes 2.0.2 with Blender 3.4.0 on M1 Max Mac Ventura 13.0


gha2012:
Hi, 
I tried to install Molecular Nodes 2.0.2 with Blender 3.4.0 on M1 Max Mac Ventura 13.0. After installing the .zip file, there is a button "install packages" in the Molecular Nodes panel. When I click it, I see a beachball and after a while it says "Failed to install required packages. Please check log file: ..." 
That log file just contains this:
-----------------------------------
Installer Started: 2022-12-17 15:55:53.117289
-----------------------------------
###################################
Installer finished: 2022-12-17 15:56:24.477652
###################################

Any idea what might be the problem?

Thanks,
Gregor

PlethoraChutney:
See if the solution here helps you, it fixed the same issue for me!

https://github.com/BradyAJohnston/MolecularNodes/issues/108#issuecomment-1355965223

In short:

1. Install miniconda
2. `conda create -n molnodes python==3.10 biotite==0.35.0 mdanalysis==2.2.0 -c conda-forge -y`
3. `mv /Applications/Blender.app/Contents/Resources/3.4/python /Applications/Blender.app/Contents/Resources/3.4/old_python`
4. `ln -s /path/to/miniconda/python /Applications/Blender.app/Contents/Resources/3.4/python`

And it should work!

gha2012:
Great! That works! Thank you!

BradyAJohnston:
Thanks for the bug report mate! Glad it's now working. Definitely seems to be an ARM mac issue. Will need to make sure the log file is a bit more useful ha.

gha2012:
FYI: Before the fix, I started Blender from the terminal and clicked on the "install packages" button. There I could see that the installation of numpy failed. Below is an excerpt. Maybe that helps to fix it.

```
            SystemError: Cannot compile 'Python.h'. Perhaps you need to install python-dev|python-devel.
            [end of output]
      
        note: This error originates from a subprocess, and is likely not a problem with pip.
        ERROR: Failed building wheel for numpy
      Failed to build numpy
      ERROR: Could not build wheels for numpy, which is required to install pyproject.toml-based projects
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× pip subprocess to install build dependencies did not run successfully.
│ exit code: 1
╰─> See above for output.

[...]

PR #162, opened 2023-02-14
Refactor dependencies setup stuff

YaoYinYing:
This PR is mainly focused on improving dependencies stuff in the following ways:

1. PyPI mirror selection, which might be helpful for those get blocked from accessing the original PyPI repo.
2. Passing the `stderr` of `pip install` to Blender scripting window for detailed reminding.
3. Trying to provide a easier way to solve the `Python.h not found` error while installing MDA via `pip install` on Mac on Apple Silicon chips. This issue has been mentioned at https://github.com/BradyAJohnston/MolecularNodes/issues/108#issuecomment-1429384983 for a more elegant solution.
4. Getting `pip install` more compatible (for future dev).
5. Remove logging from dependencies installation. Just have no idea why we have to log some date &time.
6. Bring `requirements.txt` back to simplify both the `pip install` procedure and dependencies management.

### DIFF
--- a/MolecularNodes/pref.py
+++ b/MolecularNodes/pref.py
@@ -42,21 +42,53 @@ class MOL_OT_install_dependencies(bpy.types.Operator):
     
     def execute(self, context):
         if not pkg.available():
-            stderr=pkg.install(
+            import datetime,platform
+            
+            # generate logfile
+            logfile_path = os.path.abspath(MolecularNodesAddon.logpath + 'side-packages-install.log')
+            logfile = open(logfile_path, 'a')
+            
+            logfile.write("-----------------------------------" + '\n')
+            logfile.write("Installer Started: " + str(datetime.datetime.now()) + '\n')
+            logfile.write("-----------------------------------" + '\n')
+            install_commands,install_logs=pkg.install(
                 pypi_mirror=bpy.context.scene.pypi_mirror,
             )
+
+            # log cmd and stdout/stderr
+            for cmd,log in zip(install_commands,install_logs):
+                logfile.write(f'{cmd}\n{log}\n')
+
+                # solve this Apple Silicon installation failure 
+                if ("fatal error: 'Python.h' file not found" in log) and (platform.system()== "Darwin") and ('arm' in platform.machine()):
+                    error_msg = f"ERROR: Could not find the 'Python.h' header file in version of Python bundled with Blender.\n" \
+                            "This is a problem with the Apple Silicon versions of Blender.\n" \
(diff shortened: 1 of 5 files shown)


BradyAJohnston:
Thanks again for opening up another handy PR! I"ll look into this in more detail in a few days (just handling a couple of other PRs at the moment), but it all looks good so far.

YaoYinYing:
Happy to help with MN dev!

MolecularNodes/pref.py
```diff
@@ -43,25 +42,10 @@ class MOL_OT_install_dependencies(bpy.types.Operator):
...
             
-            # generate logfile
-            logfile_path = os.path.abspath(MolecularNodesAddon.logpath + 'side-packages-install.log')
-            logfile = open(logfile_path, 'a')
-            
-            logfile.write("-----------------------------------" + '\n')
-            logfile.write("Installer Started: " + str(datetime.datetime.now()) + '\n')
-            logfile.write("-----------------------------------" + '\n')
-            
-            pkg.install()
```
BradyAJohnston, line 56: This section included the date times, but also included the result of `pkg.install()` which wrote the terminal output to the log file. On mac you can't view the terminal without first starting Blender from the terminal, so a user can't see any proper error outputs if they haven't started from the command line. 

I think we should probably still be writing some kind of log file, especially as we add the requirements for more and more python packages to add more functionality, as it'll help with debugging in the future.

BradyAJohnston (changes requested):
This all looks great and is a bit cleaner, but I think we should write some kind of log file during package installation, to help with inevitable debugging down the road.

MolecularNodes/pref.py
YaoYinYing: Okay, that is a good point. But it looks like the error message is not correctly recorded, as reported in #108 and #111 for instances. I think there might be a proper solution by redirecting the stdin, stdout and stderr respectively to pkt_install() and writing them into logfile.

BradyAJohnston: yes this is true. The logging seemed to work on my machine but was problematic with others. I need someone with more python knowledge than me to potentially fix that. Maybe we can merge this PR regardless and then try and set up some logging in another.

YaoYinYing: I think it should be fixed now. for Windows: https://gist.github.com/YaoYinYing/5ac064ad2c6e642ab151583252e6cfbe
and for Mac on M1: https://gist.github.com/YaoYinYing/77d746c3cee28a7766125c9df9d2df09

also for the error of Python header, an user guide link will appear at the Blender console: 

image

BradyAJohnston (approved):

PR #168, opened 2023-02-23
Fix m1 mac dependency

YaoYinYing:
This PR is focus on fixing the dependency failure on M1 mac seamlessly.

**Xcode commandline tools are required**: `xcode-select --install`

example `logs/side-packages-install.log`: https://gist.github.com/YaoYinYing/005ccba9efbd05db97c7d19fca4cf694

### DIFF
--- a/MolecularNodes/__init__.py
+++ b/MolecularNodes/__init__.py
@@ -36,7 +36,7 @@ from .pkg import *
 def register():
     bpy.types.Scene.mirror = bpy.props.StringProperty(
         name = 'mirror', 
-        description = 'PyPI/Conda Mirror', 
+        description = 'PyPI Mirror', 
         options = {'TEXTEDIT_UPDATE'}, 
         default = 'Default', 
         subtype = 'NONE', 
--- a/MolecularNodes/__init__.py
+++ b/MolecularNodes/__init__.py
@@ -34,13 +34,13 @@ from .pkg import *
 
 
 def register():
-    bpy.types.Scene.pypi_mirror = bpy.props.StringProperty(
-        name = 'pypi_mirror', 
-        description = 'PyPI Mirror', 
+    bpy.types.Scene.mirror = bpy.props.StringProperty(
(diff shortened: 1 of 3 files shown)


YaoYinYing:
Hold on ... It can not work with one M2 MBA from my friend....

BradyAJohnston:
I've been moving house so I haven't been on top of anything. Looks like it will remain to be problematic and maybe a bit more complicated to troubleshoot

YaoYinYing:
Well, there's two ways for solving this issue by installing external python and build MDA with system-wide gcc provided by xcode. phenotype at here https://github.com/BradyAJohnston/MolecularNodes/issues/108#issuecomment-1429384983

1. Using a helper script (https://github.com/BradyAJohnston/MolecularNodes/issues/108#issuecomment-1442794000)
2. Embed a miniconda setup working flow in `pkg.py`. I have done this and will re-open this PR for your preview.

But again, Xcode commandline tools is required by `xcode-select --install`

BradyAJohnston:
I don't think that doing all of this under the hood without the user's knowledge is necessarily a good idea. Installing conda / miniconda and then setting up an environment and installing the required packages will take quite a bit of time, and users might just quit out of Blender of it freezes for too long. 

If this script is robust and works well for a large portion of users, then we can have it as part of the documentation as a fix for M1 / M2 users but I don't know if it should be part of the initial install script.

YaoYinYing:
That's true.
How about removing this conda stuff and just keep the external Python searching?

For M1/M2 users, the guidelines looks like:
1. Install Xcode command line tools
2. Install Python3.10(whether it is from brew or an activated conda env)
3. Install Blender and launch it for the first time from the terminal
4. Install MN and click this 'Install dependencies' button 

This should work as the same way.

YaoYinYing:
And a future dev problem just came to me yesterday:
If you have pinned a higher version of any one of the package in requirements.txt, how should this button appear again in front of the user after they Install MN of a new release?

PR #172, opened 2023-03-07
Enable dependency updates after MN is upgraded.

YaoYinYing:
- refactor `run_pip` to remove redundancies and get run results fully recorded.
- reformat `side-packages-install.log` to make it looks cleaner.
- After install a newer MN, the versions of dependencies pinned at `requirements.txt` is checked against installed ones, if the version number of installed failed to match the pinned, the installation buttom will re-apeared. 
- require a restart of Blender after upgrading dependencies .
- minor fix: MN crashes with undefined `error_msg` while non-M1 error occurs.

### DIFF
--- a/MolecularNodes/pkg.py
+++ b/MolecularNodes/pkg.py
@@ -3,6 +3,7 @@ import sys
 import os
 import site
 from importlib.metadata import version
+from importlib import reload
 import bpy
 import pathlib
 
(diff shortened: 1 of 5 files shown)


YaoYinYing:
One thing that should be noted is running Blender with VSC will not work for this pip updates and I have no idea what is happening. In other words, MN must be installed as a zipped addon in Blender setting.

YaoYinYing:
There lies a tiny issue of  repetative dependency detection at 
https://github.com/BradyAJohnston/MolecularNodes/blob/eef1defc88826cfaea2285057f33b781a01f6b5a/MolecularNodes/ui.py#L295-L297

When I was working on this PR today using 
https://github.com/YaoYinYing/MolecularNodes/blob/5d5124f8df5fef440f55d92e5730ba96334d63dc/MolecularNodes/pkg.py#L78 for package version comparison tests in VSC console, this line was excuted everytime when my cursor was moving  upon the MN sidebar, meaning that Blender may use a sort of looping strategy for UI interaction. 

I am not so sure how this endless detecting affects the global performance but  it may not be so important, since the sidebar is mainly designed for installing and loading stuff.

MolecularNodes/pkg.py
```diff
@@ -32,52 +32,58 @@ def verify():
     verify_user_sitepackages(site.getusersitepackages())
 
 
-def run_pip(cmd, mirror='', timeout=600):
+def run_pip(cmd_list=[], mirror='', timeout=600):
```
janash, line 35: Just taking a quick glance through open PRs. I haven't looked too deeply into this PR, or it's purpose, but I will advise against this particular line. Empty lists should never be put as default arguments in Python  - see https://nikos7am.com/posts/mutable-default-arguments/

The preferred pattern is `cmd_list=None` and setting `cmd_list = []` within the function if `cmd_list` is None.

YaoYinYing: Wow thanks! That is quite interesting feature. I will fix this issue with adding explicit typing.

Infact this PR was expected to check whether the installed dependencies meet the version of them recorded in requirements. If no, the installation button will re-appear so that users can upgrade them in an easier way.

BradyAJohnston: Thanks for the info @janash ! The expert advice is definitely welcomed :)

BradyAJohnston:
Sorry for not having commented on this PR sooner. I think I want to potentially change how the dependencies work as well, adding an explicit button for installing each individual dependency, and a property as to whether those dependencies are available. 

That way if someone is able to install `biotite` but not `MDAnalysis` they can still use some functionality of the addon. I think it would also be good for the future as more functionality is added to the addon (EM density maps, EM topology files, solvation files etc) so that a user doesn't have to install all of the packages if they only want the one. I've gotten the code mostly re-written and I'll put it into this PR hopefully over the weekend.

YaoYinYing:
> That way if someone is able to install `biotite` but not `MDAnalysis` they can still use some functionality of the addon.

Yeah I totally agree with that.

BradyAJohnston:
Okay I have spent a bunch of time and done a pretty thorough overhaul of the whole package management system.

Package installations are now handled on a package-by-package basis. Which packages and their versions that are required are pulled from the `requirements.txt` file. In the addon preferences, the panel will show a button for each package if it needs to be installed.

image

If the machine is detected to be M1 / M2 with `is_apple_silicon()` then the button to install MDanalysis is grayed out and unclickable, with a link to the installation instructions specific for those machines:

image

Similarly, the individual panels are grayed out with a warning stating that thequired package needs to be installed via the preferences panel.

image

This is again handled on a package-by-package basis, so that `biotite` can be successfully installed and used to import structures from the PDB, without MDAnalysis having to also be installed if the user is getting errors.

Once installed and detected as being able to be imported, the buttons switch to 'Reisntall {package}'  and display information about the version that is currently installed.

image

I have also switched the logging to be done by a proper logger as well, so that the logs can be used from multiple functions and we can potentially use different logs for different purposes.

BradyAJohnston:
If you could have a look through @YaoYinYing and test it on your own machine, if everything seems to be working okay I can merge it to main.

YaoYinYing:
Great work Brady! 

I'll bring the PyPI Mirror back and do some minor refactors.

But overall this is brilliant.

BradyAJohnston:
Ah yes I forgot to add back in the option for selecting the mirror. Will wait for you to add them in :)

BradyAJohnston (approved):

BradyAJohnston:
All looks pretty good to me! Thanks for the updates and refactors.

Issue #174, opened 2023-03-09
Bug with Default Selection in MD Trajectory 


kimjoyc:
**Describe the bug**
Blender crashes when some molecular dynamics trajectories are loaded through MolecularNodes with an import filter. 

**To Reproduce**
Load the DNA system with the default import filter and attach the prmtop and dcd files from the molecular-nodes-bug repository link below located in the **Additional context** section.  Please refer to the repository to reproduce the bug for all test cases. 

The test case below involves a bond conflict with DNA where selection of resid 20 that leads to crashing, due to partial molecule selection unaccounted for in the atoms indices versus the bond indices. 

**Case 5 - Blender Crashes**
Make the following choices when importing:
- [ ] Choose "Ball and Stick"
- [ ] Type resid 20 in the default selection box.
> resid 20 represents the last DNA base pair. It will be bonded to resid 19

**Expected behavior**
Expect Blender to not crash and to load the trajectory.

**Error Codes**
Please include associated error codes that are shown in the python console when the error occurs.
> To view the python console on Windows, `Window` -> `Toggle System Console`, on Mac you must start Blender from the terminal.

Blender is crashing with exit code 139 and is found by going into Blender environment running ./blender executable on the terminal then following one of the test cases to make it crash. Then error is retraced with echo $? that yields 139. The Exit Code 139 means that the there is a SIGSEGV from Blender. It is most likely due to the bond conflicts relative to the all the atoms present.

```
# paste error codes from python console here
```

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows / MacOS / Linux]
 - Hardware: [e.g. Intel Mac, M1 Mac, x86 PC]
 - Blender Version: 3.4
- MolecularNodes Version: 2.4.1

**Additional context**

molecular-nodes-bug repository link below:
https://github.com/janash/molecular-nodes-bug 

 **Note: There is an incoming PR fix for the bug.**

BradyAJohnston:
Thanks for the info on the bug! I've had a look through https://github.com/janash/molecular-nodes-bug and using the supplied code (and what was in your PRs) certainly fixes the problems. 

There has now been two PRs that have been opened and then closed before I could review them. Is there something I can do to help out with the creation of a PR? The solution seems to work, so I'm pretty happy to review and merge.

kimjoyc:
I have opened a new PR today. Needed to edit and fix it.

https://github.com/BradyAJohnston/MolecularNodes/pull/177

PR #177, opened 2023-03-13
Fixes Issue #174

kimjoyc:
Fixes Issue #174 

Fixing the bug related to partial selection of a molecule when using Ball and Stick resulting in Blender crash.
The crash in the load_trajectory function in md.py is due to selections using MDAnalysis that references atoms that are not in the selection are left in the bond list.

A solution to resolve this is to reorder the bond indices when residues or fragments are not next to each other and discard bonding where an atom is missing.

### DIFF
--- a/MolecularNodes/md.py
+++ b/MolecularNodes/md.py
@@ -127,8 +127,8 @@ def load_trajectory(file_top,
             else:
                 bonds = univ.bonds.indices
 
-        else:
-            bonds = []
+    else:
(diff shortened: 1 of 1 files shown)


BradyAJohnston (approved):

BradyAJohnston:
There was a minor error with some indentations, which I have gone ahead and fixed. Otherwise the rest looks good. Thanks so much for the bug fix and the contributions!

PR #182, opened 2023-03-23
Fix consistency in load_trajectory function call

janash:
Hi @BradyAJohnston - thanks so much for providing this software package!

This is a small PR that changes how the `custom_selections` are obtained for the `load_trajectory` function. For all of the other function variables, the parameters are passed from the GUI in `ui.py` as function arguments. However, `custom_selections`,  it is obtained from the GUI in the function itself. This prevents the use of the `load_trajectory` function with a custom selection outside of the context of the GUI and makes the function less flexible.

This PR adds an additional argument to `load_trajectory` called `custom_selections`. It is set to `None` by default. Another change is made in `ui.py` so that the value is obtained from the GUI and passed into the function here, similar to the other variables used to load in an MD trajectory.

I think this doesn't change the behavior of anything, but makes the code more internally consistent and flexible. @kimjoyc and I are working on some visualizations for solvation shells and we'd like to be able to use the `load_trajectory` function and pass in some custom selections using Python.

Thanks! Let me know if there is some design choice I'm missing on this.

### DIFF
--- a/MolecularNodes/md.py
+++ b/MolecularNodes/md.py
@@ -71,7 +71,8 @@ def load_trajectory(file_top,
                     include_bonds = False, 
                     del_solvent = False,
                     selection = "not (name H* or name OW)",
-                    name = "default"
+                    name = "default",
+                    custom_selections = None,
                     ):
     
     import MDAnalysis as mda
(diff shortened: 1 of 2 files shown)


BradyAJohnston (approved):

BradyAJohnston:
Thanks for the PR! 

You are right that this was an inconsistency that I had missed, thanks for the quick fix!

 This was my first project with python (die-hard R fan), so I welcome any other design feedback or suggestions well from anyone who has more experience with _python_ way of doing things.

I am also happy to take a look / give feedback on the solvation analysis code as well. It's not something that I have experience with, but happy to give advice more about Blender / Geometry Nodes / Molecular Nodes internals.

PR #186, opened 2023-03-30
Add Testing Framework

BradyAJohnston:
So far I have been slack on introducing formal unit testing & CI, as it's slightly more complex when having to try and test inside of Blender, which is a desktop app and not a installable package. 

I have found the [`blender-addon-tester`](https://github.com/nangtani/blender-addon-tester/actions) which aims to solve such a problem by scraping the Blender downloads page for relevant versions, downloading them then running Blender in the headless mode to run various `pytest` tests.

Certainly the import code should be able to be tested with this method, and reported on with `pytest`. The testing of the internal nodes will be slightly more difficult. In conversation with Blender devs, saving a 3D model and comparing it to a previously created 3D model is their method of testing the nodes, so potentially will try and set up something along those lines. `.gltf` is likely the model of choice. 

This PR introduces the testing framework, which I am looking to add some more initial tests to in the coming weeks.


BradyAJohnston:
Using more robust testing in #186

PR #201, opened 2023-04-25
Reworking of UI

BradyAJohnston:
Combines the different panels into 3 different panels, 'Structure', 'Trajectory' and 'Scene'.

image

'Structure' has import options for downloading from the PDB, opening local structures, and opening EM density maps. 'Trajectory' has options for importing MD trajectories via MDAnalysis. 'Scene' currently only has importing of star files, but has potential for importing cellPACK files or other file types that mostly include instancing of lots of proteins.

### DIFF
--- a/MolecularNodes/__init__.py
+++ b/MolecularNodes/__init__.py
@@ -120,19 +120,27 @@ def register():
         subtype = 'FILE_PATH', 
         maxlen = 0
         )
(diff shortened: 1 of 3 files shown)


BradyAJohnston:
closing this, and will potentially re-do UI rework after #226

PR #203, opened 2023-04-27
Small fix to improve compatibility with starfiles.

tubiana:
Hi brady!

Starfiles from CryoEM don't have the column `rlnCoordinateZ`. I have added a test to check if the column is present, if not create it and fill it wit `0`.

### DIFF
--- a/MolecularNodes/load.py
+++ b/MolecularNodes/load.py
@@ -205,11 +205,11 @@ def get_secondary_structure(mol_array, file) -> np.array:
     
     dssp_to_abc = {
         "X" : 0,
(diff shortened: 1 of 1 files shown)


tubiana:
Correction proposal of the secondary structures. `G` and `I` are helices, but classified as `3` (coil).
see a full classification here as well : https://mdtraj.org/1.9.4/api/generated/mdtraj.compute_dssp.html
The second commit change `G` and `I` to `1`.

BradyAJohnston:
Thanks for picking these up. All looks good to me!

PR #205, opened 2023-04-27
Added ESMFold as structure import option

wjm41:
Seeing that ESMFold has been implemented in ChimeraX as well as an optional plug-in for PyMol, I figured I'd add this to MolecularNodes as well :)

Just a simple text input box to paste in an amino acid sequence, which calls the ESMFold API via `curl`.

I just saw that there is a UI rework in progress (#201 ) so that will have to be accounted for but the underlying function calling ESMFold should work - let me know if I should add in any extra warnings/guardrails to prevent improper use!

Screenshot 2023-04-27 at 20 30 42

### DIFF
--- a/MolecularNodes/__init__.py
+++ b/MolecularNodes/__init__.py
@@ -210,6 +210,7 @@ def register():
     bpy.utils.register_class(MOL_OT_Style_Surface_Custom)
 
     bpy.utils.register_class(MOL_OT_Import_Protein_RCSB)
+    bpy.utils.register_class(MOL_OT_Import_Protein_ESMFold)
 
     bpy.utils.register_class(MOL_OT_Import_Method_Selection)
     bpy.utils.register_class(MOL_OT_Import_Protein_Local)
@@ -251,6 +252,9 @@ def unregister():
     del bpy.types.Scene.mol_import_md_frame_end
     del bpy.types.Scene.mol_import_default_style
     
+    del bpy.types.Scene.mol_esmfold_name
+    del bpy.types.Scene.mol_esmfold_sequence
+    
     del bpy.types.Scene.trajectory_selection_list
     del bpy.types.Scene.list_index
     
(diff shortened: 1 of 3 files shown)


YaoYinYing:
Nice job! I got a few ideas for this PR.

- Since `biotite` has already required `requests` module for network accessing. You may reimplement this `curl` stuff with the help of `requests` to handle the POST and the complicated response in a more elegant way.
- the two functions for web structure fetching, `molecule_esmfold​` and `molecule_rcsb​`, share similar code logic. Maybe we can merge then into a general one? 
- The most importantly, @BradyAJohnston I think it's time for us to provide AFDB&ESMFoldDB accessing. But I am not sure whether `biotite` has native support of that. ;-)

But overall this PR looks cool!

BradyAJohnston:
Thanks for the PR! I like the idea and I've had a quick play around and it seems to work pretty well. The UI rework I am also unsure about ha. I started to work on it, but honestly I don't know if I like it and I think having the clearer buttons is nicer. Regardless it can be addressed later. 

Some notes on the implementation: 
- I tried some longer sequences and it failed. I assume there is a limit on the API, so stopping the user from sending longer than is possible would be good. The same with handling a couple of errors that might come back with informative error messages.
- The response times were pretty quick for me, but there is also potential for implementing modal operators, which can be dispatched and run in the background. This allows blender to give control of the UI back to the user while blender handles the request in the background until it completes and adds the structure to the scene. This is something that I have been wanting to implement for downloading of structures also but I haven't had the time to dive into it.

>the two functions for web structure fetching, molecule_esmfold​ and molecule_rcsb​, share similar code logic. Maybe we can merge then into a general one?

With these functions I was intending them to be a single function that a user can call from the python terminal, which handles everything for a particular structure to be imported. Internally they just call the various functions which are required for each step of the process. I think they can stay separate, but potentially the creation of the molecule & creation of the nodes could be wrapped into a single function. The whole python API & setup of functions could probably do with a little bit of an overhaul, but that is all stuff to be handled somewhere else I think.

> The most importantly, @BradyAJohnston I think it's time for us to provide AFDB&ESMFoldDB accessing. But I am not sure whether biotite has native support of that. ;-)

It would certainly be convenient to have them able to be downloaded. I haven't given it large priority as I figured users could always just download and open it themselves, and I also know very little about using web APIs haha. If others wish to implement something for them, that would be great. 

---
In terms of this PR, I think finding out what limits there are for the API and implementing those in the UI is required with maybe a helpful error message if the API returns an error. Otherwise it's all great and pretty ready to go!

BradyAJohnston (approved):
All looks pretty good! I've run some tests and it seems to work well. The cartoon representation is funky, but that is something that I am fixing in #215 so it's not a problem here.

PR #214, opened 2023-04-30
Detection of backbone, lipids + solvent for Martini CG models

tubiana:
This is the first step of adding the compatibility to Martini Coarse Grain support :
- `BB` atoms are recognized as "backbone" and "name CA"
- `resname W` as solvent
- lipids as lipids (should works with all atom as well)
image

Also, A Quick test showed that the ` ribbon` representation works fine on the CG protein structure.
image


BradyAJohnston:
Nice to see it working so well already!

In terms of memory, I have chatted with the Geometry Node devs about it and adding another boolean attribute is usually better than dealing with integer attributes instead of performance and memory (makes sense), so adding another one that specifies `is_lipid` I think will be good.

Rather than doing a selection through a string, you could store the lipid names inside of `data.py` as a tuple, then we can use that and do a check of the residue names themselves. It should in theory be also slightly more performant than doing the round-trip through the parsing of a MDAnalysis selection string.

```python
 def is_lipid():
        return np.isin(univ.atoms.resnames, data.lipid_names)
```

tubiana:
It works like a charmm! I will update the code :) 

You are right, my ribbon is continuous because In my system there is no chain information (yet), I use the gromacs TPR file as topology which doesn't include this information. 
But I might use another topology file later on, for now it was just a quick test ^^

Bernadette-Mohr:
Martini support would be so great!
As far as I can see it, these additions are not part of the 2.7.3 release? If I can in any way, I would be happy to help.

BradyAJohnston:
Hi @Bernadette-Mohr thanks for the interest! This isn't currently part of the main release, but martini force fields should still be able to be opened inside of Molecular Nodes regardless. There just may not be all of the information inside of imported structure which are specific to martini force fields.

@tubiana has already added a few of the attributes which would be more specific to martini force fields. I have not worked with them myself, but I am unsure if others might still need to be added. The changes as they are should be fine to be merged, I believe we were just waiting on potentially adding some more attributes.

If there are specific aspects of the martini force fields you would like to see included then we can likely include it easy enough as the changes so far have been mostly straight forward.

BradyAJohnston:
Hi @Bernadette-Mohr thanks for the interest! This isn't currently part of the main release, but martini force fields should still be able to be opened inside of Molecular Nodes regardless. There just may not be all of the information inside of imported structure which are specific to martini force fields.

@tubiana has already added a few of the attributes which would be more specific to martini force fields. I have not worked with them myself, but I am unsure if others might still need to be added. The changes as they are should be fine to be merged, I believe we were just waiting on potentially adding some more attributes.

If there are specific aspects of the martini force fields you would like to see included then we can likely include it easy enough as the changes so far have been mostly straight forward.

tubiana:
My bad on that, I thought I could finish it before a congress, and I actually didn't got time, and now I'm quite booked until the 21st of June. I will try to finish what I have to do on that before the end of the month =D
Sorry again 😅

Bernadette-Mohr:
No worries, that's the same for all of us. 
I am positively surprised and hyped that there is interest in adding Martini properties at all! From what I've seen so far, the cardiolipins are missing from the lipid tuple ('CDL1', 'CDL2'), and there is a polarizable Martini water that could be added to the exclusion string ('PW'). Apart from that, I don't immediately see any further issues. I will play with loading a gromacs tpr file as coordinates today and see if that provides enough bond information for good representations.

BradyAJohnston:
Hi @tubiana, I'd love to get this finished up and merged to main. I want to get a new release out before working on some breaking changes for Blender 4.0. 

I've reformatted the code over to the new method of importing via MDAnalysis which maintains a connection to the underlying universe. 

Are you able to provide, or link to, some example files / trajectories that we can use for running tests? I haven't used martini so I don' have access to example files.

tubiana:
Hi @BradyAJohnston ! 
Sorry I haven't used a blender for months, and I thought it was already included. In my last trials, everything used to work, but once again, I didn't took the time to test it with your last updates... 

I'm sending you a link of a sample of my trajectories in your twitter DM, since they are unpublished.

tubiana:
[MD_samples.zip](https://github.com/BradyAJohnston/MolecularNodes/files/13282143/MD_samples.zip)
Here's two short MD samples. 

However, I found new issues with the new nodes: 

1. When using the `MN_style_presets`, protein CG beads are not displayed, but they are displayed with the `Atoms` style node. 

image

image
 
 2. Backbone atoms type are recognized as "Boron" (element 5). It would be great to recognized distinguely as `BB` (same with side chains `SC`). However, The function `mda.topology.guessers.guess_atom_element(x) for x in self.ag.atoms.names` in mda.py is the one causing this issue, I don't know if we can replace it by something else. 

3. Even if "is_backbone" attribute is present on certain atom types, we cannot use the "backbone selection node" to select them
image

4. Elements are case sensitive. So when ions are in upper case, they are not recognized. We can double the element line with the upper case version (like `"NA" : {"atomic_number" : 11, 'vdw_radii': 227, "name" : "Sodium"      },`)

tubiana:
I proposed two quick fixes for my previous point 2 and 4 (issue on atoms names). It's a bit ugly but it works...
BB are now atoms considered as carbon (better for coloring) but with a bigger radius, Side chains are still consider as sulfur (same, better for coloring) with a bigger radius. 

For (1) (not working with presets 1), it's normal, I haven't realised that it should be displayed in Cartoon. It's only when I opened the node that I found this... My bad.

For (3), haven't found the solution yet.

tubiana:
I have update elements list for Martini Lipid names (GL* are glycerol, I gave them the oxygen atom type, and D2* are another type of carbon). Now they have full color 
image

Something that could be added as well in the "Attribute" node (I'm not sure where to update it...), is to add the `is_lipid` type.

BradyAJohnston:
Thanks for the sample trajectories, and for doing some testing and putting together the fixes. I'll put together some tests for it, add the 'is_lipid' to the attribute node.

We will probably want to put together some documentation on importing martini simulations, and what we are doing in terms of assigning different beads to different elements etc.

PR #219, opened 2023-05-09
Molecular Nodes Documentation

FL-ECHE:
Hi, I've been following your project for a while and wanted to work on it, so while going through the code and git, I told myself that documenting it could help me understand how you implemented it. I thought it might benefit newcomers as well for your tools has been actualized and renewed quite a lot.

I understand that it's not much compared to new functionalities, but I do intend to try some things, like using the new simulations nodes to actualize backbone state based not on position update, but actual constraints to mimic how FF/MD simulation look.

A few other things would be to try and export the PDB file of the manipulated protein, as I've seen it is possible to access atom coordinates, I just have to find the right library to write the full file. Another would be CIF support, as in 2024 new proteins will not have a PDB file generrated for them (because three letter ligand coding is exhausted it seems)

Maybe generating "docking" maps with weight maps on surface would be cool, although it might be too approximative.

So yeah I hope I can contribute in the future, if this merge is accepted, I plan on making some "artistic" nodes, and new examples leveraging as much functionalities as possible, but if not I understand and will try to make the next one more useful.

Have a nice day :)
FLECHE

PS : feel free to replace the images


BradyAJohnston:
Hi @FL-ECHE! Thanks so much for opening the PR. Documentation is incredibly important (else nobody knows how to use the thing you have created), so don't sell your contribution short, this is fantastic! 

I really appreciate the effort, and documentation is something that I haven't updated in a while as it takes a fair amount of time (which you will have come to appreciate I am sure). Like what you have written, and I might make some additions in in the form of formatting and some extra images, but otherwise it's all looking pretty good, and a great and very welcome addition!

In terms of your other ideas, I like them: 

- _PDB Export_: This should certainly be possible. The addon utilises the `Biotite` python package for reading most of the structure files, and this package also has the ability to write `.pdb` and other formats. Biotite uses `numpy` array stacks for storing the structures, which work quite well with the Geometry Nodes data formats. You would need to translate from the numerical representations of `res_name`, `atom_name` etc but this should actually be quite straight forward. Certainly there is potential for this feature if you wish to try and work on it. Adding this would also potentially enable the use of other python libraries that operate on structures like docking programs, electrostatic calculations, MD simulations etc.

- _CIF Support_: `Biotite` already provides support for these files, so they should work with the addon as far as I am aware?
- _Docking Maps_: might be complex to implement, but could work potentially work if the _PDB Export_ enabled leveraging of existing libraries. Might be hard to implement through solely Geometry Nodes.
- _New Styling Nodes_: I would be very excited to have some nodes contributed! So far I have been the sole developer of all the nodes, but it would be great to have nodes contributed by others as well. The only problem is that the nodes are opaque to `git` and so can't be version controlled well. Basically have to double check everything manually, so it might require some more back and forth but this would be great.
- _FF / MD with Simulation Nodes_: I too am tinkering around with this for potential approximations of MD atom movement. I'm starting to have [some success](https://twitter.com/bradyajohnston/status/1655217012770242560?s=20) but there is still a long way to go. If you have more success with it, I would be very interested with how you approach it. If we can get some simple atomistic simulations working, then this would open up a huge amount of potential animation tools to be created that are at least _somewhat_ realistic.

Issue #245, opened 2023-06-12
Style Split-chains broken


BradyAJohnston:
Due to renaming of surface node outputs, broken the split-chains.

PR #247, opened 2023-06-12
Fixed `create_custom_surface`

jahooker:
Hi Brady!

I have a fix here for the `Surface Split Chains` style, which previously would be generated with missing links (#245),
though I see that you've already got #246.

Best,
James

### DIFF
--- a/MolecularNodes/__init__.py
+++ b/MolecularNodes/__init__.py
@@ -34,6 +34,7 @@ auto_load.init()
 def register():
     auto_load.register()
     bpy.types.NODE_MT_add.append(mol_add_node_menu)
+
 def unregister():
     bpy.types.NODE_MT_add.remove(mol_add_node_menu)
     auto_load.unregister()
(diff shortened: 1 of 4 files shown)


BradyAJohnston:
I ended up going with the fix that I had in #246. I went with this because I also made some other changes around the surface nodes. For the linking of the different outputs, we should  link them all in the node itself regardless of whether they are used, rather than having an option for it in the function. The particular part of the node graph is only executed when it is connected to other nodes as well anyway, so there is no performance hit at all and it just ensures that everything that should be connected is connected. 

For the assembly stuff, this is something that I have been meaning to overhaul for a while. The improvements you have made a good, so you can remove the other stuff from this PR or open another one and we can probably merge that in. 

I do still plan on doing a more major overhaul of the assembly workflow, which the initial requirements are created in #148 but it ends up being quite a major overhaul so I haven't gotten to it yet.

jahooker:
Thanks for the reply!
I've `git cherry-pick`ed the changes I made to `assembly.py`.
See #253.

Issue #256, opened 2023-06-27
Cache downloaded structures when using PDB ID


BradyAJohnston:
The downloaded structures from the PDB using PDB ID currently aren't cached anywhere, and are discarded after download. 

Biotite [has support](https://www.biotite-python.org/apidoc/biotite.database.rcsb.fetch.html#biotite.database.rcsb.fetch) for caching of downloaded files, it just needs to be implemented with an option in the preferences for where to store the downloaded files.

PR #260, opened 2023-07-03
Cache PDB downloads

PlethoraChutney:
Adds biotite caching, resolving #256. Tested by downloading a structure, turning off wifi, and re-loading it successfully. The cache folder is also created and populated with models.

I'm mostly working on this one to figure out how to add stuff to MolNodes, but it does work. It looks like there is a lot of duplicating default values --- I tried to copy your style but I'd be happy with feedback.

### DIFF
--- a/MolecularNodes/load.py
+++ b/MolecularNodes/load.py
@@ -3,7 +3,6 @@ import io
 from pathlib import Path
 import bpy
 import numpy as np
-import json
 from . import coll
 import warnings
 from . import data
@@ -11,9 +10,6 @@ from . import assembly
 from . import nodes
 from . import pkg
 from . import obj
-
-with open(Path(__file__).parents[0] / 'config.json', 'r') as f:
-    config = json.load(f)
 
 bpy.types.Scene.mol_pdb_code = bpy.props.StringProperty(
     name = 'pdb_code', 
@@ -27,7 +23,7 @@ bpy.types.Scene.mol_cache_dir = bpy.props.StringProperty(
     name = 'cache_dir',
     description = 'Location to cache PDB files',
     options = {'TEXTEDIT_UPDATE'},
-    default = str(Path(*config['default_cache_dir']).expanduser()),
+    default = str(Path('~', '.MolecularNodes').expanduser()),
     subtype = 'NONE'
 )
 bpy.types.Scene.mol_import_center = bpy.props.BoolProperty(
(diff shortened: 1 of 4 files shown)


MolecularNodes/ui.py
```diff
@@ -86,12 +87,18 @@ def MOL_PT_panel_rcsb(layout_function, ):
     col_main.scale_y = 1.0
     col_main.alignment = 'Expand'.upper()
     col_main.label(text = "Download from PDB")
+    row_cache = col_main.row()
+    row_cache.prop(
+        bpy.context.scene,
+        'mol_cache_dir',
+        text = 'Cache dir',
+        icon_value = 0,
```
BradyAJohnston, line 95: I've got it a couple of places in my code, but it turns out we don't need `emboss = True` and `icon_value = 0` as these are the defaults anyway, so you can trim these if you want.

BradyAJohnston (commented):
I think you nailed it first try. This adds the feature exactly as I was hoping and from my testing works perfectly! 

I don't know how you'd do it, but potentially adding some kind of test would be good. Testing is tricky with the addon as you have to run them through the headless scripts, but maybe a test that runs and checks that the file has been downloaded would be all we need. 

Otherwise fabulous work and thanks for the contributions, it's basically ready to be merged with a couple of minor changes.

PlethoraChutney:
I guess, looking at it, the `ui.py` stuff doesn't need to know where the cache actually goes, so the default location only goes in one file. Sorta removes the need for `config.json`, I can remove if you want

BradyAJohnston:
>Otherwise, made all the changes you mentioned. The test I added only checks whether a file with the right name got downloaded --- I don't do any checks to make sure it's the right structure.

`biotite` is already quite comprehensively tested, so I think we only need to test that the file is actually cached and that should be all.

For the `config.json`, maybe we ditch it if it is only being used once, unless you can think of other things we should also be storing in there.

MolecularNodes/load.py
```diff
@@ -167,11 +180,11 @@ def molecule_local(
     return mol_object
 
 
-def open_structure_rcsb(pdb_code, include_bonds = True):
+def open_structure_rcsb(pdb_code, cache_dir = Path(*config['default_cache_dir']).expanduser(), include_bonds = True):
```
BradyAJohnston, line 183: For this function, should the default be `cache_dir = None` to match the underlying `biotite` function? In the addon code we'll be supplying a location, for interfacing it via python though would a user expect it to be cached by default?

BradyAJohnston (commented):
Looking good! Minor comments but otherwise it's good to merge once those are done

PlethoraChutney:
Agreed on both fronts. Even if I can think of a reason to use an external config it should be a different PR. And using None makes more sense for the functions.

BradyAJohnston (approved):

PR #264, opened 2023-07-04
Fix magnesium color bug

### LINKED ISSUE #254: can't change color of ions (?)
**Describe the bug**
Ions show up fuchsia in my renders, and I can't affect their color at all, even through the color by element options. 
**To Reproduce**
import a structure with ions in it, like 7r97
color by chains, use the normal cycles styles
you'll see some central fuchsia balls, and no matter what you do, their color will not change

**Expected behavior**
I wanted their color to be the same as the backbone I guess? 

**Screenshots**
These fuchsia balls are some of the ions: 
image
* I've already removed some of the ions, didn't realize there were more 

**Desktop (please complete the following information):**
 - Windows, x86 PC
 - Blender Version: 3.5
- MolecularNodes Version: 2.5.4

PlethoraChutney:
The color-by-element node was mis-wired between atomic numbers 14 and 15. This resolves #254.

image


BradyAJohnston:
Fabulous! I still haven't come up with a good way of actually versioning the hand-made nodes via git (`.blend` files are opaque to it) so I think we basically have to rely upon using tests to ensure that things don't change majorly when people edit the files.

BradyAJohnston (approved):

PR #276, opened 2023-07-24
Update md.py

tubiana:
Correction of bfactor type in md.py ("float" -> "FLOAT").

When loading a PDB with mdanalysis (useful with large PDB) the BFACTOR is ignore because the type does not match (expect "FLOAT" but "float").


BradyAJohnston (approved):

BradyAJohnston:
Ah thanks for catching this! MDAnlaysis also added support for b-factors for each timestep, if you wanted to have a go but otherwise I'll also do that soon.

PR #283, opened 2023-08-03
Adds support for importing CellPack models

BradyAJohnston:
Imports CellPack models that are stored as `.cif` files. Creates a data storage object which stores all of the required rotation / translation operations, and creates a collection which contains an object for each protein. The collection can then be used as instances for instancing on the data object to create the assembled model.

image


corredD:
Looks great, except for one protein that should not be at the surface. You can see the model here for comparison : https://molstar.org/dev/me/
image
I was able to get the branch running on my workstation, however I am really not familiar with geometry nodes. Here is what I have after loading a small cif file
image
What should I do to add some style to it ? 
About the metadata, it would help if you prepare some template function that would change the color style but with random value. I can help replace after that with some coloring based on function or other attributes.

BradyAJohnston:
Also tested on the mature ISG.
image

Fof the lipid `.cif` files, I get this error which I asume is related to the way biotite deals with the files which we discussed in the other PR: 

```python
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 131, in execute
    load_cellpack(
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 37, in load_cellpack
    obj_data, coll_cellpack = open_file(file_path, name)
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 49, in open_file 
    file_open = pdbx.PDBxFile.read(file)
  File "C:\Users\BradyJohnston\Documents\Blender Launcher\stable\blender-3.6.0+stable.c7fc78b81ecb\3.6\python\lib\site-packages\biotite\structure\io\pdbx\file.py", line 138, in read
    data_block,
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 131, in execute
    load_cellpack(
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 37, in load_cellpack
    obj_data, coll_cellpack = open_file(file_path, name)
  File "C:\Users\BradyJohnston\AppData\Roaming\Blender Foundation\Blender\3.6\scripts\addons\MolecularNodes\pack.py", line 49, in open_file
    file_open = pdbx.PDBxFile.read(file)
  File "C:\Users\BradyJohnston\Documents\Blender Launcher\stable\blender-3.6.0+stable.c7fc78b81ecb\3.6\python\lib\site-packages\biotite\structure\io\pdbx\file.py", line 138, in read  
    data_block,
UnboundLocalError: local variable 'data_block' referenced before assignment
```

BradyAJohnston:
Hey @corredD is there somewhere that brings together publically available CellPack models? `.bcif`/`.cif`? Would be helpful to have a good resource for it so I can potentially do some more robust testing

corredD:
thanks for the update! it looks great. I have been looking at it and I found some issues with my cif exporter and your operator parser. First, it appears some of my files were encoded in UTF8+BOM which were not readable by biotite until I change the encoding to UTF8. So I change my exported so that it will never write the BOM. Second, the operator parser was not handling complex description like  : 1 '(2-2,89-89,150-150,207-207,300-300,393-393,462-462,501-501)' AM. I made a pull request with some changes that should handle it. 
Our models are not stored in one place yet, once I settle on a cif exporer I will probably put all the models here : https://github.com/mesoscope/cellPACK_data .
About the Lipids, the error you see is the encoding error, so you can re-save the file in UTF8 and biotite should be able to read it. But I am thinking that in the context of molecularnode, maybe fibers and lipids should be done procedurally with some nodes. For the lipids I can provide the compartment mesh and individual lipids, and for the fiber, you already have some nodes for generating dna, so it wouldnt be too much developpmt for other type of fiber, given a list of control point and a list of atoms.

BradyAJohnston:
Thanks for the updates to the parser @corredD and the info regarding the files. 

Lipids and fibres could certainly be done procedurally through Geometry Nodes if given the correct control points etc. 

I think initially having them be imported successfully alongside the other chains should be priority, then we can try and procedurally generate some aspects. Are the lipids packed as well? From looking at the models it seems like blocks of the lipids are packed, then the block of lipids is instanced and you don't necessarily pack individual lipids around embedded membrane proteins.

BradyAJohnston:
@corredD I fixed the getting of rotations & translations from the `.bcif` files, so the parsing of those is currently working well, but I can't seem to get the `chain_id`s to match up, so it isn't instancing the correct chains. I am kind of confused how things seem to keep changing between `auth_asym_id` and `label_asym_id` and which one is being used for the operators.

I think I am going to need some help to clear that up, because I have the correct rotations and translations now being import (although I think the scale for the translations might be off by a factor of 10?), but the instancing of the correct chains currently isn't working properly. Works for importing from `.cif` but the `.bcif` is still proving problematic

corredD:
the way I understand it is from https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Categories/pdbx_struct_assembly_gen.html and it seems the default would use the label_sym_id. If you look at some example in the PDB like 2plv (https://files.rcsb.org/view/2PLV.cif) the auth_asym_id are 1,2,3,4, the label_asym_id  are A,B,C,D and the one used in the _pdbx_struct_assembly_gen loop are A,B,C,D.  I made a pull request that fix some of the issues in the bcif parser, but I still have some issue with some models and I am trying to figure out why.

BradyAJohnston:
Thanks for the fixes. I tried poking around and testing with the covid `.bcif` and the synaptic vesicle `.bcif` things are still weird. 

It seems that there just doesn't seem to be matching between the `chain_id` that is extracted from the symmetry operators and the `chain_id` which exists in the model. I kept tweaking it and changing it but it would either have a different name for each chain, which would result in lots of indiviudual models but symmetries that only wanted 4-6 models, or you would get 4-6 models but the individual models would be combinations of chains from random different proteins / lipids that shouldn't be going together. 

I think I am getting the mapping of the IDs etc wrong when expanding out from the symmetry operations but I can't figure out where.

corredD:
Theses are Petworld models (http://download.yasara.org/petworld/index.html), they extended the mmCIF categories and require a different parsing approach. Here is the description from their excellent paper : "
Models consist of objects (proteins, membrane blocks, and nucleic acids, possibly consisting of multiple chains/molecules). Each model can be saved in the mmCIF file format to enable straightforward cross-platform data sharing. Each object is stored as a PDB “model” (not to be confused with the entire model) with increasing atom_site.PDB_model_num. Pet models are saved atom by atom without instances and thus without compression, using chemical element symbols O1–O9, OA, OB, and OC, which should be displayed as spheres with radii from 0.1 to 1.2 Å. All-atom models use object instances and are compressed by storing them as biological assemblies, which means that only one transformation matrix (i.e., one line in the mmCIF file) is needed per instance (_pdbx_struct_oper_list.matrix/.vector). To make sure that each _pdbx_struct_assembly_gen entry can reference a certain object, _pdbx_struct_assembly_gen.PDB_model_num items were added as a little extension. Also, the mmCIF format does not normally allow to give a separate name to each object (= PDB model), that is why we added _pdbx_model entries that provide the object/model name (_pdbx_model.name) and the number of instances (_pdbx_model.instances). If the parser encounters a model with more than one instance, it creates them immediately from the pdbx_struct_assembly data. Otherwise, the biological assembly data are only read, but ignored until the user explicitly requests an oligomerization." (https://pubs.acs.org/doi/full/10.1021/acs.jcim.1c00743)

corredD:
not ideal but I made a small hack to load the petworld bcif file, in another pull request. 
image

corredD:
How do you map the chain Id and the transformation ? I can see how you parse the data but not how they are linked. Which node is used ? Or how it generated that table : 
image

BradyAJohnston:
That little fix seems to have fixed loading for everything: 
image

BradyAJohnston:
I'm not entirely sure I understand which part of the pipline you are asking about. The setup currently is that that an object is created with a single vertex per instance, with that vertex containing the rotations & translations required. 

Then inside the node groups, there is a group that just instances chains from a collection, and picks them based on the `chain_id`, applying the rotation and translation that was present on that single vert.


image

Blender is _kind of_ limited in that I just have to make sure that the chains being instanced are correctly ordered inside of the instance collection that I am pulling from. As long as they are in the order that matches their `chain_id` then they will be picked correctly.

image

BradyAJohnston:
Even seems to work for the really really big ones

image

BradyAJohnston:
@corredD I did a bunch of cleanup and refactoring and I was pretty happy with how it was handling everything. Thanks for the assistance in getting it all working. Happy with how it is currently for initial support. 

The next things to add would be using the custom color palettes and adding other specific attributes that could be useful.

PR #287, opened 2023-08-10
Access and Stream Trajectories with MDAnalysis

### LINKED ISSUE #292: Implement Access and Stream Trajectories with MDAnalysis inside MolecularNode
### Issue Description:
We are planning to enhance the capabilities of MolecularNodes by integrating the ability to access and stream trajectories using MDAnalysis. This feature would enable users to
- efficiently load molecular dynamics trajectories without dumping all the frames into memory.
- leverage the power of Blender as a live viewer for trajectories e.g. inside a jupyter notebook with [BNotebooks](https://github.com/BradyAJohnston/BNotebooks) Blender kernel.
- pave the way for scripting rendering pipelines.

### Proposed Design:
To implement this feature, we propose the following design in scripting:

#### start a session
```python
mda_session = mn.mda.MDAnalysisSession(world_scale=0.01)
```

#### show atoms

```python
universe = mda.Universe(top, traj)

custom_selections = {'name_ca': 'name CA', 'ion': 'resname SOD CLA'}

# This will create three representations inside Blender
mda_session.show(universe, name='protein', selection='protein', custom_selections=custom_selections)

# Add new representations for new selected atoms
ligand = universe.select_atoms('resname LIG')
mda_session.show(ligand, name='ligand', representation='ball+stick')

# Add representations for dynamically selected selected atoms
around_ligand = universe.select_atoms('protein and byres (around 4 resname LIG)', updating=True)
mda_session.show(around_ligand, name='around_lig', representation='vdw')

# add new trajectory and an offset of the frame can be set
universe2 = mda.Universe(top2, traj2)
mda_session.show(universe2, name='protein', selection='protein', frame_offset=100)
```

#### Persistence

```python
# saving
file_path = "blender_and_mda.blend"
bpy.ops.wm.save_as_mainfile(filepath=file_path)

# loading
bpy.ops.wm.open_mainfile(filepath=file_path)
mda_session = bpy.context.scene.mda_session
universe = mda_session.universe

# persistent even when trajectory files are deleted
mda_session.transfer_to_memory(start, step, end)

# or use legacy code with Blender UI
```

Example:
- showcase: https://drive.google.com/file/d/1A_WlN8oUaMqjV6MsKP4F2lSvtRKa6YVt/view?usp=sharing
- notebook: https://gist.github.com/yuxuanzhuang/25aa5a1588c7e19efbfc1b7275718b58
### Expected Benefits:
- a consistent python API to interact between Blender and MDAnalysis
- better organization of atoms representations
- Memory efficiency
- [On-the-fly transformations](https://userguide.mdanalysis.org/stable/trajectories/transformations.html) e.g. fixing PBC, designed rotation, smoothing (position averaging).

### Expected Caveats:
- the current implementation of blender kernel will open blender automatically in the initialization stage. Ideally, it should be only after a new MDAnalysisSession is started.
- Need to have a consistent connection to the trajectory in the session.
- During the animation, the state of the Universe will be altered. A racing condition is likely to get induced when the trajectory frames are iterated in the notebook simultaneously.
- `MDAnalysisSession` (together with `Universe`) will be dumped into a tmp location during saving .blend file i.e. `~/.blender_mda_session/` by default.
- a relative path of trajectory is saved for Universe -> where does Blender start matters!
- ~~potentially create conflicts with the current animation engine and the selection process within the geometric node.~~

### Tasks (~~task~~ will not be implemented in this PR):
- [x] Design and implement `MDAnalysisSession` and `AtomGroupInBlender`.
- [x] Loading and Saving (Persistence)
- [x] Deletion
- [x] Start with different representations
- [ ] ~~Function to switch representations~~
- [x] `Atomgroup` with variable n_atoms i.e. a dynamic selection of atoms
- [ ]  ~~Update attributes if the topology is modified~~
- [x] Multiple universes
- [ ] ~~ways to add new time-evolved attributes and display in blender (e.g. occupancy)~~
- [ ] Benchmark on streaming
- [x] Write unit tests to ensure reliability and stability.
- [x] Update documentation/tutorial to reflect the new feature and its usage.

### Future direction
- visualize analysis results e.g. simple distances, angles, etc.

yuxuanzhuang:
Fix #292

Summary of Changes:

- Registration of handlers to update trajectory frames and save the mda session now occurs at the outset.
- The initialization of the previous import subroutine in Blender AI incorporates the `use_old_import` flag (I'm open to changing this to a more appropriate option or making it the default setting).
- Code responsible for generating the MDAnalysisSession and creating objects has been relocated to mda.py.
- The utilization of the AtomGroupInBlender class facilitates the conversion and retrieval of topology/attributes from AtomGroups to Blender.
- 'molecule' tag is added to all molecular objects in Blender. This helps to decide whether an mda session should be loaded when a blend file is loaded.


BradyAJohnston:
Thanks for the PR! I have been wanting to try and attempt this for quite a while, and it seems like you have mostly nailed a bunch of it straight away.  

I have had a bit of a play around with it and it seems to work really well initially. If I save and load a file I lose the connection to the trajectory and it stops updating. It also occasionally loses the connection to the data / universe while I am playing around inside of the scene, but I am unsure what specifically I am doing that is causing it.

Potential fixes for this (and for your note about multiple universes) is that the universe information could be stored on the object itself. I currently store some data on the objects (like chain IDs etc) so we could potentially store that data there instead of on the scene level.

I think having the option on import for the user to either stream the trajectory, or load in to memory, would be good as well. That way you can have a self-contained `.blend` file instead of having to worry about connections to other files if you don't want to. That would also still enable geometry-nodes playback through that import method.

I'm away for the weekend so I won't be able to give feedback on specific code until next week, but this is working really well already.

BradyAJohnston:
Additionally, if you have other suggestions for refactoring the `md.py` script, please do suggest them! I wrote the majority of it a year or so ago (very early to python then), and I haven't revisited it in a while so I would gladly accept suggestions and improvements on the larger codebase.

yuxuanzhuang:
> I currently store some data on the objects

It seems that only simple data types can be saved as attributes in `bpy.types.Object` and the `Universe` is not simple!

I think one possible way to rejuvenate the animation is by dumping/serializing the `Universe` into a string and deserializing it during the reloading process. Besides, I think most Python-level stuff will not be saved into `.blend` files e.g. functions inside `frame_change_post` so it also needs to be reappended during reloading. I just need to have a look at how it can be achieved :) and please let me know if you have any insights!

yuxuanzhuang:
Still WIP but now the whole session can be reloaded. Try the notebook with (https://github.com/cheng-chi/blender_notebook) here https://gist.github.com/yuxuanzhuang/25aa5a1588c7e19efbfc1b7275718b58

> It also occasionally loses the connection to the data / universe while I am playing around inside of the scene, but I am unsure what specifically I am doing that is causing it.

I don't seem to have this issue during the session but I also haven't played around with anything inside the geometric nodes yet...

BradyAJohnston:
I just had a play around and the saving worked well, the same with the adding of multiple objects. Even on testing on my old 2017 MacBook the streaming of trajectories works _really, really_ well.

The connection was stable for me while scanning through even 50,000 frame trajectories without any issue. I interacted with the node trees for a while and it was fine, then I changed something in the node tree and the connection was lost. unsure exactly what I did that caused the issue, but will have to try and poke around and find out what is causing the issue, otherwise I am really liking this implementation and it's working really well.

We might want to try and brainstorm some more design ideas around how to better interact with Blender through more convenience functions and classes. I don't really have much experience with classes, having the object be a class with a bunch of useful methods like changing the style or other common operations would likely help with interactions from the notebook itself. The same with interacting with the Blender scene in general, like setting cycles render engine etc, something to help with with the scene is likely a good idea, but I wouldn't know what would be best in terms of implementation, out of my depth in that respect.

Very excited by the prospect of this though, it has already turned Blender into a pretty great molecular viewer. Amazing what you've managed to do so quickly.

yuxuanzhuang:
I managed to reproduce the disconnection (error msg: `ReferenceError: StructRNA of type Object has been removed`) by simply undoing anything. It's probably due to the memory address that holds the pointer to the data changing every time an undo is performed. The latest commit should have fixed it.

MolecularNodes/mda.py
```diff
@@ -0,0 +1,409 @@
...
+    # default location to store the session files
+    session_tmp_dir = f"{os.path.expanduser('~')}/.blender_mda_session/"
+
+    def __init__(
+        self,
+        universe,
+        name="atoms",
+        md_start=0,
+        md_end=-1,
+        md_step=1,
```
BradyAJohnston, line 27: We can probably get rid of the start / end / step. By linking the change of frames to the change of the frame inside of blender, the user can set the scene frame range to select frames, and it's probably best that it remains linked to the underlying universe, so we can just initialise the entire trajectory and you select the frames by scrubbing through the timeline.

MolecularNodes/mda.py
```diff
@@ -0,0 +1,409 @@
...
+    @persistent
+    def update_trajectory(self, frame):
+        # TODO: how about a dynamic selection with variable number of atoms?
+        self.trajectory[frame]
+        for name in self.ag_names:
+            ag = self.ags[name]
+            mol_object = bpy.data.objects[name]
+            locations = ag.positions * self.world_scale
+            for vert, loc in zip(mol_object.data.vertices, locations):
+                vert.co = loc
```
BradyAJohnston, line 358: For updating the locations, you can write directly to the attributes of the mesh from numpy arrays instead of having to loop through the vertices individually. 

In my testing, I got approximately ~100x speed up by changing these two lines to the following: 
```python
for vert, loc in zip(mol_object.data.vertices, locations):
    vert.co = loc
```

```python
mol_object.data.attributes['position'].data.foreach_set('vector', locations.reshape(-1))
mol_object.data.update()
```

The only _gotcha_ is that currently to write to vector attributes such as position, you have to supply a 1D array so we have to just reshape it before setting. Not doing this can segfault blender. They have plans to improve this API eventually but this is what we currently have to work with.

MolecularNodes/obj.py
```diff
@@ -45,6 +45,7 @@ def create_object(name: str, collection: bpy.types.Collection, locations, bonds=
     mol_mesh = bpy.data.meshes.new(name)
     mol_mesh.from_pydata(locations, bonds, faces=[])
     mol_object = bpy.data.objects.new(name, mol_mesh)
+    mol_object['type'] = 'molecule'
```
BradyAJohnston, line 48: This is something I have been thinking about for a while. I haven't come up with any kind of scheme, but the different objects that are added to Blender though Molecular Nodes currently aren't tagged with any identifying information. Tagging with `molecule` and maybe other info is a good idea. Open to suggestions for naming schemes.

MolecularNodes/md.py
```diff
@@ -8,25 +8,27 @@
...
 from . import obj
 from . import nodes
+from .mda import MDAnalysisSession
 
 bpy.types.Scene.mol_import_md_topology = bpy.props.StringProperty(
     name = 'path_topology', 
     description = 'File path for the toplogy file for the trajectory', 
     options = {'TEXTEDIT_UPDATE'}, 
-    default = '', 
+    default = '/nethome/yzhuang/jupyter_playground/blender/md.tpr',
```
BradyAJohnston, line 23: Fine for these defaults to remain while you are developing, we'll change them later when we go to potentially merge.

BradyAJohnston (commented):
Like I've said I'm not _super_ familiar with a lot of class-based stuff (I've mostly come from R programming background) but it all looks pretty good.

It seems like already _as is_ this works REALLY well. 

A potential _gotcha_ would be when deleting an object. Currently if I delete an object I then get errors as it tries to update the object. There might be some kind of `on_delete` function we can register that cleans up objects when deleted.

BradyAJohnston:
I think I am pretty happy with most of this, and it might be a good idea to make a MN release that includes it pre MDA UGM. 

I might make some minor tweaks and fix the tests, then we might just need a couple of small tests and some minor documentation and I think a minimum version might be ready to merge, what do you think?

Then once it's in place for initial import via the Blender GUI, then we can look at improving the API for better scripting interface, probably during the hackathon.

yuxuanzhuang:
Thanks! That sounds reasonable! I’m not sure if it’s a minor version, though, since it breaks quite a lot of the API. Maybe label it as an alpha/beta version?

Also, let’s make sure not to duplicate work. Do you want to add testing? Otherwise, I would also love to.

BradyAJohnston:
@yuxuanzhuang if you want to add some new tests that would be great. Some of the 'blender-specific' tests are kind of weird to make. At the moment my strategy for testing different styles / nodes etc is to get the vertices of the final 3D object and use that as a snapshot to compare against. If you want to have a go at putting together some new tests that cover the new functionality, I can ensure the existing tests get fixed once yo have the new tests in place (feel free to fix them if you would like).

BradyAJohnston:
There remains a minor issue when enabling the add-on if you don't already have MDAnalysis installed.

```python
ModuleNotFoundError: No module named 'MDAnalysis'
```
Currently when the add-on is enabled, it loads all of the submodules in order before it can enable properly. Because of this, if a required package isn't installed, the enabling fails. Without being enabled, you can't install the packages. 

The solution I have been using currently, is to only `import MDAnalysis` inside the functions that have actually been using the packages, to ensure that the import isn't tried until the functions are called. That way it can enable and you can install before then using the functions. 

This normally isn't a problem for regularly installed python packages, but alas working in Blender makes it more difficult. 

It's problematic because it makes it hard to specify input and return types for functions, without having MDAnalysis imported globally in the file.

If you can come up with a better way to handle this problem, then that would ideal and would fix the problems that I am having for the other submodules. If we can't figure out a better way to do it, then we'll just have to move the imports into the functions / classes where it is being used, to stop it breaking when imported at the top of the files.

yuxuanzhuang:
See my new [commit](https://github.com/BradyAJohnston/MolecularNodes/pull/287/commits/ada9cacbb019baef9b81ef8739b3e4d83e706090) with [mocking](https://docs.python.org/3/library/types.html) MDAnalysis module. Should be applicable to other modules as well.

BradyAJohnston:
Thanks for figuring that out! Seems to work a treat. I think we'll have it just for MDAnalysis in this PR and I can create a separate one to update it for the other modules later. 

I've also fixed the style selection through the UI, which now works with the `styles_mapping`. It's an ENUM now which it really should have been initially so it's good to switch it over.

BradyAJohnston:
I can't for the life of me figure out why the docs keep failing to auto-build. I assume it might be just a permissions thing for this PR, but they are building fine for me and I will be able to deploy them myself later.

I've added a quick overview in the `03_moelcular_dynamics.md` file. If you want to add more detail, please do! I have just been uploading screenshots & videos to imgur and then linking to them, which seems to be the best stratergy.

yuxuanzhuang:
> I can't for the life of me figure out why the docs keep failing to auto-build. I assume it might be just a permissions thing for this PR, but they are building fine for me and I will be able to deploy them myself later.

Maybe try this? https://github.com/bobheadxi/deployments/discussions/78#discussioncomment-2224895

BradyAJohnston:
I generated a new PAT (is that the right one??) and added it as a secret to the repo for GHA, but it seems to get `null` when it access the token so I am beyond stumped at the moment. Going to bed so might figure something out when I wake up or if you can figure something out that would be amazing. 

Like I said, not the end of the world because it will work outside of this PR, just failing for this PR.

yuxuanzhuang:
No problem! Just revoked all the edits on the yml file.

BradyAJohnston:
I think this is ready to merge in to main. It seems to be fairly robust, and doesn't remove any features that were already there. There are lots of exciting things to build on top of it, but think this is good to get into main and we can create a release and maybe a new branch for the MDA hackathon.

yuxuanzhuang:
Thanks!!!

BradyAJohnston:
@yuxuanzhuang the thanks is all for you! thanks so much for working on this. it's going to enable a lot of very cool stuff in the future

PR #295, opened 2023-08-21
simple hack to load petworld bcif files

corredD:
Petworld cif files define each macromolecule to be instanced with model number. This hack merged all chain under the model number.


BradyAJohnston:
That seems to have totally fixed everything, thank you!

PR #317, opened 2023-09-29
Add frame interpolation for MDA Universes in Blender

BradyAJohnston:
Default behaviour when importing MDAnalysis objects, that each frame of a trajectory is mapped to a frame in Blender. 

This makes sense as a default, but the user might want to interpolate between two frames of a trajectory for the sake of animation. This PR adds the option for interpolating between two frames of the trajectory, based on `subframe` value. 

Mapping of frames and subframes, where the float component of the result is used to interpolate between frames. For `5.667` it interpolates between frame 5 and 6.

| frame| subframe = 0| subframe = 3| subframe = 5|
|--------|--------|--------|--------|
| 0   | 0 | 0 | 0 |
| 10 | 10| 3.333 | 2 |
| 20 | 20 | 5.667 | 4 |
| 50 | 50 | 16.667 | 10 | 

https://github.com/BradyAJohnston/MolecularNodes/assets/36021261/dc2bad3d-3455-4d99-85d9-bb0bc349ce2b


yuxuanzhuang:
I also have some codes for frame mapping (at home unfortunately) that might be a good plus to this!

BradyAJohnston:
@yuxuanzhuang great I'll leave this for now till we can combine forces. At the moment it creates a property on the object that can be altered, but creating some better UI for this would be good. 

I think it is getting to the point where we should create a new panel for interacting with molecules more generally, which I might create inside of the viewport rather than in the current import panel location.

BradyAJohnston:
@yuxuanzhuang did you have code you wanted to add to this? Otherwise I might clean it up a little and merge.

yuxuanzhuang:
Sorry for not responding for a while!  I was busy with my defense last week. I think I will create a separate PR for frame mapping based on the changes introduced in this PR.

BradyAJohnston:
@yuxuanzhuang no worries at all! I hope that the defence all went well :)

yuxuanzhuang:
I was actually asked to sell MolecularNodes as a complementary tool for trajectory visualization during my defense! I think I've managed to attract some more people to use this tool!

Regarding #331, I think the best approach would be merging it into your branch (if you like this addition ofc) and I can add more comments to it here.

BradyAJohnston:
@yuxuanzhuang that's good to hear! The more the merrier. Will hopefully have time to go over it and merge it all this weekend :)

BradyAJohnston:
@yuxuanzhuang I've gone through and titied up the subframes code a bit, but I seem to be breaking some tests in a way that I can't seem to figure out. 

All of the mda tests pass, except for the `in_memory` ones. Weird thing is that I can't find anything that I would have changed that would have affected the `in_memory` import.

CleanShot 2023-10-31 at 10 38 33

Okay maybe some minor position changes. If I check it all manually it seems to be fine. If I accept that and just try to update the snapshots with `--snapshot-update` then additional tests fail that weren't failing previously with the old snapshots??

CleanShot 2023-10-31 at 10 40 38

I've poked around but I can't for the life of my figure out what is going on.

yuxuanzhuang:
I will try to find some time and take a look!

BradyAJohnston:
Okay I think I managed to figure it out and fix them. The problem was that the `in_memory` was interpolating the 5 frames of the animation over 250 frames in Blender. To fix it I had to set it so that it would interpolate over the frames 0-4 like the not in memory version. I think I want to tweak how that node works in the future, but it should all be working now!

PR #323, opened 2023-10-02
Change dependency in pkg.py: from pkg_resources to importlib.metadata

### LINKED ISSUE #311: Move away from `pkg_resources`
Currently get this warning for the `pkg` submodule, when checking if packages are installed or not.

```
/Users/bradyjohnston/git/MolecularNodes/MolecularNodes/pkg.py:9: DeprecationWarning: pkg_resources is deprecated as an API. 
See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import get_distribution, DistributionNotFoun
```

Should move away from this submodule, as the API is deprecated.

ljcolson:
This pull request resolves #311 

The link in the DeprecationWarning (https://setuptools.pypa.io/en/latest/pkg_resources.html) recommends porting pkg_resources to importlib.metadata. That's why I changed the dependency to importlib.metadata. Importlib.metadata has a function called 'version' which performs the same as pkg_resources' `get_distribution(package).version`. To avoid naming conflicts I imported the function under the new name `get_version`. Similarly DistributionNotFound is now PackageNotFoundError 

This change works in my testing.

### DIFF
--- a/MolecularNodes/pkg.py
+++ b/MolecularNodes/pkg.py
@@ -6,7 +6,7 @@ import subprocess
 import sys
 import os
 import logging
-from pkg_resources import get_distribution, DistributionNotFound
+from importlib.metadata import version as get_version, PackageNotFoundError
 import bpy
 import pathlib
 import platform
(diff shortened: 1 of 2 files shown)


BradyAJohnston:
Thanks mate, this is exactly what I was after! I'll run some more tests but it all looks good to merge.

BradyAJohnston:
added a couple of tests just to slowly increase test coverage. There is still the same warning that also shows up during testing, but that is from `starfile` so not much we can currently do for it.

PR #332, opened 2023-10-26
Remove ESMFold

BradyAJohnston:
It seems that MetaAI has shut down the servers for ESMFold. This PR
removes the panel and tests related to it from Molecular Nodes.

- remove esmfold
- remove esmfold tests


BradyAJohnston:
@wjm41 I am unable to figure out if ESMFold has actually been shut down yet or not, or if something is just broken. The tests fail and I can't get anything to work, so I was going to remove it. Would you want to have a look at it first as you implemented it, before I remove it?

wjm41:
lemme have a look - I can still make curl requests and get pdbs back so must be something else

BradyAJohnston:
thanks @wjm41 , I might have written it off a bit early. I was under the impression Meta was shutting down the service, but I can't actually find any information about that anywhere so I might have been mistaken.

BradyAJohnston:
I'm going to remove it as it's wreaking havoc with the tests at the moment. If you get something working it can always be added back in before the next version is released.

PR #333, opened 2023-10-27
Save mda_session to the same location as .blend file

yuxuanzhuang:
Changed in this PR
- no uuid is added to the mda_session and mol_object
- The session file is saved in the same place as .blend file instead of ~/.mda_session/uuid.pkl

As discussed, it makes more sense to store .mda_session file in the same place as .blend file.


molecularnodes/mda.py
```diff
@@ -428,18 +424,17 @@ def __init__(self, world_scale: float = 0.01, memory: bool = False):
         self.universe_reps = {}
         self.atom_reps = {}
         self.rep_names = []
-        self.uuid = str(uuid.uuid4().hex)
```
BradyAJohnston, line 431: This now breaks some of the tests that were checking the `uuid`, can you fix them?

BradyAJohnston (commented):
looks good, should just need to fix some of the MDA-specific tests. (some of the others are currently breaking for an unknown reason, don't worry about them I'll handle it).

yuxuanzhuang:
Just back from my holiday! Tests should all pass now!

BradyAJohnston:
Welcome back from your much deserved holiday!

Issue #366, opened 2023-11-30
Error in conversion of certain euler angles when loading star file


jojoelfe:
**Describe the bug**
Some eulerangles are not converting correctly into blender coordinate system. It might be related to this bug, https://github.com/alisterburt/eulerangles/issues/8, in the eulerangles package used for the conversion. 

This can be fixed when falling back to the previous conversion code from 

```
target_metadata = ConversionMeta(name='output', 
                                   axes='xyz', 
                                    intrinsic=False,
                                    right_handed_rotation=True,
                                    active=True)
eulers = np.deg2rad(convert_eulers(euler_angles, 
                               source_meta='relion', 
                               target_meta=target_metadata))
```

to

```
eulers = R.from_euler(
        seq='ZYZ', angles=euler_angles, degrees=True
    ).inv().as_euler('xyz')
```

Happy to submit a PR, but noticed you are currently making the 4.0 transistion and refactoring a bit, so I can wait until this is more or less finished.

Thanks,

Johannes

**To Reproduce**
I pasted a diagnostic starfile below

**Expected behavior**
Currently this loads like this (notice incosnisten behaviour in central cross):

![image](https://github.com/BradyAJohnston/MolecularNodes/assets/6081039/0f108b39-fc5e-46e5-b853-4b8ef7278bf4)

while it should look like this

![image](https://github.com/BradyAJohnston/MolecularNodes/assets/6081039/151adf23-846c-4deb-a749-53847b2e5a33)

**Error Codes**
Please include associated error codes that are shown in the python console when the error occurs.
> To view the python console on Windows, `Window` -> `Toggle System Console`, on Mac you must start Blender from the terminal.
```
# paste error codes from python console here
```

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows / MacOS / Linux]
 - Hardware: [e.g. Intel Mac, M1 Mac, x86 PC]
 - Blender Version: 
- MolecularNodes Version:

**Additional context**

```
# Created by the starfile Python package (version 0.5.1) at 10:44:33 on 30/11/2023


data_particles

loop_
_rlnCoordinateX #1
_rlnCoordinateY #2
_rlnCoordinateZ #3
_rlnOriginXAngst #4
_rlnOriginYAngst #5
_rlnOriginZAngst #6
_rlnAngleRot #7
_rlnAngleTilt #8
_rlnAnglePsi #9
_rlnImagePixelSize #10
_rlnMicrographName #11
[...]

PR #371, opened 2023-12-01
Fix starfile imports

jojoelfe:
Hi Brady,

this does not address #366 yet, but I noticed that the relion import was broken, so I first went ahead and fixed things up before I started. What I did here is:

- Fix starfile import for both relion and cistem
- Add a relion starfile test to the pytest suite
- Moved the starfile nodetree into the `MN_data_file.blend` file instead of creating it by code
- Replaced the placeholder cone with colored axes, which makes debugging orientation easier

image

The things I would like to get you input on is:

- Version control and diffing of the `MN_data_file.blend` is difficult/impossible. Do you have some specific rules or strategies to keep it tidy?
- I have not linked up the modifier inputs yet, but I noticed you stopped doing that for other things as well and things now seem controlled by the "Object" and "Scene" tabs of the "MolecularNodes" panel. Is that what you want to do going forward?

Best,

Johannes


BradyAJohnston:
Thanks for the fix!

You are right that versioning the `.blend` file is basically impossible. So far it has mostly just been myself interacting with it, and I just have to have a bunch of tests that test the nodes inside of it for any changes. It it still a system that I'm really not happy with, but one that I don't know how to improve at this point.

The additional `Object` and `Scene` tab is an effort to just bring some of the settings that are hidden throughout Blender, all to one place, which should hopefully save users time and confusion about how to change some commonly tweaked settings. I have been setting it up so that the `style` node will be displayed inside of the object as well, so that users can interact with it through that, or through the node tree itself. I think moving the starfile node group to be within another node group is good, and is more in line with how the other node groups in MN work.

PR #374, opened 2023-12-04
Convert cryoEM eulerangles in GeoNodes

jojoelfe:
This addresses #366 by converting ZYZ eulerangles into blender rotations inside of a nodegroup. This removes the dependency on `eulerangles` and is not dependent on `scipy`. The correctness of the conversion is, however, tested against `scipy`. As such I added it as a dev-dependency.

The testing gave me a bit of trouble, so I'll explain it here. The main obstacle was that the rotation attribute of instances is not available from python (at least to my knowledge after ~30 min of researching). I therefore implemented a debug switch into the star nodetree, that when activated output the original points with the quaternion stored as a named attribute.

image

During testing this switch is activated, the quaternion values are read and compared against a rotation calculated from the euler angles by scipy. (Of course the conventions is opposite. `scipy` is x, y, z, w.  Blender doc claims to use the same. but in my test it produced w, x, y, z).

I also tried to hook up the starfile object to the "Object" Panel, but the object-input slot is broken. This seems to be a blender issue to me. I noticed that the color of the object socket changed in 4.0, maybe something is not hooked up yet...

image

### DIFF
--- a/molecularnodes/blender/nodes.py
+++ b/molecularnodes/blender/nodes.py
@@ -137,6 +137,20 @@ def get_style_node(object):
     group = object.modifiers['MolecularNodes'].node_group
     return style_node(group)
 
+def star_node(group):
+    prev = previous_node(get_output(group))
+    is_star_node = ("MN_starfile_instances" in prev.name)
+    while not is_star_node:
+        print(prev.name)
+        prev = previous_node(prev)
+        is_star_node = ("MN_starfile_instances" in prev.name)
+    return prev
+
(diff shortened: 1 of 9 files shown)


BradyAJohnston:
Thanks for doing this! 

Testing is tricky and at times problematic. My solution so far has been to realise the modifiers on the mesh to access GN attributes, but what you have done is better and doesn't destructively apply the modifiers. I will be going through and updating some of the testing myself with this as well.

PR #386, opened 2024-01-01
fix #385

### LINKED ISSUE #385: Selection > Multiple Residue Selection doesn't work as expected
When trying to select multiple residues using the Molecular Nodes > Selection > Multiple Residue Selection, the automatically generated node only selects the last residue id in the list. This is due to problems in the connections in the autogenerated node.

To reproduce, follow these steps:
1. launch Blender (there is no need to load a protein)
2. enter Geometry Nodes workspace and click "New" at the top of the Geometry Nodes pane to create a new group
3. in the Geometry Nodes pane, type A to add a new node, and select Molecular Nodes > Selections > Multiple Residue Selection
4. in the Multiple Residue Selection dialog, type "1,2,3" (no quotations). This generates a new node called "Select Res Id Custom" with three residues (1, 2, and 3) defining the selection.
5. Select the new node by clicking on it, then type tab to enter the node
6. It it clear that the selection connected to Group Output only contains the last selection

Rather, I would expect it to result in all specified residues to be selected. See image of the node logic that has been made to function as expected.

image
image
image
image

I'm using Blender 4.0.2 and Molecular Nodes 4.0.6 on Windows 10

P.S. this is *amazing* software and I deeply appreciate your work on this!

BradyAJohnston:
- fix for multiple residue selection
- test for creation of custom selection

### DIFF
--- a/molecularnodes/blender/nodes.py
+++ b/molecularnodes/blender/nodes.py
@@ -708,6 +708,7 @@ def resid_multiple_selection(node_name, input_resid_string):
     group_link = residue_id_group.links.new
     new_node = residue_id_group.nodes.new
     
+    prev = None
     for residue_id_index,residue_id in enumerate(sub_list):
         
         # add an new node of Select Res ID or MN_sek_res_id_range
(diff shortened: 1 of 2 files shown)


lesburg:
Thank you so much..and such a quick and simple fix!

PR #401, opened 2024-01-07
fix node setup and add more tests

BradyAJohnston:
Fixes node setup for density import. Additionally adds more tests with snapshots that tests use of the operator.


BradyAJohnston:
@jojoelfe I figured out how to run the tests involving `pyopenvdb` on my local machine. You use Blender's python, which has access to `pyopenvdb`. Weirdly it doesn't seem to have access to `bpy` or many other modules (I assume something to do with paths). 

But the following code installs `molecularnodes` and `bpy` etc (something weird might happen with installing `bpy` into Blender which already has `bpy`... but so far nothing has exploded. But doing this allows me to run the tests myself.

```
/Applications/Blender.app/Contents/Resources/4.0/python/bin/python3.10 -m pip install . -e
/Applications/Blender.app/Contents/Resources/4.0/python/bin/python3.10 -m pip install pytest
/Applications/Blender.app/Contents/Resources/4.0/python/bin/python3.10 -m pip install pytest-snapshot
/Applications/Blender.app/Contents/Resources/4.0/python/bin/python3.10 -m pytest tests/test_density.py
```

jojoelfe:
I've been copying the pyopenvdb.so file into the site-packages folder of the poetry python environment, exactly how it is done in the github actions. But I am only on Linux machines and don't know if this works on windows or a mac.

BradyAJohnston:
Okay that's good to know that it's doable on Linux. Probably a better approach than what I'm using on mac.

PR #402, opened 2024-01-07
Refactor: Change internals to use a Class-based parsing system

BradyAJohnston:
Working towards improving the API internals, this adds a more class-based system for parsing structures.

Currently just for the 'molecules', the `.pdb`, `.pdbx/mmcif`, `bcif` and `mmtf` files are all subclasses of `Molecule`. This gives a common input for creating a 3D model based on the `AtomArray` data. The resulting 3D created models should also be something, potentiall `MoleculeModel`s, which can then be aware of the data that created them, giving rise to a simpler way to update the models after they are loaded and other similar methods. 

This could then be expanded to importing via CellPack, starfile, densities etc.

Open to suggestions for how this should be approached.


jojoelfe:
This is great! I've been looking at some crystal structures lately and implementing better support for them would be much easier after this.

- I've written a (very simple) importer for the dsn6 format the PDB uses for crystal structures: https://github.com/jojoelfe/dsn6file . I guess for densities there could be a similar approach where there is a abstract BaseClass (That under no circumstance is called `ElectronDensity`) and different formats inherit from it, producing a vdb file.
- I've also been trying to import crystal packing information and display it. I wonder if crystal packing, biological assemblies, starfiles, and cellpack could all be unified under something like an `Ensemble` class that provides location/rotation information for the instancing.

BradyAJohnston:
thanks @jojoelfe for the input!

Great minds think alike. I had created the `Ensemble` (ensemble.py) but hadn't actually ported over the cellpack / starfile info yet.

I have been wanting to properly import crystallography / dsn6 files for a while, so support though that package would be great!

I think next I'll be doing the Ensemble. I want to still improve / refactor the the blender interface, but I think that might be best addressed in another PR.

PR #415, opened 2024-01-15
Add .mol file support

TheJeran:
Added .mol support


BradyAJohnston:
thanks for the PR! You're right this is simple enough. If you could add an example `.mol` and `.sdf` file to the `tests/data` directory, then add some tests for opening them, then it'll be good to go :)

TheJeran:
Alrighty. You can run those tests if you like. But it works. Didn't notice MOLFile also supports sdf so added that bit as well.

BradyAJohnston:
I expanded the tests to include a bunch of snapshots of the created 3D geometry. This is currently the only way for me to check that the nodes for creating the styles are working correctly. It generates a lot of files for the snapshots leading to large diffs, but it's the best solution that I have at the moment. 

Thanks for contributing this! Once these test pass I'll merge :)

PR #421, opened 2024-01-21
More topology related nodes

BradyAJohnston:
Adds another vector utilitiy for computing the angle of two vectors, while projected perpendicularly onto a thirds vector which is the axis (for computing torsion angles). 

Additionally adds nodes which do the Phi and Psi calculations for backbone torsion angles.

Additionally adds the 'pLDTT' color node, for AlphaFold coloring #391


BradyAJohnston:
@rbdavid unfortunately fort the `MN_data_file.blend` there is no merging of versions, so it has to just be one continuous history and I have to manually resolve conflicts myself. I'm looking in to better solutions but it's a major limiting factor at the moment

PR #422, opened 2024-01-23
update startup file

BradyAJohnston:
adds some initial props (backdrop) lights and switches colorspace to AgX


rbdavid:
After downloading the molecularnodes_4.0.9.zip file, the ~/molecularnodes/assets/template directory only contains a MolecularNodes.zip file. When I unzip it, a startup.blend file and a MolecularNodes directory are created. The directory contains a different startup.blend file; this latter version is the updated one. Just FYI. 

This directory structure/zip file is not mirrored in either of the source downloads.

BradyAJohnston:
@rbdavid thanks for the details. Some weird things seem to keep happening with templates, so I will look into it.

PR #446, opened 2024-03-04
Optionally display micrograph for starfiles

jojoelfe:
Hi Brady,

this is something I have been using for a while, but now have been trying to implement somewhat cleanly. It allows the user to display a micrograph corresponding to a starfile. Below is a short demo using the relion SPA tutorial data files:

[Screencast from 2024-03-04 14-36-20.webm](https://github.com/BradyAJohnston/MolecularNodes/assets/6081039/7fa16b2d-39ae-4622-9b67-fc01e5893832)

In order to do this micrographs are converted to tiff and loaded as an image texture when necessary. This is implemented by subscribing to the depsgraph update event.

Some notes about this:

- In order to keep merge conflicts to a minimum, I've moved starfile nodes and materials to a seperate data file that is in turn linked into the main data file. In the main data file there is only one nodegroup that acts as a "container", because objects that are only linked into the file cannot be appended.

- I added pillow as a dependency for the tiff conversion. I know this should not be done lightly, but it is already a dependency of mdanalysis I think, so it shouldn't add much most of the time. 

- In star.py I moved more code into methods of the Ensemble object

This is note quite ready yet, but I thought I should let you take a look before I clean things up.

TODOS:

- [ ] More tests to make sure micrograph is actually displayed with correct texture
- [x] Reestablish the depsgraph_update event upon loading a file
- [x] Test more on tomography data. With Blender 4.2 alpha one can use the "Sample Grid" node to make nice slices, but not sure when this will hit a stable release.


BradyAJohnston:
Hi mate! Thanks so much for putting this together. It's looking pretty good, and being able to import the micrographs as well will be great.

Blender 4.2 will be bringing a whole lot of great grid / density nodes which will make a lot of this easier and nicer - that release won't be until June though.

>In order to keep merge conflicts to a minimum, I've moved starfile nodes and materials to a seperate data file that is in turn linked into the main data file. In the main data file there is only one nodegroup that acts as a "container", because objects that are only linked into the file cannot be appended.

I like this idea a lot, of splitting out the different sections into their own `.blend` files so that smaller changes don't result in just the entire file being versioned (and like you said, less version conflicts). Can you explain some more about the container node group though - I'm a little unsure about what is happening there with linked files etc.

I know this is an area of assets that are a high priority for the devs to improve. Currently if you use two different node group assets that both utilise an underlying node, then you get duplicates of the underlying node. They have plans to improve all of that and reduce conflicts etc, which should also help us as well. 

>I added pillow as a dependency for the tiff conversion. I know this should not be done lightly, but it is already a dependency of mdanalysis I think, so it shouldn't add much most of the time.

I'm fine with this. Outside of the 'primary usage' which would be downloading structures from the PDB and opening `.pdb/mmcif` files which is all handled by `biotite`, I'm happy to add an extra package or two for dependencies for other import methods which I would consider to be extensions on the base functionality. An extra button click during install is fine currently.

> In star.py I moved more code into methods of the Ensemble object

If you can think of other improvements / changes to this class as well I am open to that. The overarching workflow still has lots of room for improvement.

jojoelfe:
> Can you explain some more about the container node group though - I'm a little unsure about what is happening there with linked files etc.

My original thought was to move all the NodeGroups and materials into the new starfile. 



And then link them into the main data file, where they now appear as linked items:



However, these cannot be directly appended into a new file using `bpy.ops.wm.append` because they are not actually in the main data file. Therefore I created a nodegroup in the main data file called `MN_starfile_components` that contains `MN_starfile_instances` (which in turn depends on all the other nodes and materials). Then by appending `MN_starfile_components`, `MN_starfile_instances` and the other nodegroups and materials are appended as well. 

> If you can think of other improvements / changes to this class as well I am open to that. The overarching workflow still has lots of room for improvement.

I'll think about it. The main things that I think about are: 

(1) What are the differences between the MN Object and the object in blender? Can there be cases where they contain conflicting information and which one is correct in that case? 

(2) Is there a way to restore the MN object when loading a blend file? I tried putting it as an attribute (python attribute) on the blender object, but blender only  allows certain types.

BradyAJohnston:
Okay I've had a play around with appending linked node groups etc and I understand what is going on.

I think having to have a node group which is just there to wrap others so they can be appended will get messy. If we are going to split node groups into different `.blend` files, we should just do that and not having to worry about having them all 'indexed' in the same `MN_data_file.blend`.

In this case, if the node starts with `MN_starfile_*` then we just look for it in the `MN_starfile.blend` or something similar. 

Some of the utility nodes will be linked to from different files, but I think splitting them all into their own separate files might be the best strategy going forward.

BradyAJohnston:
> (1) What are the differences between the MN Object and the object in blender? Can there be cases where they contain conflicting information and which one is correct in that case?

I am unsure what you mean by this. So far at least whenever I am referencing an `object`, it is a `bpy.types.Object` object inside of Blender (so your standard blender object).

> (2) Is there a way to restore the MN object when loading a blend file? I tried putting it as an attribute (python attribute) on the blender object, but blender only allows certain types.

For this are you meaning the `Ensemble` or `Molecule` objects? This I am unsure of. @yuxuanzhuang got it working to reload MDAnalysis `Universe` objects by dumping them beside the `.blend` file on save, then loading them back in: 

https://github.com/BradyAJohnston/MolecularNodes/blob/f3dbff0f4dcc48e0f9429f74324e1b5cd940534f/molecularnodes/io/parse/mda.py#L895-L970

I'm not sure what the best approach would be here - I'm certainly a bit out of my depth with this kind of thing.

jojoelfe:
> I am unsure what you mean by this. So far at least whenever I am referencing an `object`, it is a `bpy.types.Object` object inside of Blender (so your standard blender object).

Yes, I am sure there are proper words for this, but I'm also not the greatest expert. For example for the new `StarFile` class that inherits from `Ensemble`, there is the `.data` property, which contains a table with the position and rotation of all molecules, but there is also the data in blender object. And I think there might be many scenarios where one might want to change those, but it is unclear, which one. For example, I might want to write some python script that imports a starfile like this

```python
ensemble = parse.StarFile(file_path)
ensemble.create_model(name=name, node_setup=node_setup,
                      world_scale=world_scale)
```

If I know change the .data property like this:

```python
ensemble.data = ensemble.data[ensemble.data["Score"] > 10.0]
```

this of course will not change the actual data in blender unless we call a function that copies this into blender. On the other side, I might go into Edit mode in blender and start deleting points or in the future use some MN-Tool to change rotation of points interactively, this will not update the `ensemble.data` object.

This might be fine and we could decide that these might be different and it is up to the user/programmer to sync them as they like or we might make the decision that one of them is always the correct one and the other one should not be touched. There is also the option to do something fancy with getter/setter in python to automatically update the blender data if the `ensemble.data` property is changed.

BradyAJohnston:
> this of course will not change the actual data in blender unless we call a function that copies this into blender. On the other side, I might go into Edit mode in blender and start deleting points or in the future use some MN-Tool to change rotation of points interactively, this will not update the ensemble.data object.

I don't know what the best implementation for this will be. For the time being I've got them all there for convenience. I think the idea that changing the 'data' should update the data displayed in Blender as vertices I think is probably the 'best' implementation. Like you said it'll take some fancy get / set work. I don't know how you would then handle someone manually editing objects in edit mode.

The way I am thinking about it currently - is that if someone is manually interacting with objects via the GUI, then they likely won't (or maybe _shouldn't_) be interacting with the objects at the same time via the python API. It may just be "this is the way it should be used, and outside of this way is off-label and anything could happen".

BradyAJohnston:
I'm certainly very open to suggestions about implementation design. The more I am thinking about it, the more I think that we should probably not have the `self.data`, and instead have a `self.as_array()` function that could return data frame extracted from file, or if an object exists it takes the data from the object and returns an array. 

Even this seems a bit messy though. I did experiment with doing `Molecule['b_factor'] = ...` being possible for getting and setting. 🤷

jojoelfe:
Ok, for now what I'll do is switch the appending to the way you suggested here: 


> In this case, if the node starts with `MN_starfile_*` then we just look for it in the `MN_starfile.blend` or something similar.

Maybe the `append` method  could first `try` importing from the main file and if that fails `try` importing from the specialized file.



> For this are you meaning the `Ensemble` or `Molecule` objects? This I am unsure of. @yuxuanzhuang got it working to reload MDAnalysis `Universe` objects by dumping them beside the `.blend` file on save, then loading them back in:

I'll try to do something similar with the StarFile class to re-enable the micrograph change events when loading a file.

BradyAJohnston:
> I'll try to do something similar with the StarFile class to re-enable the micrograph change events when loading a file.

If it ends up being a more general purpose maintaining of `Molecule/etc` objects then we should maybe have that in a separate PR. If you can't get something reliably working - then maybe just an option that appears in the panel to re-link the images to the object that the user will have to run when reloading the file (less ideal though).

> Maybe the append method could first try importing from the main file and if that fails try importing from the specialized file.

Once this PR is merged I can start moving the other categories out to their own separate files as well.

jojoelfe:
Hi Brady,

I think this might be good to go. Some more explanations on what I've done:

- Similarly to the MDAnalysis setup there is no a callback during initialization of MN that will reestablish callback for starfiles. I named in "rehydrate" and not "rejuvinate", not quite sure if either is the right word. I guess "deserializing" would be the other option. Let me know if you have opinions on that.
- To enable the "rehyadration" the starfile class now has to classmethods to act as constructors `from_starfile`, which is used when a new starfile is being parsed, and `from_blender_object`, which is used during the rehydration. 
-  The only change to the MN_data_file should now be that the starfile NodeGroups have been deleted.
- I do think it would be nice to have more tests, but I think for that I should compile a better list of starfiles "out there" (relion 3 vs 4 vs 5, STA vs SPA, etc..) and that might be it's own PR.

Johannes

molecularnodes/blender/nodes.py
```diff
@@ -237,6 +250,16 @@ def material_default():
 
     return bpy.data.materials[mat_name]
 
+def MN_micrograph_material():
```
BradyAJohnston, line 253: This currently doesn't append the material?

molecularnodes/blender/nodes.py
```diff
@@ -237,6 +250,16 @@ def material_default():
...
+    """
+    Append MN_micrograph_material to the .blend file it it doesn't already exist, 
+    and return that material.
+    """
+
+    mat_name = 'MN_micrograph_material'
```
BradyAJohnston, line 259: Additionally I've started moving towards an emphasis on 'human readable' names for node groups and materials etc.

It's not _ideal_ for programming them, but they are only ever reference as strings anyway. The default material is called 'MN Default' now - so I think going forward we try to keep things named for people searching and reading them. Maybe `MN Micrograph`, as it is a material group so we won't need to reiterate that

molecularnodes/io/parse/star.py
```diff
@@ -19,90 +56,166 @@ def _n_images(self):
...
                       'rlnCoordinateZ']].to_numpy()
             pixel_size = df['rlnImagePixelSize'].to_numpy().reshape((-1, 1))
-            xyz = xyz * pixel_size
+            self.positions = self.positions * pixel_size
             shift_column_names = ['rlnOriginXAngst',
                                   'rlnOriginYAngst', 'rlnOriginZAngst']
             if all([col in df.columns for col in shift_column_names]):
                 shifts_ang = df[shift_column_names].to_numpy()
-                xyz = xyz - shifts_ang
+                self.positions = self.positions - shifts_ang
```
BradyAJohnston, line 88: could be `self.positions -= shifts_ang` ?

molecularnodes/blender/nodes.py
BradyAJohnston: This is only really for user-facing components like node groups, internal components I think we stick to cleaner naming but when users are searching or reading, not having the underscores helps

molecularnodes/blender/nodes.py
BradyAJohnston: additionally I think it could be renamed to `material_micrograph()` to align with `material_default()`

molecularnodes/io/parse/star.py
```diff
@@ -19,90 +56,166 @@ def _n_images(self):
...
+                else:
+                    raise FileNotFoundError(f"Micrograph file {micrograph_path} not found")
+            micrograph_path = pot_micrograph_path
+
+        tiff_path = Path(micrograph_path).with_suffix('.tiff')
+        if not tiff_path.exists():
+            with mrcfile.open(micrograph_path) as mrc:
+                micrograph_data = mrc.data.copy()
+
+            # For 3D data sum over the z axis. Probalby would be nicer to load the data as a volume
```
BradyAJohnston, line 147: This is something that could be address in a new PR maybe

molecularnodes/io/parse/star.py
```diff
@@ -19,90 +56,166 @@ def _n_images(self):
...
-        object = bl.obj.create_object(
-            xyz * world_scale, collection=bl.coll.mn(), name=name)
-
-        object.mn['molecule_type'] = 'star'
-        object.mn['star_type'] = star_type
-
-        # create the attribute and add the data for the image id
-        bl.obj.set_attribute(object, 'MNImageId', image_id, 'INT', 'POINT')
+    def create_model(self, name='StarFileObject', node_setup=True, world_scale=0.01):
+        import bpy
```
BradyAJohnston, line 187: these imports can go to the top of the file. The `bpy` is already imported there

BradyAJohnston (changes requested):
Left some minor comments but otherwise looking pretty good, LGTM

molecularnodes/blender/nodes.py
jojoelfe: This gets imported automatically when the MN_starfile_instances nodetree gets loaded, since it's being used there. I had some loading code there, but (I) the testcoverage report complained that this is never executed and (II) I would have to replicate that logical to import from the right data file. So I just deleted it and things work fine. 

This material should be somewhat hidden from inexperienced users anyway, as it get modified by events, which might have hard to understand consequences.

BradyAJohnston:
I've just been having a play around with the micrograph import - when importing the `cistem.star`, I can include the image - but I am unsure about the default settings for the micrograph. I am unfamiliar the data so it might be obvious - but why is the pixelsize negative? What coordinate system does this map to? Is there are reason the micrograph starts at -10?

CleanShot 2024-03-21 at 11 26 10@2x

Can you add tooltips for these inputs so that they can show up in the documentation (this will be auto-generated from the nodes like this https://bradyajohnston.github.io/MolecularNodes/nodes/style.html).

molecularnodes/blender/nodes.py
BradyAJohnston: that's fair. yeah it's annoying with the test coverage not liking certain aspects like the UI components that are hard to test.

BradyAJohnston:
Also there is a data point int he `cistem.star` that is _way_ up to the top. I figure it might be some kind of outlier bug, but I think we should remove it from the dataset for testing, as when you do `View Selected` on the object it does a massive zoomed out view.

CleanShot 2024-03-21 at 11 34 51@2x

jojoelfe:
> I am unfamiliar the data so it might be obvious - but why is the pixelsize negative? What coordinate system does this map to? Is there are reason the micrograph starts at -10?

Ah yes, that should be better documented. 

When pixel-sizes are negative the pixel-size is read from the starfile, but for positive values this is overwritten. This made sense to me, because in a lot of our software one enters "-1" to get automatic determination, but in this context it might be better to have a checkbox "Manually set pixelsize" or something.

The Z-position of the micrograph is not really defined as it is a projection along Z. -10 just ends up putting it under most of the instances most of the times, which is what you want. Since the Z-position of the instances is normalized so that the mean is 0, if it were set to 0, half of the instances would be hidden. The more elegant way would be to set the Default Z- position to the minimal Z of the instances.

BradyAJohnston (approved):
Thanks so much for all the hard work!

BradyAJohnston:
Thanks for all the hard work on the PR. 

I'd like to get us to focus on improving the documentation for the overall package. Would you be able and willing to write a tutorial on how to import starfiles? It would similar to the other tutorials (https://github.com/BradyAJohnston/MolecularNodes/blob/main/docs/tutorials/03_molecular_dynamics.md).

It can be a separate PR though - and otherwise I can work on it when I am going through and updating documenation for 4.1

Issue #459, opened 2024-03-29
Load local PDB File using the python API.


kainszs:
Hi, as part of the GSoC I wanted to get to grips with the api and ran into some problems. 

**Describe the bug**
Loading local pdb files or pdb files from the pdb database with the python api.

**To Reproduce**

1. created a fresh conda environment with python 3.10 like asked in the project.tml
2.  cloned this repo and installed it via `pip install -e  .`
3. created a python file containing the following lines:

```python
import molecularnodes as mn
mn.io.local.load("tests/data/1cd3.pdb")
```

**Expected behavior**
I expect it to load the file, like blender does.

**Error Codes**
```python
File [c:\Users\<...>\MolecularNodes\molecularnodes\io\parse\molecule.py:596]
--> [596](file:///C:/Users/<...>/MolecularNodes/molecularnodes/io/parse/molecule.py:596) mol.mn['molcule_type'] = 'pdb'
AttributeError: 'Object' object has no attribute 'mn'
```

If I just comment it out, I encounter the next bug:

```python
File [c:\Users\<...>\MolecularNodes\molecularnodes\io\parse\molecule.py:295]
 
[294](file:///C:/Users/<...>/MolecularNodes/molecularnodes/io/parse/molecule.py:294) 
def __repr__(self):
   return f"Molecule({self.data})"

AttributeError: 'PDB' object has no attribute 'data'
```

And the PDB has no attribute data, although the ABC molecule uses this attribute extensively.

After that, I tried to load an example from the pdb database:

```python
import molecularnodes as mn
mn.io.wwpdb.fetch("1cd3")
```

exception:

```python
File [c:\Users\<...>\MolecularNodes\molecularnodes\io\wwpdb.py:37]

[29](file:///C:/Users/<...>/MolecularNodes/molecularnodes/io/wwpdb.py:29) model = molecule.create_model(

[37](file:///C:/Users/<...>/MolecularNodes/molecularnodes/io/wwpdb.py:37) 

model.mn['pdb_code'] = pdb_code
model.mn['molecule_type'] = 'pdb'

AttributeError: 'Object' object has no attribute 'mn'
```
The same AttributeError as above. I checked this function in Blender and it works properly. Do I miss something?

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Hardware: Intel i9
 - bpy version 4.0.0
- molecularnodes version 4.0.12

kainszs:
[...]

PR #462, opened 2024-03-30
Clean up molecule class

kainszs:
Fix Issue: #459

Changes:
- remove template functions, which throw errors
- suggest a __repr__ method, which is especially for notebooks a nice to have


kainszs:
If we tackle the molecule class, I will be happy to write the tests for it as well

BradyAJohnston (approved):

Issue #455, opened 2024-03-26
Use an MDA Universe as a "data twin" to the blender objects


rbdavid:
This is a bit inspired by the idea of a "digital twin" for computer-aided manufacturing/architecture/etc but is a bit different. So here goes:

Currently, once the MolecularNodes instantiated blender object(s) are created, the associated Molecule object(s) will disappear at EOF for a script or upon killing the python interpreter panel. The Molecule object(s) are never accessible if the gui panels are used to create the blender objects. The blender objects as the final output from MN's IO and file parsing architecture. This blender object is the "visualization twin".

Currently, physical attributes associated with atoms are encoded to the blender object using biotite or MDAnalysis, depending on input files. These modules are used to gather atomic positions as well as other physical properties and encode them into the blender Object's vertices' attributes. A few of these vertex attributes, such as `is_backbone`, `is_solvent`, etc, are boolean arrays denoting regularly used atom selections. But, there are instances where more dynamic selections might be desired. Getting the indices associated with atoms for complex or dynamic selections can be challenging without the use of MDAnalysis, mdtraj, biotite, or similar libraries. 

It could be really useful to have access to, for example, a MDAnalysis Universe object even after the blender object has been created. This "data twin" to the blender object could mirror all relevant changes to the blender object's attributes, such as updated atomic positions after translation/rotation. But maybe more helpful, MDA's atom_selection() could be used to dynamically select atoms, gather those atom indices, and create an attribute for the blender object's vertices that can be used in the MN geometry nodes to visualize specific selections. 

Something like this "data twin" implementation would bring blender/MN up to par with pymol/VMD that have their own built-in atom selection->visualization capabilities.

rbdavid:
I haven't dug into the implementation of the MDAnalysisSession object (in `molecularnodes/io/parse/mda.py`) and its associated methods. That might have all of the functionality that I'm hoping for. 

I am a bit worried that the MDA session is dumped to a pickle file. Although very useful for this and many other usecases, pickle files have pretty serious security vulnerabilities. So implementing any functionality around dumping/loading pickle files should be handled very carefully.

BradyAJohnston:
[...]

Issue #440, opened 2024-02-26
Add trajectory smoothing when importing via MDAnalysis


BradyAJohnston:
Support potentially trajectory smoothing when importing an MD trajectory via MDAnalysis. (https://docs.mdanalysis.org/stable/documentation_pages/transformations/positionaveraging.html)

Potential for in theory to add enable / disable support for a variety of transformations. Some of these will be best implemented through Geometry Nodes node groups, but when streaming the trajectory from disk, transformations like averaging will be best done via MDAnalysis as Geometry Nodes won't have access to any frames other than the one that is currently loaded.

BradyAJohnston:
Seems that it might be a bit more complex, as we can't directly alter a transformation after it has been applied (https://github.com/MDAnalysis/mdanalysis/discussions/4477#discussioncomment-8669043). Potential for us to have a panel for the trajectory to add or alter a transformation, but that it would require a bit of load time as the universe is re-loaded.

Dan-Burns:
Smoothing frames would be great. I don't know the blender python API well enough but it's easy to take the coordinates from two frames you want to interpolate between and make the intermediate coordinate arrays.

BradyAJohnston:
Subframes that interpolate between frames of a trajectory are already implemented and can be accessed via the 'Object' tab of the 'Molecular Nodes' panel after importing a trajectory. As you increase the number of subframes, the number of interpolated frames in between are increase.

![image](https://github.com/BradyAJohnston/MolecularNodes/assets/36021261/6c45da77-54e5-46eb-9011-2a85a45a5c93)

This enhancement is more about utilising the trasnformations inside of MDAnalysis, which can load several frame (say 5 at a time) and then average their coordinates to create a smoother overall trajectory, rather than just interpolating between existing frames of a trajectory

Dan-Burns:
That's great - glad I know that now!

PR #466, opened 2024-04-01
Adds 'Live' MDAnalysis Selections

### LINKED ISSUE #475: 'Live' MDAnlaysis session loses connection upon object name change
If you change the name of a trajectory that has been imported via MDAnalysis, the connection between the object and the Universe it is related to is lost if the name of the object is changed. 

I believe this is because the list of AtomGroups are stored on the scene level, with reference to each object that is associated with the atom group. I believe we should change this so that each object / topology object has a reference to the universe that it is related to. The atomgroups / universes are in the list, and can be removed / deleted when an object is deleted. This also fixes the problem of not being able to 'look up' the universe that an object is getting its coordinates from. Currently you have to loop through all of the atom groups / universes and check for the names of objects. This then breaks if the objects themselves are renamed (which is often done), leading to the current problem.

BradyAJohnston:
Completely re-works how MDA universes are added and updated under the hood, while adding a panel for MDA selections. These are 'Live' selections which are updated when the selection string is updated and on frame change. Now that the universe linking is working really well - the 'in memory' option is removed.

Fixes #475, #455 

Enables functionality to implement #440 

The selections appear as boolean attributes on the mesh. It is a True / False for whether the atoms are selected via `universe.select_atoms(selection)` to the imported universe.

New features: 
- Live selections panel
- Options for enable / disable the updating of the selection, as well as being periodic for geometry selections
- Whether to interpolate when using subframes. Non-interpolated means the trajectory skips every _n_ frames before moving onto the next one.

Todo: 
 - [x] Implement `MNSession` for tracking imported molecules / universes / ensembles
 - [x] change tests to work with new updating system
 - [x] Handle deletion of custom selections so they remove the attribute also
 - [x] When adding and deleting selections, update and remove the corresponding attribute on the object / mesh
 - [x] Have a UUID based linking between blender objects and universes for updating.
 - [x] Add tests for updating selections

### Selections

The selections are defined by creating an `AtomGroup` and checking which universe atoms are inside of that group. Only on an update to the `selection_str` is a new `AtomGroup` created. For regular frame changes the same group is used, and the selection will updating internally if `updating` is selected.

- `name`: Name that will become the attribute name on the object
- `selection_str`: used inside of `mda.Universe.select_atoms(selection_str)` for defining the atom group
- `updating`: whether the selection is updated on change of frame or change of `selection_str`
- `periodic`: for geomtric selections whether to look into different periodic images when searching

Proximity selections are _a lot_ faster inside of Geometry Nodes / Molecular Nodes - so combining a selection with the proximity selection node is the fastest way to achieve distance selections. Need to improve proximity selection node related to #524 

### `MNSession`
Class for tracking the imported objects in the scene. Each dictionary for the correpsonding items uses a `uuid` that is generated when imported - which is also stored on the corresponding blender object when created. This allows for cleaner lookup of objects and universes, without relying on the `bpy.data.objects['name']` approach which is very easy to break (a frustrating limitation of Blender).

There is a dictionary for each type of item, (`Molecule`, `MNUniverse` and `Ensemble`). They are then available to the session after import is finished. The session won't persist between `.blend` save and load, so on save we pickle the file and save it next to the `.blend` file as `.blend.MNSession`. When loading a `.blend` file the pickle is loaded again and everything that was in the old session is appended to the new session. You can't pickle Blender's Object or Collection objects, so just before pickling a reference to their names are saved and then on load they are retrieved using the reference.

This session system now handles refreshing for universes, molecules and ensembles, so that it is all united in one location. 

https://github.com/BradyAJohnston/MolecularNodes/assets/36021261/e5e163ce-a6bd-4af1-a00c-5e4f21ad05d8


BradyAJohnston:
@rbdavid this is an example of using the connected MDA Universe that is maintained between sessions. This won't be applicable to other import methods, but a similar approach could be done.

rbdavid:
This is awesome! From the video, it seems like you can add any number of selection atom groups where `custom_selection` is the named boolean attribute associated with the selection string. So this should enable a user to make multiple selections and use different materials, representation styles, etc to highlight those. Awesome work! Let me know if you want a second pair of eyes on the code.

BradyAJohnston:
Yes the idea is that you can add any number of 'custom selections'. Each listed custom selection becomes a named attribute on the mesh, with a boolean selection that is generated through MDAnalysis. This can then be used inside of geometry nodes for styling / selecting etc.

BradyAJohnston:
Related to #440 - I'm thinking we could manage universe transformations (smoothing, translations etc) from a similar list.

BradyAJohnston:
@rbdavid if you did want to have a look over the code - I'm pretty happy with how this implementation is now. I explain mostly in the edited PR, but along with adding the 'live' selections - this has now completely overhauled how MDA Universes are imported and tracked in a global `MNSession` object.

rbdavid:
Looks great! Sorry for the radio silence from me. I've had many distractions pushing my blender work to lower priority. I'd like to get back to finishing my visualizations as well as helping in development where possible. Hopefully things clear up for me soon. 

I likely won't be able to test out the `MNSession` implementation for at least a week, maybe longer. Also, I'm still stuck on v4.0.9 or so for my visualization work -- whomp whomp. But from the video, the selection strings look intuitive and helpful.

One question, and this might show how out-of-date my knowledge is, the `MNSession` is only available when the user imports a structure via the MD trajectory import field? Importing from a .pdb or .mmcif file still uses biotite?

BradyAJohnston:
No worries @rbdavid . I think I'll merge this into the #501 branch and if you wanted to do a more broad review there that would be useful.

Downloading from the PDB, opening `.pdb` / `.mmcif` files are still handled by `biotite`, but they are also store & restored via the `MNSession` object under `MNSession.molecules`. Now that I've figured out how to reliably load, save and restore the session, we can track any arbitrary extra python objects. I haven't implemented any extra features for these other `Molecule` objects yet, but this enables all sorts of functionality that would rely upon the structure file that is currently not possible to implement in geometry nodes.
```py
class MNSession:
    def __init__(self) -> None:
        self.molecules: Dict[str, Molecule] = {}
        self.universes: Dict[str, MNUniverse] = {}
        self.ensembles: Dict[str, Ensemble] = {}
```

Issue #464, opened 2024-03-31
Optimize molecule class


kainszs:
As described in the issues #454 and #459, the molecule class can be optimized and unified with the AtomGroupInBlender class.

Goals:

- [x] clean up s. #454
- [x] write `molecule` class, with atomarraystruc as data object
- [x] write `molecule` class, with mda.atomgroup as data object
- [x] write `MoleculeInBlender` class

Further steps:
- test implementation
- optimize code, reduce redundancies
- integrate in code base

> Regarding the `Molecule` class. Currently when structures are imported, a 3D model is created inside of Blender. The idea is that the `Molecule` will contain information about the file that this model came from, and provide an interface that the user can change the associated style and edit the underlying information of the structure through this class. The currently existing python API isn't setup well for interacting with molecular structures, or changing the Geometry Nodes tress that are associated with the models. The class will hopefully provide a neater interface for loading, editing and styling the model, in a way that is nice to script with inside of a Jupyter Notebook or similar script.

_Originally posted by @BradyAJohnston in https://github.com/BradyAJohnston/MolecularNodes/issues/459#issuecomment-2028547718_

kainszs:
@BradyAJohnston would you assign the issue to me?

BradyAJohnston:
I just opened a PR with a new feature for 'live' MDA selection strings that update on each frame change in Blender #466 

This is the sort of thing that I envision for adding more connections between Blender and the MDA Universe underneath, to enable more user-friendly interaction with it.

kainszs:
That looks very cool. How does the data twin (#455) play into this? I would also like to implement it. Because that fits in with the molecule class?

rbdavid:
Firstly, my understanding of MN may be flawed so I defer to Brady. 

But, the `MoleculeInBlender` class you are describing is already implemented as the `bpy.types.Object` created from the `Molecule` object. This blender object has data associated with the structure stored in verts and edges associated with atoms and bonds. These verts/edges are used to draw the meshes seen in the Blender viewport. The (bio)physical/chemical information associated with verts/edges are stored as attributes. In my analogy to a "data twin", the Blender object is the visualization twin so adding a MN class for `MoleculeInBlender` isn't necessary. 

[...]

PR #469, opened 2024-04-01
Molecule class

kainszs:
Proposal for Issue #464

Changes:
- abc for molecules
- molecule classes that want to inherit from it require two methods: 

```python
@classmethod
@abstractmethod
 def _get_structure(cls, fpath: str):
      ...
 @abstractmethod
  def _assemblies
```

- Every annotation in the AtomArrayStack is dynamically set as an attribute to the molecule class. E.g. PDB(Molecule) has the extra field `sec_struc`.
- the handling of the molecule object remains the same, but the annotations, which are set as class attributes, can be used to easily create masks that can be applied to the blender object.

- MoleculeInBleder class
  -  [ ] simplified setting of attributes (every attribute of the molecule class can be set as attribute in the MIB class.)

### DIFF
--- a/molecularnodes/blender/molecule.py
+++ b/molecularnodes/blender/molecule.py
@@ -88,27 +88,27 @@ class MoleculeInBlender:
 
 
     
-    #def __setattr__(self, name: str, value: Any) -> None:
-    #    """
-    #    Set attr to the instance of the class, if the attr is in the __slots__, 
-    #    else to the molecule object of the instance.
-    #    """
-#
-    #    if name in self.__slots__:
-    #        object.__setattr__(self, name, value )
-    #    else:
-    #        set_attribute(object=self.molecule, name=name, data=value)
-    #
-    #def __getattr__(self, name):
-    #    """
-    #    Get attr from the instance's obejct of the class, if the attr was not found in the __slots__ by the __getattribute__ method.
-#
-    #    Raises:
-    #    ---
-    #    AttributeError: If the attribute is not found in the instance of the class nor in the molecule object.
-    #    """
-#
-    #    return get_attribute(object=self.molecule, name=name)
+    def __setattr__(self, name: str, value: Any) -> None:
+        """
+        Set attr to the instance of the class, if the attr is in the __slots__, 
+        else to the molecule object of the instance.
+        """
+
+        if name in self.__slots__:
+            object.__setattr__(self, name, value )
+        else:
+            set_attribute(object=self.object, name=name, data=value)
+    
+    def __getattr__(self, name):
+        """
+        Get attr from the instance's obejct of the class, if the attr was not found in the __slots__ by the __getattribute__ method.
+
+        Raises:
+        ---
+        AttributeError: If the attribute is not found in the instance of the class nor in the molecule object.
+        """
+
+        return get_attribute(object=self.object, name=name)
     
     def list_attributes(self, evaluate=False) -> list | None:
         """
--- a/molecularnodes/blender/molecule.py
+++ b/molecularnodes/blender/molecule.py
@@ -2,12 +2,15 @@ from typing import Optional, Any, Union, List, Tuple, Dict
 import warnings
 import time
 import numpy as np
+from functools import singledispatchmethod
+
 import biotite.structure as struc
 from biotite import InvalidFileError
 
-import bpy
+import bpy 
 
-from ..io.parse.molecule import Molecule
+from ..io.parse.molecule import MoleculeAtomArray
+from ..io.parse.mda import MDA
 
 from .obj import set_attribute, get_attribute, create_object
 from .nodes import assembly_insert, create_starting_node_tree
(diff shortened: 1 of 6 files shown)


kainszs:
It doesn't seem to be that easy to add a new attribute to the blender object. I'll read up on it tomorrow

jojoelfe:
> It doesn't seem to be that easy to add a new attribute to the blender object. I'll read up on it tomorrow

There are only a few things you can add as an attribute (floats, ints, strings...). For the starfile class I only save the filename of the starfile as an attribute on the blender object, which is all that is needed for "rehydration".

https://github.com/BradyAJohnston/MolecularNodes/blob/12fb08d3e2709797da892b908f8171b904d50fae/molecularnodes/io/parse/star.py#L6-L46

https://github.com/BradyAJohnston/MolecularNodes/blob/12fb08d3e2709797da892b908f8171b904d50fae/molecularnodes/io/parse/star.py#L214

kainszs:
So it is not possible, to store an additional list?

BradyAJohnston:
Blender can be very particular about extra things you can assign to blender objects. You can assign some arbitrary python-defined things like dictionaries, but more complex python objects like numpy arrays will cause a hard crash.

```py
# blender object (bob)
bob = bpy.data.objects['Default Cube']
bob['anything'] = {'hello': 'world'}
bob['a_list'] = ['foo', 'bar', 'baz']
```

More 'officially supported' options are any of the `bpy.props.*` (https://docs.blender.org/api/current/bpy.props.html)

These 'arbitrary' properties can be used inside of python, but they can't reliably be displayed in the UI as they are just python objects and not 'official' blender properties (which can be displayed / interacted with via the UI).

For setting information on the 3D objects themselves (for each vertex etc) you can use the `molecularnodes.blender.obj.set_attribute()` which use the associated blender methods for setting attributes.

kainszs:
Please note that this is only a rough suggestion and that there is still a lot of room for improvement. Above all, it is intended to show that it is possible to streamline the workflow in PR #469.

It is possible to create the same `MoleculeInBlender` object from two different `Molecule` classes, which follow the following pattern:

`Molecule(ABC)` -> `MoleculeAtomArray(Molecule)`->`PDB`, and others
         |
`MDA(Molecule)`

The abstract `Molecule` class, which takes a generic `AtomList` like biotite.structure.AtomArray, or mda.AtomGroup, which defines the behaviour of the child classes and the rest is specified in the further classes. 


The idea is, that the `MoleculeInBlender` class is capable to be constructed from every Molecule class, so that new molecules could be easily added.

At this point, I've just restructured your code and chopped it up a bit to show what I want to show. For the more concrete implementation of how the `MoleculeInBlender` is created by the MDA molecule class, we need to have a closer exchange. At the moment there are two different init functions, but they have the potential to be unified.

I am looking forward to your opinion.

kainszs:
Good morning @BradyAJohnston,

I would like to continue working on improving the MoleculeInBlender class. After I have fully integrated the MDA session, I would like to add more molecules to the framework.

However, I have some questions:

1. Are you still interested in the idea? Or should I just close the PR and Issue?

If yes:

2. Does the MoleculeInBlender still need all attributes if they are already present in the molecule from which the MoleculeInBlender is created? For performance reasons, you could consider leaving them out and just calculating them as soon as they are needed.

3. What is the purpose of this line? 

https://github.com/BradyAJohnston/MolecularNodes/blob/ccb6a64b3c1d4fff5d5451e3789132d23d0f6f3d/molecularnodes/io/parse/mda.py#L575

BradyAJohnston:
I think the idea is solid - but lets not rush the implementation. 

Have you spent some time with the add-on, using Blender through the GUI? If not - I suggest that you spend some time playing around with Blender itself, and go through some of the existing documentation around Molecular Nodes to create some 3D renderings. 

While the GSoC project will mostly be focused around implementing better integration with MDAnalysis, it's important to understand what quirks and limitations we have to work around when dealing with Blender and the fact that we are an add-on for Blender first, and a python package second.

The line you are referencing makes the newly created object to be selected, once it is imported to Blender. This is an important step when using Blender through the GUI.

It might also be a good idea to tinker with creating a very simple add-on for Blender as well, as a way to familiarise yourself more with how Blender's API works. There are lots of YouTube tutorials and documentation around doing that. 

It'll take me some time (which I don't have a lot of currently) to go more over the proposed new implementation.

kainszs:
Sorry, I didn't mean to bother you either. I have already experimented with the add-on in Blender. But it seems to me a reasonable idea to write an Add-on myself to better understand the project.

I have nothing to do until next Wednesday, so I wanted to get started because it's a major change as announced. I agree with you that it shouldn't be rushed, as the code has its pitfalls.  Here is an example of the current functionality:

extended_functionality_mn

The `select_atoms` method is exactly the method from MDAnalysis. So you can easily create an object in blender using MDAnalysis, its AtomGroups or even apply MDA selections on existing molecules within blender. I hope you like the direction.

The implementation is by no means presentable yet, as I understand more each time I look at the code and work with it.

In addition, the class currently still supports a trajectory, but only plain molecules. However, this is only a minor extension, which is mainly related to the indexing of the `\_\_getitem\_\_ ` method.

PR #470, opened 2024-04-01
Documentation for starfiles

jojoelfe:
This should add the starfile node to the documentation. I am still working on the tutorial, but figured I open the PR to see if the CI for the docs works correctly.

ALso included are some small fixes for relion import that I noticed while making a tutorial.


BradyAJohnston:
The docs build OK on my machine - I need to look into why they aren't auto-triggering a build for the PR.

BradyAJohnston:
Now changed so that the docs will build & deploy on every PR :)

BradyAJohnston:
OK for security reasons we can't have a preview deployment unless the PR is from me - but the docs now at least still build & report if successful.

BradyAJohnston:
no rush on this - but I will be giving a tutorial to some CryoEM and CryoET people in a couple of weeks and would love to have a written tutorial for the starfile importing :)

jojoelfe:
Hey, I am slowly coming back to things after traveling.

I think the tutorial for cryo-EM SPA is in fairly good shape. It would be great to have something similar for cryo-ET, but there the problem is to find good datasets. An new choice would be some data from the CZI portal (https://cryoetdataportal.czscience.com/), but they don't use starfiles, but a json format. Maybe thats trivial to implement, but I haven't looked at it carefully yet.

BradyAJohnston:
Just put together an initial `.ndjson` parser in #493 if you wanted to take a look. If you can point out some good datasets to use as testing on that portal (I've had a look, but this isn't my field of expertise).

PR #493, opened 2024-05-09
Read `.ndjson` CryoET Files

BradyAJohnston:
For instancing of particles, reads `.ndjson` files and creates points for instancing.

Currently just the initial barebones importer which can parse the positions and the rotations for the points. Doesn't currently handle particle picking or anything similar.

Probably requires some refactoring of naming for the parsing classes. The importer panel should really be "CryoET" or something similar, with the possibility to open `.starfile` or `.ndjson` point instancing files.


jojoelfe:
OK, looks like the location coordinates are in pixels of the tomogram, but all densities get scaled to be in physical coordinates (as it should be). Unfortunately, the ndjson has no info about the pixel size, so for now we have to create a way for the user to manually enter the pixelsize or connect it with the volume, so the pixelsize can be read from there.

BradyAJohnston:
I think we can add a 'scale pixel size' to the node group which defaults to 1.0. Don't need to use it if the coordinates have been input but can use if needed for these datasets.

BradyAJohnston:
And yes I think we can rename the ensemble node group as well to be consistent.

jojoelfe:
Actually, I am pretty sure the rotations are wrong. I did some tests with this: https://cryoetdataportal.czscience.com/runs/506?metadata=run&tab=metadata and the Z-axes (blue arrow) should point out of the viral particles. (Inverting the rotation doesn't help either)

image

I just saw you are working on the matrix attribute, so maybe the thing to do is to write the matrix directly in there and go from there.

BradyAJohnston:
The matrix data format won't be until Blender 4.2 in early July. Currently I am taking the `xyz_rotation_matrix` and turning it into a 4x4 matrix. I believe it's currently filling the matrix row-wise first, but it might be defined column-wise? Do you know if there is a spec somewhere for the `.ndjson`? That way we can support more potential column names etc as well.

jojoelfe:
I haven't found a spec. I just asked in their repo how the matrix is defined: https://github.com/chanzuckerberg/cryoet-data-portal/discussions/749

BradyAJohnston:
Great - thanks @jojoelfe !

jojoelfe:
After some more tinkering:

```python
if has_rotation:
        matrix[:3, :3] = data['xyz_rotation_matrix']
        matrix[:3, :3] = np.flip(matrix[:3, :3])
```

makes things work:

https://github.com/BradyAJohnston/MolecularNodes/assets/6081039/b3feac4d-c175-46b3-956f-40055c01f5c3

BradyAJohnston:
Thanks for figuring that out @jojoelfe  - I've fixed the rotation and added a test for some basic import. Will be interested what they say about spec for the data formats.

BradyAJohnston:
I'm revisiting this. Since working on this a bunch of internal stuff has changed, including recently #703 which changed handling of cisTEM / RELION starfiles. 

I've just updated all of the code for this, and the importer now works properly, but I am still not 100% on the rotations. 

I tried importing the example data that you were showing, but the imported points also don't line up with the imported density data. The point positions are scaled by ~ 1.4 compare to the imported density data.

Any chance you could take a look and let me know which one is off?

BradyAJohnston:
Closing this PR for now but it's still something I would like to pursue in the future.

Issue #465, opened 2024-03-31
Defaul Materials


jojoelfe:
Hi Brady,

I would like to add a new Material called "MN AO Emission". 

![image](https://github.com/BradyAJohnston/MolecularNodes/assets/6081039/22e9f58f-b389-4a19-ab4d-9b0b1fb9fcaf)

I use it a lot in my work, mostly because it makes renders work without any lights setup. Settings up lights can be difficult when dealing with cellular scenes and is cumbersome to do for automated visualization scripts. Combined with the Cycles Spheres style it is also ridiculously performant in the live preview.

The main question I have is how the materials are organized at the moment. They are all in the data_file, but from what I can tell one cannot access them from default scenes. They are there when using the MN startup file, but I have not using it that much, because it seems more geared towards rendering single molecules. 

So concretely what I am wondering is:

- I would have to add the material both to the MN_data_file and startup file, correct?

- Maybe I should create a startup file more geared towards tomography/montaged Cryo-EM?

- Is there a mechanism to import the materials into blender files that are not derived from the MN template?

BradyAJohnston:
I like the material! I have in the past use just plain AO for the same reasons that it is _very_ performant with point clouds and viewport, but this is a nicer improvement on that material.

Currently I have the materials in both. The original idea was have them just in the data file, but I added them to the template so they would be immediately available in the template for users. I was hoping to have them also appendable in the materials workspace from the data file, just like the GN / MN node groups, but I haven't gotten around to making the panel yet.

I think a new template better setup for tomography / Cryo-EM would be good. I defer to you as to what would be useful in there. In the other template I have included an HDRI from poly-haven (which are CC0), so if you think an HDRI would be good you can use the same / different.

As I mentioned there currently isn't a mechanism to import, but we could add one without too much trouble (and should).

I believe this should probably be the realm of 'Asset Libraries'. I think ideally when we install the add-on, the data files become part of the asset library - then they become available automatically inside all parts of Blender. I did look into this briefly once, but found it to be a bit difficult to add a filepath to the asset library inside of Blender when installing. I will have another look into this perhaps.

BradyAJohnston:
Just revisited it in #467 and managed to get a button to install anything in the `assets` folder of the add-on as an asset library.

This opens the way to include all sorts of more niche node groups for more specific use cases like tomography etc. The nodes, models / materials / lighting can be in separate `.blend` files and all be accessible through the assets interface as well as being accessible through general scripting

[...]

PR #501, opened 2024-05-22
4.2

BradyAJohnston:
This PR changes Molecular Nodes to be compatible with the Blender 4.2 extensions system, and brings a number a number of other major upgrades. 

## Extensions Platform
Changed architecture of add-on to be compatible with new extensions platform in Blender 4.2 ([here](https://extensions.blender.org/add-ons/molecularnodes/)). Bundles the add-on with the required python packages, so installation is as simple as clicking 'install' from inside of Blender's extension system. No need to manually download or install anything additional. Updates for the add-on can also be delivered and installed this way.

### Color Tag
Utilising the new `color_tag` attribute for custom node groups, all provided node groups are now colored appropriately.
image

### Selection Nodes
Selection nodes have been split into individual node groups and given `And` and `Or` inputs to more easily construct selection chains with common boolean operations.
image

### Topology Nodes
Topology nodes have been partially overhauled, with some nodes moved to utility nodes and extras included such as the DSSP and Phi / Psi nodes. Phi / Psi nodes do not require pre-computation of any values and can be used just like any other field.
image

### Color Nodes
Res Name and Element nodes now go higher, utilising the `Index Switch` node underneath and panels to ensure the UI remains usable.

image

## Trajectories

The underlying method for importing and managing MDAnalysis trajectory / Universes has been overhauled. They are now imported as a `Trajectory` object. Trajectories along with all other molecular data that is imported is now kept around inside of a `MNSession`, which is stored under `bpy.context.scene.MNSession`. This session is pickled and save next to the `.blend` file on save, so that all of the data can persist between Blender sessions on load. 

### Selections

Because the connection to the universe is maintained, we now have live and updateable selections. 

### Calculations

Initial support for functions can be added to the Trajectories. These functions will be executed with the universe as an input, and the output is stored on the resulting mesh. This gives access to custom calculations to be executed in python and the results to be used inside of Geometry Nodes.

Implement changes for compatibility with Blender 4.2

Current build is approved and hosted on the extensions website - need to finish final changes before 4.2 release (https://extensions.blender.org/add-ons/molecularnodes/)

- [x] Change to distribution via the Blender Extensions platform (#500)
- [x] Add support for 4x4 matrix data format (#499)
- [x] Change node groups to using colors and add descriptions (#522)
- [x] Test cleanout of superfluous `.whl` files (like `numpy`, which is included with Blender)
- [x] Refactor MD selections to be 'live' rather than only on import (#466)
- [x] All relevant documentation is updated
- [x] Update default materials and include some more (#465)
- [x] Cleanup the add node menu (#537 )
- [ ] Update tests to run by building and installing the add-on 
- [x] Change panel to use [new panel code](https://developer.blender.org/docs/release_notes/4.2/python_api/#layout-panels)
- [x] Support adding selections from existing `AtomGroup`s (#538)
- [x] Rename `MNUniverse` to `Trajectory` (#541 )

Some feedback from extensions approval: 
- [x] Cleanup commented out lines of the `blender_manifest.toml`
- [x] .DS_store and some other files can be cleaned up (potentially download .whl files then use Blender to build package)
- [x] remove `requirements.txt` file as no longer needed for package management


molecularnodes/io/universe/universe.py
```diff
@@ -0,0 +1,517 @@
...
+                frame_a = int(frame / (subframes + 1))
+
+            # get the next frame
+            frame_b = frame_a + 1
+
+        if frame_a >= universe.trajectory.n_frames:
+            return None
+
+        # set the trajectory at frame_a
+        universe.trajectory[frame_a]
```
yuxuanzhuang, line 488: Reloading the trajectory timestep (`universe.trajectory[frame]`) can negatively impact performance because it reads the data from the file and overwrites the current frame, even if it's already at that frame. A solution is to check the current frame before reloading:

```python

@staticmethod
def _goto_frame(universe, frame):
    if universe.trajectory.frame != frame:
        universe.trajectory[frame]
```

yuxuanzhuang: More generally, I'm considering the possibility of implementing caching for these operations. I have tried the following approach:

```python
from functools import lru_cache

@lru_cache(maxsize=10)
def _goto_frame(self, frame):
    if self.universe.trajectory.frame != frame:
        self.universe.trajectory[frame]
    return self.positions
```

This method caches the 10 most recent outputs, avoiding the need to repeatedly reload the trajectory.

However, this approach results in different visualizations with interpolation. I'll need to investigate further to understand why. If you can figure it out, it would be great!

yuxuanzhuang: FYI, by performance, I mean the frame rate. If I switch the frame rate to 240 fps, the actual frame rate without subframe processing is 90 fps, and it drops to 60 fps with the old implementation.

yuxuanzhuang (commented):
The refactoring is really amazing! It's great to see how all of these components come together with a unified API. I have a few in-line questions and comments.

molecularnodes/io/universe/universe.py
```diff
@@ -0,0 +1,517 @@
...
+from ...utils import lerp, correct_periodic_positions
+from .selections import Selection, TrajectorySelectionItem
+
+
+class MNUniverse(MNDataObject):
+    def __init__(self, universe: mda.Universe, world_scale=0.01):
+        super().__init__()
+        self.universe: mda.Universe = universe
+        self.selections: Dict[str, Selection] = {}
+        self.world_scale = world_scale
```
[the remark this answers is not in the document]
BradyAJohnston: This could be set in the `MNDataObject` - thinking about it and the potential to change it after the fact - this should really be a `bpy.context.scene.MN_world_scale` property, so that when it changes it can be access from anywhere. This way it could be changed inside of node trees when the property is updated also. Unsure if we should actually want to support this though, as it has the potential to get messy. It is easy enough for users to just scale the actual object inside of the Blender scene instead.

I think in the meantime this can be moved to the `MNDataObject` though you are right. There are still lots of little bits of cleanup / optimisation like that to be done around the place.

molecularnodes/io/universe/selections.py
```diff
@@ -0,0 +1,161 @@
+import bpy
+import MDAnalysis as mda
+import numpy.typing as npt
+import numpy as np
+from bpy.props import StringProperty, BoolProperty
+from .handlers import _update_universes
+
+
+class Selection:
```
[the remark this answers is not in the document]
BradyAJohnston: I think what you propose is the way to go about it. Should be straight forward enough (if you wanted to implement it, I would very much appreciate the PR!)

molecularnodes/io/universe/universe.py
BradyAJohnston: Thanks for the tip - I didn't realise that it would reprocess the frame if it was already on it. I don't know how beneficial frame caching would be - users are usually playing back continuously I would imagine? Potential to implement, but maybe in a later PR rather than for the 4.2 deadlin (july 15) I am pushing for.

BradyAJohnston:
Thanks @yuxuanzhuang for the comments! I'm still interested in improving the API more, if you have suggestions about how it maybe should work. It is very satisfying though to have things be "clicking into place" now. 

One thing I am little unsure about is the name for the `MNUniverse` - as it also references a `mda.Universe`. I don't know if you have any suggestions for a better name to help avoid confusion.

molecularnodes/io/universe/universe.py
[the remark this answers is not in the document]
yuxuanzhuang: > It is easy enough for users to just scale the actual object inside of the Blender scene instead.

Yes, that’s what I thought—it seems to be a duplicate of the scale in Blender the way it’s currently set. I think setting a `bpy.context.scene.MN_world_scale` property, as you mentioned, makes more sense, and maybe set it as a property of MNSession in the meantime.

molecularnodes/io/universe/selections.py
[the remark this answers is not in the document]
yuxuanzhuang: I would love to write one!

molecularnodes/io/universe/universe.py
yuxuanzhuang: > Potential to implement, but maybe in a later PR rather than for the 4.2 deadlin (july 15) I am pushing for.

Yes, I completely agree.

I was mainly thinking about the potential cost of I/O operations when we want to stream trajectories from e.g. AWS in the future.

yuxuanzhuang:
> One thing I am little unsure about is the name for the MNUniverse - as it also references a mda.Universe. I don't know if you have any suggestions for a better name to help avoid confusion.

Any reason to not use `Trajectory` (not only as the class name but also as the folder name) since it juxtaposes with other classes (density, molecule, etc.)? I think Universe is only understandable to people who are familiar with MDAnalysis.

molecularnodes/io/universe/ui.py
```diff
@@ -0,0 +1,112 @@
...
+    universe = mda.Universe(top, traj)
+
+    mn_universe = MNUniverse(universe=universe)
+
+    mn_universe.create_model(name=name, style=style, subframes=subframes)
+
+    return mn_universe
+
+
+class MN_OT_Import_Protein_MD(bpy.types.Operator):
```
yuxuanzhuang, line 54: Just an idea for a more OOP design: I think there are a couple of similar code segments for importing in the UI. Maybe it's a better design to have a base class `MN_OT_Import_Base` for all of them, with a function `load` for each class. Also, I think `load` is doing dual jobs here of both loading and creating the model so maybe separating them into two functions as well.

BradyAJohnston: When I first created the add-on I wasn't holding on to the python objects after import, I was having a single function for each import method that would load and create the object. Now that everything has it's own class that hangs around, I think removing the `load` functions for everything makes more sense. 

Once you parse something into one of the classes, the user can call the `create_model()` or `create_object()` function (unsure about naming) to then create the actual object associated with the class / model / universe / molecule / etc. I want to keep the operator calls simple, so having the contents for each of the `load` functions just be the operator `execture()` functions I think makes sense and simplifies the code base a bit. 

I do however like the idea of this as the lines required to import and create a model into a scene.
```py
import molecularnodes as mn
mn.io.fetch('1cd3')
```

This is slightly more verbose, but I suppose more 'pythonic' and a better representation of the data flow in the scene.
```py
import molecularnodes as mn
mol = mn.io.fetch('1cd3')
mol.create_model()
```

BradyAJohnston:
> Any reason to not use `Trajectory` (not only as the class name but also as the folder name) since it juxtaposes with other classes (density, molecule, etc.)? I think Universe is only understandable to people who are familiar with MDAnalysis.

Just to clarify - you are in favor of using `Trajectory`? I was thinking about this as a potential name. I hadn't changed it yet because a universe also contains a trajectory, but I think it makes sense for what users are doing "I am importing an MD trajectory into Blender", "I am importing a molecule into Blender", "I am importing a protein ensemble into Blender"

molecularnodes/io/universe/universe.py
BradyAJohnston: For streaming this would definitely make sense!

yuxuanzhuang:
Yes, I think `Trajectory` makes more sense as a class name for general users/developers who want to contribute to the code. If you want to avoid confusion, it can also be `MNTrajectory`, and then the other class will be `MNDensity` and so on.

molecularnodes/ui/panel.py
```diff
@@ -162,34 +150,112 @@ def ui_from_node(layout, node):
...
+
+def panel_session(layout, context):
+    session = context.scene.MNSession
+    # if session.n_items > 0:
+    #     return None
     row = layout.row()
-    row.label(text="Experimental", icon_value=2)
-    row.operator("mn.add_armature")
+    row.label(text="Loaded items in the session")
+    row.operator("mn.session_reload")
```
yuxuanzhuang, line 243: You are probably aware of this, but this function doesn’t exist.

BradyAJohnston: Ah yes - this was from another PR. Will track it down and clean up.

BradyAJohnston:
Thanks for the feedback @yuxuanzhuang - I agree that `Trajectory` will be better - so I'll rename it in another PR soon.

PR #535, opened 2024-06-30
Per-Frame Calculations

BradyAJohnston:
Adds potential functions that are called each frame on the universe. The results of these are then stored on the object mesh for use in geometry nodes.

The appended functions are executed on the `bpy.app.handlers.frame_change_post` call. They use the universe and state of the trajectory that the universe is currently in (although this could be changed within the update function).

Potential to have a slightly different interface, but the easy setup is to just have a dictionary of functions that are called. The key in the dictionary becomes the attribute's name, and the type of the attribute is auto-detected (although this could potentially be improved).

Could potentially have some UI components, where you could change variables in the UI, that are then used inside of the function call, so that users could interact with the code as it is being executed on the universe.

```py
import bpy
import numpy as np

def distance_to_index(universe):
    pos = universe.atoms.positions
    return np.linalg.norm(pos - pos[1], axis=1)

session = bpy.context.scene.MNSession
bob = bpy.data.objects['NewTrajectory']
session.universes.get(bob.mn.uuid).calculations['testing'] = distance_to_index
```

https://github.com/BradyAJohnston/MolecularNodes/assets/36021261/6697198b-17b1-4c8f-8265-7c8f8c69fd90

Currently any calculations are removed on saving (seems that there are problems with pickling them). Certainly could be improved potentially.

TODO: 
- [ ] Handle changing of output property type (remove old attribute, create new one with new type)
- [ ] maybe cleanup old attributes if a function no longer exists (might be best to just leave to the user to be responsible for)
- [ ] Potential for 'updating' toggle / boolean for if to evaluate
- [ ] Maybe a bespoke class rather than just a dictionary


BradyAJohnston:
Ping @yuxuanzhuang and @rbdavid if either of you were interested in having a look / giving some input

PR #543, opened 2024-07-10
Handle existing mda session being None

prihoda:
When parsing a MD file for the first time with molecular nodes 4.1.0 as well as 4.1.0, I got an error that existing_session is None here: https://github.com/BradyAJohnston/MolecularNodes/blob/33a07854f471eebe8f131d260b3fcbc9416f69b0/molecularnodes/io/parse/mda.py#L438

Not sure where this is coming from but this fixes the issue

### DIFF
--- a/molecularnodes/io/parse/mda.py
+++ b/molecularnodes/io/parse/mda.py
@@ -431,7 +431,7 @@ class MDAnalysisSession:
             raise ImportError("MDAnalysis is not installed.")
 
         # if the session already exists, load the existing session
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
Hello! Thanks for catching this and for the PR. 

All of this is being completely overhauled in the upcoming 4.2 release #501 which will also fix this issue. I'm happy to accept this and put out a small bugfix for 4.1 though!

Was this only happening after installing MDAnalysis / python packages for the first time? Was it fixed after restarting Blender maybe?

PR #616, opened 2024-09-22
Add trajectory reloading

### LINKED ISSUE #614: MNSession File System Is Bugged
**Is your feature request related to a problem? Please describe.**
Everytime blender crashes or closes I lose my trajectory and have to re-import. 

**Describe the solution you'd like**
I would like to be able to cache to my disk so that I can use the same object if I need to close the app or if it craches

### LINKED ISSUE #615: Backup for MD filepaths
Sometimes the loading of the `.MNSession` file can break. To restore the link to the underlying universe, if we store the file paths to the topology / trajectory on the object itself as well, then it can be restored from the Object without the need for any other data. Having the option to change / restore from new filepaths as well is likely a good option.

BradyAJohnston:
Fixes #615 
Fixes #614 

Helps to fix errors where the trajectory loses it's connection to the underlying universe. Adds operator to more easily restore the trajectory when the connection is lost

https://github.com/user-attachments/assets/e3f176af-d11c-466d-8215-cd53a1883023

- **Operator to reload trajectory**
- **fixed problem with pickling while saving**

### DIFF
--- a/molecularnodes/__init__.py
+++ b/molecularnodes/__init__.py
@@ -45,8 +45,8 @@ def register():
     for op in all_classes:
         try:
             bpy.utils.register_class(op)
-        except Exception:
-            # print(e)
+        except Exception as e:
+            print(e)
             pass
 
     bpy.types.NODE_MT_add.append(MN_add_node_menu)
--- a/molecularnodes/blender/__init__.py
+++ b/molecularnodes/blender/__init__.py
@@ -10,3 +10,10 @@ def path_resolve(path: Union[str, Path]) -> Path:
         return Path(bpy.path.abspath(str(path)))
     else:
         raise ValueError(f"Unable to resolve path: {path}")
+
+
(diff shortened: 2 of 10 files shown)


CeraveWasTakenLOL:
Just curious, does this reset the entire geometry node system? If so it might be advantageous to add a manual caching system for the MNSession

BradyAJohnston:
@CeraveWasTakenLOL this doesn't change anything about the Object / Geometry Nodes setup (so any keyaframes etc will be maintained) it just re-links the underlying `MDAnalysis.Universe` so that the positions update on frame change

BradyAJohnston:
@CeraveWasTakenLOL also there is a `Bake` node inside of Geometry Nodes so if you are wanting to bake & create a permanent save state of the trajectory you can use that node to bake the positions to disk (in Blender 4.3 it will be able to bake into the `.blend` file)

Issue #662, opened 2024-11-20
Error Creating Relative Path When Absolute Path Is Used as Input


yuxuanzhuang:
**Describe the bug**
When providing an absolute path as input to `make_path_relative`, the output is incorrect.

**Error Codes**

```python
import os
import molecularnodes

# Error path
molecularnodes.session.make_path_relative('/Users/yzhuang/test.pdb')
> ./Users/yzhuang/test.pdb

os.path.relpath('/home/yzhuang/test.pdb')
> ../../../home/yzhuang/test.pdb

# This only works with the current working directory
molecularnodes.session.make_path_relative('/current_working_directory/test.pdb')
> ./test.pdb
```

**Desktop (please complete the following information):**
 - OS: [e.g. Windows / MacOS / Linux]： MacOS
 - Hardware: [e.g. Intel Mac, M1 Mac, x86 PC]: M1 Mac
 - Blender Version: 4.3
- MolecularNodes Version: develop

BradyAJohnston:
Thanks for the report! This was something that was kind of hacky and that I had meant to get back to and clean up.

BradyAJohnston:
This seems fixed in the latest 4.5

PR #663, opened 2024-11-21
Fix relative filepaths

BradyAJohnston:
If we import a molecular dynamics trajectory before first saving the `.blend` file, then the file paths will be absolute for the topology and trajectory. This means the `.blend` is no longer portable. 

This is fine if we remain on the same machine, but breaks if we try to move a folder onto another machine for rendering. 

To fix this, part of the saving / serialisation process has to be turning absolute file paths into relative. This was previously quite janky (#662 ), so needs to be improved and streamlined.


BradyAJohnston:
will revisit in another PR / issue.

Issue #658, opened 2024-11-14
Problem when loading pdbs with no bonds 


brenerrr:
# The problem
I encoutered some problems when trying to load PDB files that do not have bonds information and more than one chain identifier, such as 

```
REMARK original generated coordinate pdb file
ATOM      0   C1   BGX   1      -3.563 -12.592-256.215  1.00  0.00           C
ATOM      1   H1   BGX   1      -2.591 -13.101-256.040  1.00  0.00           H
ATOM      2   C5   BGX   1      -2.763 -11.186-258.118  1.00  0.00           C
ATOM      3   H5   BGX   1      -1.754 -11.645-258.040  1.00  0.00           H
ATOM      4   O5   BGX   1      -3.299 -11.258-256.729  1.00  0.00           O
ATOM      5   C2   BGX   2      -4.485 -13.321-257.148  1.00  0.00           C
ATOM      6   H2   BGX   2      -5.452 -12.773-257.146  1.00  0.00           H
END
```

When trying to load it, There will be an error from biotite on line 29 of `molecularnodes/entities/molecule/pdb.py`

```
array = pdb.get_structure(
            pdb_file=self.file,
            extra_fields=["b_factor", "occupancy", "charge", "atom_id"],
            include_bonds=True,
        )
```
because of the "include_bonds=True" flag. 

# My solution 
I forked the main repository and made the following modifications 

```
array = pdb.get_structure(
    pdb_file=self.file,
    extra_fields=["b_factor", "occupancy", "charge", "atom_id"],
    include_bonds=False,
)
try:
    array.bonds = connect_via_residue_names(array)
except AttributeError:
    print("Not able to find bonds via residue.")
```
Now the array is loaded without bonds, which are built afterwards. In case biotite raises an AtributeError, then the atoms are loaded without any bond information. 

If you think this is a good solution I can submit a pull request later.  I am also open to suggestions about this solution.

BradyAJohnston:
Thanks for catching this! A PR would be great, but I think we should instead also do `try` / `except` for the `pdb_get_structure`. 

```py
try:
    array = pdb.get_structure(
        pdb_file=self.file,
        extra_fields=["b_factor", "occupancy", "charge", "atom_id"],
        include_bonds=True,
    )
except (ErrorThatIsRaised):
    array = pdb.get_structure(
        pdb_file=self.file,
        extra_fields=["b_factor", "occupancy", "charge", "atom_id"],
        include_bonds=False,
    )
[...]

PR #666, opened 2024-11-23
Pdb Loading Fix

brenerrr:
Previously, an error would be raised when attempting to find bonds in a PDB file that lacked bond information and contained multiple chain identifiers. Now, the atoms are first loaded and then connected. This is mainly an issue about how biotite handles getting the structure from the pdb.

More information can be found on #658.


BradyAJohnston:
Thanks for this! Are you able to add a test case with some example data that would fail on the old method but now succeeds with this fix?

brenerrr:
Sure. I just added a new test that loads a pdb called "no_bonds.pdb" very similar to the one I posted on #658 and checks the position and bond types (which are none). When executing it without the fix the test fails.

BradyAJohnston:
I tweaked it to add back in the attempt to import with bond information. People like to abuse the heck out of the PDB format, so they could have all sorts of weird and wonderful bond information. I'd rather try and get the bonds first, then if that fails fall back on finding via residue names.

PR #671, opened 2024-11-25
Add option for trajectory position averaging

### LINKED ISSUE #440: Add trajectory smoothing when importing via MDAnalysis

BradyAJohnston:
Adds option for trajectory averaging, fixes #440. 

Increasing the `Average` number, accesses that number of frames before and after the current frame, then gets the average position for all of the frames and uses that to update the object's vertex positions.

https://github.com/user-attachments/assets/9fdb6760-5fb1-4ba4-a966-baff22bdb673

### DIFF
--- a/molecularnodes/bpyd/object.py
+++ b/molecularnodes/bpyd/object.py
@@ -428,14 +428,14 @@ class BlenderObject:
             A list of attribute names if the molecule object exists, None otherwise.
         """
         if evaluate:
-            strings = list(self.evaluate().attributes.keys())
+            strings = list(self.evaluate().object.data.attributes.keys())
         else:
-            strings = list(self.object.attributes.keys())
+            strings = list(self.object.data.attributes.keys())
 
         if not drop_hidden:
             return strings
         else:
-            return filter(lambda x: not x.startswith("."), strings)
+            return [x for x in strings if not x.startswith(".")]
 
     def __len__(self) -> int:
         """


BradyAJohnston:
@yuxuanzhuang if you wanted to have a look and give any feedback before I merge tomorrow

molecularnodes/utils.py
```diff
@@ -40,6 +45,37 @@ def correct_periodic_positions(
...
+
+    frame_a = frame
+
+    if subframes > 0:
+        frame_a = int(frame / (subframes + 1))
+
+    return frame_a
+
+
+def frames_to_average(frame: int, average: int = 0) -> list[int]:
```
yuxuanzhuang, line 74: When `frame=0`, this will also generate negative values; maybe better just filtering those values out?

yuxuanzhuang (commented):
It looks great overall! One potential concern could be the performance impact of reading multiple frames (both in average and subframe). It might be worth considering adding a caching mechanism in the next major release to address this.

molecularnodes/utils.py
BradyAJohnston: good idea, added it in

BradyAJohnston:
I've now added in a caching mechanism. By default it will cache up to 2 frames, so when interpolating without averaging the switch the next frame will only have to fetch the next-next frame. 

When averaging, there will be a window of frames that are collected and cached, so moving forward by 1 frame with a window of 3 frames, will drop the no longer needed frame and fetch only 1 extra frame of positions before averaging. 

The current setup just uses a dictionary under `self.cache` which has the frame number (corresponding to the universe's trajectory frame number) and the numpy array with atom positions for it.

PR #687, opened 2024-12-14
Change BlenderObject lookup

BradyAJohnston:
`BlenderObject` and thus `MolecularEntity` references to blender's objects are tweaker to just use a name (and potentialy `.uuid` backup reference) instead of reference to the actual `bpy.data.objects['oject']` data block.

## The Problem
Previously the `BlenderObject`/`MolecularEntity` had the property `bob.object` which returned the object in Blender's database that was linked to this entity. This was previously stored as the actual reference to the object itself. This was bad because sometimes that object would change in Blender's database and we would lose connection. This could be re-established with the `object.uuid` custom property defined. If the connection was lost we would just loop through all objects in the database for something with that uuid and re-establish the connection. 

This didn't work well with pickling though, as `bpy.data.Object` couldn't be pickled.

## The Solution
References to objects, collections, and other components of Blender's database are instead stored as a string of their name. This way every time we want to access the `object` it queries the database for the object. This fails if objects are renamed, because there is no deeper reference to the objects inside of Blender that we have access to, so with the `bpy.types.Object.uuid` property, we can do the same as previously and just iterate through the database looking for our match.


BradyAJohnston:
The tests are all passing on GHA but the final tests related to trajectories fail on my local machines.

When running the individual tests with `-k traj` they pass fine, it's only when running all tests and they are last they fail. But running all tests and they are last on the GHA runners they pass. I am going insane.

BradyAJohnston:
The tests pass on local machine if running tests in a `uv` environment instead of inside of Blender, but I can't run the tests involving density that way as the `bpy` python module doesn't ship with `pyopenvdb`.

```bash
uv venv python=3.11
source .venv/scripts/activate
uv pip install -r pyproject.toml --all-extras
uv pip install -e .
uv run -m pytest
```

PR #695, opened 2024-12-19
swap out internal bpyd for databpy

BradyAJohnston:
Now uses the [`databpy`](https://pypi.org/p/databpy) package over the internal `bpyd`. 

Internal package has been split out into a separate python module for use by other projects.


BradyAJohnston:
Failed test is using using databpy==0.0.2 which is why it is failing. Will merge anyway.

PR #565, opened 2024-08-07
Mesoscale cif and bcif clean-up

corredD:
Use biotite cif and bcif parser
Add support for some color file, as a json dictionary
Need a patch on biotite to support YASARA petworld file ( attached )
[cif.txt](https://github.com/user-attachments/files/16535616/cif.txt)
Replace Blender\4.2\extensions\.local\lib\python3.11\site-packages\biotite\structure\io\pdbx\cif.py
Default node show a clipped fraction of the model as point in the viewer

image

PR #708, opened 2025-01-09
Cellpack bcif

BradyAJohnston:
This address most of what was being addressed in #565 but with the updated state of the edd-on. 

- [x] YASARA files: 
  - A custom reader the overrides the `biotite.structure.io.CIFFile` reader has been added which addresses the prefixed spaces in YASRA files that were being problematic.

- [x] Remove legacy cellpack reading code
  - This moves everything to use biotite's `biotite.structure.io.pdbx` readers for both CIF and BCIF files, greatly simplifying the overall code base and requiring a lot less cellpack-specific code. 
- [ ] Colors from palette.json
  - Currently don't have example .json palette files to work from

@corredD rather than trying to merge with the PR, I've developed this separately which does most of what you implemented. What I don't have working is the colors being imported and would love your input on. 

Everything should be much cleaner now, with the major thing to change is that for both BCIF and CIF files, they read in a `self.molecules` which is a dictionary of individual `AtomArray`s, for each chain that will be instanced. The color code that was there previously and I have mostly moved over but not implemented was assuming one large AtomArray.


BradyAJohnston:
I'm going to merge this, as it addresses the important stuff of being able to import the YASARA files. Coloring based on .json files is more general and can be introduced in a later PR.

PR #745, opened 2025-02-20
ignore minor error when initialising selection

BradyAJohnston:
I can't currenlty figure out a nice way to have selections update when their UIList properties are updated, but not throw errors when starting, so for now we can just ignore them if they aren't present.

### DIFF
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -49,14 +49,13 @@ class Trajectory(MolecularEntity):
         periodic: bool = True,
     ):
         "Adds a new selection with the given name, selection string and selection parameters."
-        selection = Selection(
-            trajectory=self,
+        selection = Selection(trajectory=self, name=name)
+        self.selections[selection.name] = selection
+        selection.add_selection_property(
             selection_str=selection_str,
-            name=name,
             updating=updating,
             periodic=periodic,
         )
-        self.selections[selection.name] = selection
 
         return selection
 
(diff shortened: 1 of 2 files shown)


BradyAJohnston:
@PardhavMaradani I can't seem to assign you as a reviewer but if you wanted to check it out that would be great

PardhavMaradani:
The changes looks fine to me.

This is how I had to break it apart to get past this error. Just fyi - in case you think it is any better.

```
diff --git a/molecularnodes/entities/trajectory/base.py b/molecularnodes/entities/trajectory/base.py
index 652fd52..165fa87 100644
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -49,14 +49,13 @@ class Trajectory(MolecularEntity):
         periodic: bool = True,
     ):
         "Adds a new selection with the given name, selection string and selection parameters."
-        selection = Selection(
-            trajectory=self,
+        selection = Selection(trajectory=self, name=name)
+        self.selections[selection.name] = selection
+        selection.add_selection_property(
             selection_str=selection_str,
-            name=name,
             updating=updating,
             periodic=periodic,
         )
-        self.selections[selection.name] = selection
 
         return selection
 
diff --git a/molecularnodes/entities/trajectory/selections.py b/molecularnodes/entities/trajectory/selections.py
index dd56820..0f9ddc8 100644
--- a/molecularnodes/entities/trajectory/selections.py
+++ b/molecularnodes/entities/trajectory/selections.py
@@ -4,14 +4,14 @@ import numpy as np
 from uuid import uuid1
 
 class Selection:
-    def __init__(
-        self, trajectory, name: str = "selection_0", selection_str: str = "all",  updating=True, periodic=True
-    ):  
+    def __init__(self, trajectory, name: str = "selection_0"):
         self._ag: mda.AtomGroup | None = None
         self._uuid: str = str(uuid1())
         self._name = name
         self._current_selection_str: str = ""
         self.trajectory = trajectory
+
+    def add_selection_property(self, selection_str: str = "all", updating=True, periodic=True):
         self.trajectory.object.mn_trajectory_selections.add()
         self.trajectory.object.mn_trajectory_selections[-1].name = self.name
         self.trajectory.object.mn_trajectory_selections[-1].uuid = self._uuid
```

BradyAJohnston:
I've been trying to keep the creation of the UIList property tightly coupled to the creation of the `Selection`, but maybe going with your method and decoupling then is a better strategy. Might reduce headaches overall.

PardhavMaradani:
As the `add()` method of a `CollectionProperty` returns the property added, it could be simplified to something like:
```
def add_selection_property(self, selection_str: str = "all", updating=True, periodic=True):
    prop = self.trajectory.object.mn_trajectory_selections.add()
    prop.name = self.name
    prop.uuid = self._uuid
    .....
```
as well.

Issue #750, opened 2025-02-24
Direct Biotite to MN method


zachcp:
**Is your feature request related to a problem? Please describe.**

I  would like to be able to load/manipulate Biotite AtomArrays and then convert them directly to MN Molecules.

**Describe the solution you'd like**

Extend the  [Molecule.__init__()](https://github.com/BradyAJohnston/MolecularNodes/blob/976eb660d3c1e4128cfaa2c410f62bb71a64da81/molecularnodes/entities/molecule/base.py#L52) function to allow AtomArray or AtomStack.

**Describe alternatives you've considered**

```python
# using the internal FN.....
# https://github.com/zachcp/prettymol/blob/main/src/prettymol/core.py#L31C1-L31C47
obj, _ = _create_object(arr, name=molname)
```

**Additional context**

Related to the Biotite `interface` effort to define interfaces with other programs.  https://github.com/biotite-dev/biotite/issues/709

zachcp:
( cc: @BradyAJohnston  )

BradyAJohnston:
This seems like a very reasonable addition for a class method to ingest biotite.AtomArray/Stack directly. 

It makes my think that maybe it shouldn't all go through the init method, but maybe should instead be individual class methods for creating a Molecule via different methods. On that aspect I defer to those with more python experience to say how it should work to be more "pythonic".

BradyAJohnston:
Additionally for the overall MN design, currently I don't use `Molecule` directly but instead subclass it into `PDB / CIF / BCIF` etc for the different import methods and methods of getting assembly information etc. It would ultimately be better to have them all just ultimately be a "Molecule" I think? But I don't actually know how best to try and implement that

zachcp:
So I definitely like the idea of simplifying the conversion from AtomArray to Blender. However, after looking a bit deeper under the hood I am wondering if it's worth adding a  middle layer  that is basically a table/dataframe of all of the data prior to conversion to blender.  I am imagining 3 types and 2 conversion:s

## Types/Classes
1. 
2. Biotite AtomArray/AtomStack
3. BlenderFrame or MNFrame (a class / implementation that sits on top of a data frame library and has all of the additional properties that MN uses)
4. the Blender Object with Meshes/data/modifiers etc.

## Conversions

1 -> 2 would encapsulate the core conversion logic of "is_carbon" etc.
2->3 would  be as simple as possible - e.g. datable BOB+StyleNode+Material.

## Benefits

[...]

PR #751, opened 2025-02-24
AtomArray Direct Loading

zachcp:
WIP RE: issue  #750 

Pencilling out a solution before getting too deep.

### DIFF
--- a/molecularnodes/entities/molecule/array.py
+++ b/molecularnodes/entities/molecule/array.py
@@ -1,0 +1,56 @@
+import biotite.structure as struc
+import numpy as np
+from biotite import InvalidFileError
+from biotite.structure import (
+    BadStructureError,
+    annotate_sse,
+    spread_residue_wise,
+    connect_via_residue_names,
+    AtomArray
+)
+from biotite.structure.io import pdb
+
+from .assembly import AssemblyParser
+from .base import Molecule
+from .pdb import _comp_secondary_structure
+
+
+class Array(Molecule):
+    def __init__(self, array: AtomArray):
+        self.file = "ARRAY_LOADED_DIRECTLY"
+        self.array = self._validate_structure(array)
+        self.n_atoms = self.array.array_length()
+
+    def read(self, file_path):
+        pass
+
+    def _validate_structure(self, array: AtomArray):
+        print(array.get_annotation_categories())
+        # TODO: implement entity ID, sec_struct for PDB files
+        extra_fields = ["b_factor", "occupancy", "charge", "atom_id"]
+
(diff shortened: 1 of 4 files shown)


BradyAJohnston:
LGTM, some discussion around overall implementation in the issue

molecularnodes/entities/molecule/array.py
```diff
@@ -0,0 +1,52 @@
...
+        self._uuid = str(uuid.uuid4())
+        self.array = self._validate_structure(array)
+        self.n_atoms = self.array.array_length()    # note: I am not sure if this is consistent
+        self.create_object(name=name)               # Create the Blender object
+
+    def read(self, file_path):
+        pass
+
+    def _validate_structure(self, array: AtomArray):
+        # TODO: implement entity ID, sec_struct for PDB files
```
zachcp, line 25: A complete implementation should handle these fields as well by setting defaults.  The assumption here is that the Array is being passed and the file is no longer available. I letft it unimplemented pending a good solution

molecularnodes/entities/molecule/base.py
```diff
@@ -65,6 +64,12 @@ def __init__(self, file_path: Union[str, Path, io.BytesIO]):
         self._frames_collection: str | None
         self._entity_type = EntityType.MOLECULE
 
+
+    @classmethod
```
zachcp, line 68: Preferred entry point is now a class method instead of over loading the __init__/ Much cleaner

molecularnodes/entities/molecule/array.py
```diff
@@ -0,0 +1,52 @@
+import uuid
+from biotite.structure import AtomArray
+
+from .base import Molecule
+from .pdb import _comp_secondary_structure
+from ..base import EntityType
+
+
+class Array(Molecule):
```
zachcp, line 9: All the. logic is in a new Array Class analogous to the others. This has to do a lot less work though!

zachcp:
Much simplified PR that is narrower in scope and follows the suggestions from the issue thread. Namely:

1. Subclass Molecule as you do for other IO
2.  Do not use __init__

BradyAJohnston:
After our discussion this morning I think we came to the conclusion that a new distinct `AtomArray` isn't the right way to go, but instead just rolling more things into the `Molecule` and instead computing and storing all of the array relevant stuff on the `Molecule.array`, with a potential init or class method to read a `biotite.AtomArray` directly.

zachcp:
Hey Brady, I'll write up my notes but we will need a way to handle and validate the inputs including for Array. So I would merge this and I can reorganize all of the inputs as a follow up

zachcp:
Closing.

Issue #734, opened 2025-02-12
A discussion on scripting with Molecular Nodes and Blender more generally


BradyAJohnston:
This is an attempt to collect thoughts and projects related to scientific scripting and rendering with Blender. It would better fit in a more general-purpose place for scientific interaction & visualisation with Blender (like a GitHub organisation for such), but as it currently doesn't exist I figured here would be a good a place as any to collect links, thoughts and comments from projects and people working in the space. 

The text below on this initial post will be updated with additional links to projects that are relevant as they are found.

## Molecular Nodes related projects
- [Molecular Nodes](https://github.com/BradyAJohnston/MolecularNodes/)
- [ggmolvis](https://github.com/yuxuanzhuang/ggmolvis): potential for more scripting and API usage with Molecular Nodes
- [prettymol](https://github.com/zachcp/prettymol): potential for more scripting and API usage with Molecular Nodes
-  [basimol](https://github.com/zachcp/basimol): experiment with scripting with Molecular Nodes via baslisp language
- [experiments with GPU module](https://github.com/MDAnalysis/mdanalysis/discussions/4862): using the `gpu` module inside of Blender with MDAnalysis to visualise data & results in the viewport, enabling better interaction & feedback
- [Visualising Viruses](https://www.youtube.com/watch?v=adhTmwYwOiA): Molecular Nodes talk at BCON22
- [MN & MDAnalysis workshop](https://www.youtube.com/watch?v=3zKBjnRnAMg): recording of workshop showing intro to MDA and MN

## Molecular data related: 
- [Microscopy Nodes](https://github.com/oanegros/MicroscopyNodes): add-on focused on microscopy data import
- [batoms](https://github.com/beautiful-atoms/beautiful-atoms): more focused on small-molecule and material science rendering of atomic data in Blender
- [Atomic Blender](https://extensions.blender.org/add-ons/atomic-blender-pdb-xyz/): original import of molecular data including .pdb files into Blender (used to ship with Blender, now an extension)
- [pyrite](https://github.com/durrantlab/pyrite): older add-on for import MD data into Blender
- [upy & AutoPack](https://www.autopack.org/install/blender-installation): older add-on for importing molecular data, coupled with CellPack usage
- [CellBlender](https://github.com/mcellteam/cellblender): simulation of molecular data in Blender
- [Cella](https://onlinelibrary.wiley.com/doi/full/10.1111/ppl.14068): Plant single cell transcriptomics in Blender

## General science-related:
[...]

PR #760, opened 2025-02-27
Molecule Selection API

zachcp:
This implements the chained selection creation discussed in  #734  and originally from [prettymol](https://zachcp.github.io/prettymol/)

- add `MoleculeSelector`
- adds mol.select` property on the `Molecule` class

## Examples

This this approach, we can chain selections and then store the results as named attributes on the molecule's mesh.

```python
import molecularnodes as mn
mol = mn.Molecule.fetch("8H1B")
mol.select.atom_name("CA").chain_id(["A"]).store_selection("custom_selection")
```

This allows for a cleaner API, where we build the selection iteratively with chained calls, and only for the final `store_selection()` is it actually evaluated and stored on the mesh.
```py
mol.select.reset().atom_name("CA").store_selection("custom_selection")
(
    mol.select.reset()
    .is_peptide_backbone()
    .res_id(range(50))
    .chain_id(["B"])
    .res_name(["LYS", "ARG", "HIS"])
    .store_selection("custom_selection")
)
```

```py
sel = mn.entities.MoleculeSelector

# the selections here are staged, but not evaluated as there is no array or molecule yet associated with the selection.
(
    sel.res_id(range(200))
    .chain_id(["A", "B"])
    .res_name(["LYS", "ARG", "HIS", "GLY", "PRO"])
    .is_backbone()
)

sel.evaluate_on_array(mol.array) # returns a numpy array that can be used to subset the `AtomArray`.
```


BradyAJohnston:
Like I've said before, I like the approach of chaining masks to more cleanly create selections, but I think there are a couple of changes we should make.

I've just written up some thoughts in #762 about how currently we filter out some points when creating the object, and instead we shouldn't do any filtering and instead import all points and just use selections as named attributes for filtering when creating the final geometry. The "Mesh" should instead 100% represent the points that are in the dataset we have imported.

For the masks / selections, first I think we need to align on naming with Trajectories (which use "selections"). Maybe we have `mol.mask` that can be used to create a mask, but only once it is applied to the `AtomArray` or the `Object` does it become a "selection"? Additionally I think the selection methods should all be behind a top level, such as behind `mol.mask` like this:
```py
from molecularnodes import parse

mol = parse("tests/data/8FAT.bcif")

# calling the method on the mask updates the underlying array
mol.mask.chain("A").resnum(range(100).resname(["LYS", "ARG"])

# calling .array() returns the numpy array
print(mol.mask.array())

# calling set_selection would then write the values to the `AtomArray` as a column, and if it exists the object as
# a named attribute as well
mol.mask.set_selection("selection_name")

# this will then use the named attribute "selection_name" for the style, which we could also update later 
# without recreating and objects
mol.create_object(style="ribbon", name="Ribbon", selection="selection_name")
```

Additionally, there was discussion [here](https://github.com/yuxuanzhuang/ggmolvis/pull/22) about how best to approach multiple representations of the same underlying molecule / trajectory (for our own usage but also in an effort to align and support MolViewSpec). The final idea we settled on was to only have a single Object associated with a single dataset (molecule or trajectory etc). Then inside of the GN node tree, we create different branches for each representation instead. This keeps everything cleaner in terms of having a single object inside of Blender associated with a single dataset.

zachcp:
- I am fine with whatever naming scheme you prefer. Can you comment in the code in  PR?
- In my code, the mask is only used once we hit `create_object`. e.g its just an array until the Blender Object is created.
- Regarding putting everything behind another name e.g. `mol.max.chain....` instead of `mol.chain`. This actually introduces some implemention complexity. What I think might be cleaner sis to change the names to denote selection - e.g. `mol.select_chain('A')`. the  selection accounting is tracked by and attribute (I am using `mask`).

BradyAJohnston:
I think I still prefer having everything behind a top level `mol.select.chain().resname()`.

Reasoning being that it reduces number of methods on top level `Molecule` and I think would reduce confusion.

```python
# to have this work, we would need to have the whole molecule be returned, so in theory these two 
# methods of calling chained or on different lines are the same, but not clear that in the second
# instance we aren't just creating a whole new selection
mol.select_chain("A").select_resname("LYS")

mol.select_chain("A")
mol.select_resname("LYS")

# this is cleaner that every talk we start using `mol.select.` we would start a new selection
mol.select.chain("A").resname("LYS")
```

BradyAJohnston:
I was just thinking about this on my cycle home today, and that I could make some changes to it and merge as I think I've sorted out how adding of styles should work.

zachcp:
If you can modify it and incorporate it that would be great. There was a lot moving API changes and I didn't want to do anything further until they've settled. But if you have a good sense of how it all ties together, fantastic.

BradyAJohnston:
My current thoughts are that it can become part of the molecule class behind `mol.selector` or `mol.selection`, but can also be used to be passed into another function as an "anonymous" selection function that is applied later on.

```py
mol.selector.chain_id(["A", "B"]).res_name(["LYS", "ARG"]).store("selection_1")
mol.add_style("cartoon", selection = "selection_1")

sel = Selector().chain_id(["A", "B"]).res_name(["LYS", "ARG"])
mol.add_style(
    "cartoon",
    selection = sel
)
```
This would mean that the `Selector/Selection` class would support "lazy" evaluation of the selection, stacking the requested selections to be evaluated, even if it doesn't currently have access to an array to apply the selection to.

BradyAJohnston:
Okay I think this is in the shape where I wanted it to be. The selection is now accessible from the `Molecule` class, but we can also just construct selections indpendent of a molecule / actual structure using the `Selector` class and it's methods. Then just using the `Selector().evaluate_on_array(array)` method applies all of those selections and returns the boolean array.

Am interested to see what you think about this approach.

molecularnodes/entities/molecule/base.py
```diff
@@ -321,3 +321,50 @@ def __repr__(self) -> str:
             The string representation of the Molecule object.
         """
         return f"<Molecule object: {self.name}>"
+
+
+class MoleculeSelector(Selector):
```
zachcp, line 326: In general, I would recommend fewer classes. E.g. just use the Selector class and remove one layer of indirection.  If there were additional inheritors of Selector I think it might make sense. So I would personally vote for adding the store_names_attribute to the selector class and ditching this Class. Alternatively: rename Selector  to `MoleculeSelector`.

But in the end its style not substance. So non blocking.

tests/test_selector.py
```diff
@@ -0,0 +1,28 @@
...
+def test_select_ca():
+    mol = mn.Molecule.fetch("8H1B")
+    mol.selector.atom_name("CA").chain_id(["A"]).store_named_attribute(
+        "custom_selection"
+    )
+    assert sum(mol.named_attribute("custom_selection")) == 190
+    mol.selector.reset().atom_name("CA").store_named_attribute("custom_selection")
+    assert sum(mol.named_attribute("custom_selection")) == 384
+    (
+        mol.selector.reset()
```
zachcp, line 13: I didn't see how you were thinking but this is quite clean as an API. 

This shows the selections work.  I am still not sure how you are imagining tying this together for VIZ. But that is out of scope.

zachcp:
@BradyAJohnston  - this has a nice ergonomics. I highlighted two issues above,  both non-blocking. Great work; thanks.

molecularnodes/entities/molecule/base.py
BradyAJohnston: I split it up so one could be completely 'anonymous' and work separate to any molecule, and then the `MoleculeSelector` is the one that is directly associated with the class. I could combine them though like you said.

tests/test_selector.py
BradyAJohnston: Thanks. It'll hopefully become more clear when I can merge others as well and it starts coming together.

zachcp:
My 2 cents: 

- most users of this API will be programmers of some sort.
- `selector` is more verb-like and `selection` is more noun-like
-  I would stick with `selector` and use `selection` for the final "thing"
-  in that vein you might consider `mol.select` instead of `mol.selector` and make the final call in the chain `create_selection` or something to that effect. e.g. its not anything "tangible" to the end used until the final call.

PR #766, opened 2025-02-28
Collapse classes for Molecule

BradyAJohnston:
Collapses classes for different data formats that are all now handled by the `Molecule` class.

Simplifies much of the attribute creation logic and code. All potential attributes are initially computed when reading in the file, and set as annotations on the `AtomArray`. 

Then there is just a generic step where all numeric annotations are stored on the mesh as named attributes. This additionally then supports any generic numerical attributes into the future.

New import methods currently don't support:
- [x] building assembly on import or with code
- [x] centering on import
- [ ] adding additional styles
- [ ] using selections when adding styles


molecularnodes/__init__.py
```diff
@@ -1,7 +1,8 @@
 from . import assets, blender, color, session, ui
 from .assets import template
-from .entities import fetch, parse
+from .entities import Molecule, Trajectory
```
zachcp, line 3: love this!

molecularnodes/entities/__init__.py
```diff
@@ -1,9 +1,3 @@
-from . import molecule, trajectory
-from .base import EntityType, MolecularEntity
 from .ensemble import CellPack, Ensemble, StarFile
-from .molecule import BCIF, CIF, PDB, SDF, Molecule
-from .molecule.pdb import PDB
-from .molecule.pdbx import BCIF, CIF
-from .molecule.sdf import SDF
-from .molecule.io import fetch, load_local, parse
+from .molecule.base import Molecule
```
zachcp, line 2: again. this simplification is great.

molecularnodes/entities/molecule/reader.py
```diff
@@ -0,0 +1,196 @@
...
+from pathlib import Path
+from io import BytesIO
+from biotite.structure import AtomArray, AtomArrayStack
+from biotite import structure as struc
+import numpy as np
+from ...assets import data
+from ... import color
+
+
+class ReaderBase(metaclass=ABCMeta):
```
zachcp, line 11: I love this.

molecularnodes/entities/molecule/reader.py
```diff
@@ -0,0 +1,196 @@
...
+            try:
+                self.array.set_annotation(key, func(self.array))
+            except Exception:
+                pass
+
+    def get_structure(self) -> AtomArrayStack | AtomArray:
+        raise NotImplementedError("Subclasses must implement this method.")
+
+
+def _compute_mass(array):
```
zachcp, line 54: style suggestion: add all of these methods to the class.  Are they used elsewhere?

molecularnodes/entities/molecule/base.py
```diff
@@ -60,15 +49,108 @@ def __init__(self, file_path: Union[str, Path, io.BytesIO]):
...
+        file_path : Union[Path, io.BytesIO]
+            The file path to the file which stores the atomic coordinates.
+        """
+        if isinstance(file_path, io.BytesIO):
+            self._reader = pdbx.PDBXReader(file_path)
+        else:
+            if isinstance(file_path, str):
+                file_path = Path(file_path)
+
+            match file_path.suffix:
```
zachcp, line 82: nice

molecularnodes/entities/molecule/base.py
```diff
@@ -60,15 +49,108 @@ def __init__(self, file_path: Union[str, Path, io.BytesIO]):
...
+        self._create_object()
+
+    @property
+    def code(self) -> str | None:
+        """
+        Get the code for the molecule.
+        """
+        return self._code
+
+    def _read(self, file_path: str | Path | io.BytesIO) -> None:
```
zachcp, line 67: Question: why not just load in the data? E.g. instead of returning a handle,  why not just create the AtomArray with the fields of interest. You could optionally set the handle as well. Seems like an extra bit of indirection

molecularnodes/entities/molecule/base.py
```diff
@@ -60,15 +49,108 @@ def __init__(self, file_path: Union[str, Path, io.BytesIO]):
...
-        self.file: str
-        self.array: np.ndarray
-        self._frames_collection: str | None
+        self._frames_collection: str | None = None
+        self._code: str | None = None
         self._entity_type = EntityType.MOLECULE
+        super().__init__()
+        self._read(file_path)
+        self.atom_array: AtomArray | AtomArrayStack
+        self._create_object()
```
zachcp, line 58: I think you are using self.tree in other locations? Maybe it would be helpful to declare it here as well.

zachcp: whoops.  no you are using self.object

molecularnodes/entities/molecule/base.py
```diff
@@ -60,15 +49,108 @@ def __init__(self, file_path: Union[str, Path, io.BytesIO]):
...
+        """
+        file_path = download.download(
+            code=code, format=format, cache=cache, database=database
+        )
+        mol = cls(file_path)
+        mol.object.mn["entity_type"] = "molecule"
+        mol._code = code
+
+        if centre:
+            mol.position + -mol.centroid(centre)
```
zachcp, line 138: might want to declare `mol.position` as an attribute at top. 

Otherwise it might be `mol.object.postion`?

molecularnodes/entities/molecule/base.py
```diff
@@ -156,171 +192,130 @@ def n_models(self):
...
-                array = array[:, mask]
-            else:
-                array = array[mask]
-
-        obj, frames = _create_object(
-            array=array,
+            is_stack = False
+            array = self.atom_array
+
+        self.object = databpy.create_object(
```
zachcp, line 301: I'd like to reiterate how much easier this is to follow the MN logic.  :thumbs-up:

molecularnodes/entities/molecule/base.py
```diff
@@ -338,7 +333,7 @@ def assemblies(self, as_array=False):
             transformation matrices, or None if no assemblies are available.
         """
         try:
-            assemblies_info = self._assemblies()
+            assemblies_info = self._assemblies
```
zachcp, line 336: Maybe add _assemblies as a top level attribute in Init.

tests/test_nodes.py
```diff
@@ -64,7 +65,8 @@ def test_selection_working(snapshot_custom: NumpySnapshotExtension, attribute, c
 @pytest.mark.parametrize("code", codes)
 @pytest.mark.parametrize("attribute", ["chain_id", "entity_id"])
 def test_color_custom(snapshot_custom: NumpySnapshotExtension, code, attribute):
-    mol = mn.entities.fetch(code, style="ribbon", cache_dir=data_dir)
+    mol = mn.Molecule.fetch(code, cache=data_dir)
+    mol.add_style("ribbon")
```
zachcp, line 69: I prefer `Set style` to `Add_style` .  `Add_style` implies more than one style.

BradyAJohnston: I'm going with `add_style()` so that we can add multiple styles on the same Molecule, which will just be additional branches in the node tree. Currently calling `add_style()` once will be equivalent to the previous style setyp.

zachcp:
Looking good!  I think the only thing still broken  is that the docs are using `fetch` which has been moved:

```
 File "/Users/runner/work/MolecularNodes/MolecularNodes/.venv/lib/python3.11/site-packages/quartodoc/builder/blueprint.py", line 177, in get_object_fixed
    raise WorkaroundKeyError(
quartodoc.builder.utils.WorkaroundKeyError: Cannot find an object named: fetch. Does an object with the path molecularnodes:fetch exist?
```

Can we leave the other 2 issues for a different PR?

```
 adding additional styles
 using selections when adding styles
```

zachcp:
#771  attempts to address the issues I mentioned above.

BradyAJohnston:
Yes other items will be left for additional PRs, so far everything should mostly be consistent otherwise.

zachcp:
Merge?

BradyAJohnston:
feels good

CleanShot 2025-03-11 at 11 45 01@2x

PR #771, opened 2025-03-09
Fix Doc Build

zachcp:
There are several broken aspects of the Docs due to API changes. 

##  Problem 1:  Imports

- fetch/parse are now part of `Molecule`
- fix: commented out for now

##  Problem 2:  Molecule

- issue in imports
- I would recommend leaving this out until the API is stable or it will keep breaing.

##  Problem 3: Materials

- the materials aren't available
- is there a change to how Materials are loaded?


BradyAJohnston:
I removed the Molecule-level methods for getting & setting materials, as it will need to change going forward to support multiple styles.

Issue #770, opened 2025-03-09
Update MDAnalysis guesseer


BradyAJohnston:
The API for guessing elements has been updated in MDAnalysis. Currently we get a noisy message when running tests: 

```
   /home/runner/work/MolecularNodes/MolecularNodes/molecularnodes/entities/trajectory/base.py:141: DeprecationWarning: `guess_atom_element` is deprecated!
  `guess_atom_element` will be removed in release 3.0.0.
  MDAnalysis.topology.guessers is deprecated in favour of the new Guessers API. See MDAnalysis.guesser.default_guesser for more details.
    else mda.topology.guessers.guess_atom_element(x)
```

This is the offending code and needs to be updated before MDAnalysis 3.0.0:

https://github.com/BradyAJohnston/MolecularNodes/blob/efbadb6f379d2c5a4dd89ab14cb305c8bcb163c2/molecularnodes/entities/trajectory/base.py#L132-L147

PR #773, opened 2025-03-12
- Update MDAnalysis guesser

PardhavMaradani:
- Fixes BradyAJohnston/MolecularNodes#770

### DIFF
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -135,10 +135,11 @@ class Trajectory(MolecularEntity):
             return self.atoms.elements
 
         try:
+            default_guesser = mda.guesser.default_guesser.DefaultGuesser(None)
             guessed_elements = [
                 x
                 if x in data.elements.keys()
-                else mda.topology.guessers.guess_atom_element(x)
+                else default_guesser.guess_atom_element(x)
                 for x in self.atoms.names
             ]
             return np.array(guessed_elements)


PardhavMaradani:
Hi Brady, submitting this PR as part of the process defined after the pre-proposal acceptance.

BradyAJohnston:
Thanks for looking into this! There are still some warnings about guessers in the logs (see below) but I think this might be from the parsing of some topology files by MDA, as I don't think there are other usage of any guessers explicitly by MDA. 

Brings the number of warnings down from 500K to 8K during testing which is much nicer.

```
tests/test_trajectory.py: 8040 warnings
  /Applications/Blender.app/Contents/Resources/4.4/python/lib/python3.11/site-packages/MDAnalysis/guesser/default_guesser.py:255: PendingDeprecationWarning: Unknown masses are set to 0.0 for current version, this will be deprecated in version 3.0.0 and replaced by Masse's no_value_label (np.nan)
    warnings.warn(
```

PardhavMaradani:
Hi Brady, yes, I saw these warnings in my local test run. However, they were present before the fix too and I looked at the MDAnalysis tests and didn't see a way this could be fixed, so I thought this was an intended warning as described under [guess_atom_mass](https://docs.mdanalysis.org/2.8.0/documentation_pages/guesser_modules/default_guesser.html#MDAnalysis.guesser.default_guesser.DefaultGuesser.guess_atom_mass) and hence didn't explore further.

BradyAJohnston:
Yeah I think it's a warning from upstream. Thanks again!

PR #774, opened 2025-03-13
Consolidate `Molecule` fn interface

zachcp:
The `mn.Molecule` class is shaping up to be the primary class around which loading, modifying, and converting data is happening. After the recent large reorg in #766 its worth trying to simplify that class a bit. I noticed that there are 2 fns - `create_object` & `_create_object`.: the first calls the second.  Furthermore, the `Cellpack` file uses `_create_object`.   

To simplify reuse I have:

- moved the `create_object` and dependent `atom_array_to_named_attributes` to its own utility file.  
- updated Molecele to use those
- updated CellPack to use those.

Note: it seems that something similar might be useful for mesh data. In this case these FNs convert:

- biotite--> vertices

We also have a case of:

- MdAnalysis -> Mesh.

### DIFF
--- a/molecularnodes/entities/ensemble/cellpack.py
+++ b/molecularnodes/entities/ensemble/cellpack.py
@@ -7,6 +7,7 @@ from databpy import AttributeTypes, BlenderObject, store_named_attribute
 
 from .base import Ensemble
 from ..molecule.base import Molecule
+from ..utilities import create_object
 from ... import blender as bl
 from ... import color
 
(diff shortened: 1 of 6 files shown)


BradyAJohnston:
I'm open to suggestions and discussion around this, but as you can see from the tests failing, the `_create_obejct` is used elsewhere as a more generic method for creating objects from biotite arrays. If you can think of a nicer way to implement it then collapsing down into one function would be nicer.

BradyAJohnston:
LGTM. 

As for the MDAnalysis backend, I think having a function separate to the class that takes a Universe and creates a mesh representation is a good idea. That would be used for the initial setup but could also be used on it's own. Would have to be separate to this function that, I don't think it's worth trying to unify their interfaces currently given how different they are. Something to tackle elsewhere though.

BradyAJohnston (approved):

PR #775, opened 2025-03-13
WIP: Materials API

BradyAJohnston:
This brings the `TreeInterface` to materials, allowing for better programmatic access to properties inside of a node tree for a material. Rather than having to index into the tree, find the node and then change the default value, values are exposed on a class that is created for that material.

Currently all interfaces are build dynamically when requested. The example of the `MN Ambient Occlusion` material that is included as a default currently (pictured below), produces the list of properties:

image

```py
['ambient_occlusion_color',
 'ambient_occlusion_distance',
 'ambient_occlusion_normal',
 'emission_color',
 'emission_strength',
 'emission_weight',
 'links',
 'material',
 'math_value',
 'mix_a',
 'mix_b',
 'mix_factor',
 'node_tree',
 'nodes']
```

The `links`, `nodes` and `node_tree` currently all access things to do with the node tree itself, but all the rest are created based on node names and their input names.

This doesn't work that great, because `mix_*` isn't a particularly descriptive name. If a node is given a label, then it should default to instead using the label for property creation. In the creation of materials it we should be instead adding labels to the nodes to better improve the final interface.

Currently the interface works so that if a value is requested from a socket that is currently linked, an error is raised to say that the `default_value` has no use or sway over the final material as the socket is linked. 

If a value is given for a socket is linked, then the socket is removed and the value set. If another socket is given, then a link is created between the given socket and the current property.

--- 

It would potentially be more useful to define a subset of sockets & name them, to expose a limited set of options for a material to the user. In the above example, there is not really any point in exposing `mix_a` and `mix_b` when really only the factor between the two is what we want to allow the user to interface with.

I've focused on creating a general solution first. In the future we can maybe provide a subset of node & socket names, that defines an API for a node group, but I think I'll leave that for later and instead go with this more general (but less ideal) API to begin with.

```py
import molecularnodes as mn
mn.register()
mn.assets.install()
can = mn.Canvas()
can.scene_reset()
can.resolution = (860, 540)
can.render_engine = "EEVEE"
mol = mn.Molecule.fetch("8H1B").add_style("surface", material = mn.material.AmbientOcclusion())
can.frame_object(mol)
can.snapshot()
```

image

```py
mol.styles[0].material.ambient_occlusion_distance = 0.1
can.snapshot()
```

image


BradyAJohnston:
The same machinery could be used for defining separate interfaces to different styles in the same node tree @PardhavMaradani

PardhavMaradani:
Hi Brady, my quick thoughts on this:

> The particular approach of mol.material will need to change in the future to better support multiple styles per entity, but currently running with this to prototype the node tree interface.

Materials should eventually be tied to styles (like how it is already in the various style nodes today) and not the whole object, right? I believe this is what you were referring to.

One potential direction to explore is to use shader attributes. This will avoid the problem of having several different materials (all of which are parameter changes to the packaged materials and not necessarily underlying shader node tree changes). This also has the added advantage that these attributes can be easily pulled up into any GUI - so any customizable material settings are right below the material selection (Just like how all style specific settings are below the style selection).

The `iso positive` and `iso negative` in the density visualization example are such attributes that are passed over from the geometry node tree to the shader node tree - changing these values through the API/GUI pass the values to the shader node tree.

Here is an example of how they show up in the GUI (shown under modifier settings here):

shader_attributes_passing

and how they are configured through the API:

```
# Add component with custom control values
bu2.clear_components()
bu2.add_component_density(
    SAMPLE_APBS_VDB,
    options={
        "isoType": "Int",
        "isoDefault": 1,
        "isoMin": 1,
        "isoMax": 100,
        "isoPositiveDefaultColor": (0, 0, 1, 1),
        "isoNegativeDefaultColor": (1, 0, 0, 1),
        "sliceXDefault": 0.4,
        "sliceXMin": 0.4,
        "sliceXMax": 0.62,
    },
)
```

One other alternative to having individual getters/setters is to pass a json/dict (like in the example above) - maybe making it a bit more data driven.

tylerjereddy:
> would appreciate thoughts for this kind of approach in defining a 'public' API for a node group

This appears to be a step toward scriptable assigning and modification of materials to objects, which I'm obviously in favor of based on my downstream experiments in https://github.com/yuxuanzhuang/ggmolvis/pull/23.

> This isn't ideal, and having a .copy() method that duplicates the material and looks up the new version is maybe the way to go, but for initial version I didn't want 500 copies a material spawning.

I'd suggest considering [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) semantics--if a material is modified in any way, then the object modifying it could point to a copy that is only created as needed upon mutation. `pandas` recently adopted copy-on-write semantics so you might even be able to "borrow" a bit from what they did rather than reinventing the wheel. I suspect many reasonable use cases may only modify a subset of materials to emphasize mechanistic events or particular parts of a system, and much of the rest of the system may point to "built-in" materials without needing to modify/make copies.

> We don't necessarily want to expose every single option defined within a node tree for an API

It might be helpful to offer one of two options: 1) the user is able to provide their own custom material object that is fully modified to their liking (in more advanced cases); 2) a way for the user to access the "raw" material object at their own peril for deeper customization. Similar to my original PR above, I like the first option personally, perhaps alongside adoption of a standard "format" for the users/community to provide/read in the materials (not sure if it would be popular enough to warrant a library/plugin repo for additional mats).

On that note, for

> The node groups could be pre-made in a .blend file or built via a script then wrapped in these interface classes

I'll again indicate, similar to my PR above, that I'd much prefer to see the infrastructure in raw source rather than binary format wherever feasible, so it is easy to see/regenerate, etc. This is also especially useful for the concept of custom materials "plugins--" it seems to me like the best place to start for designing your own custom material is to have access to the raw source code that generates the material closest to your target. Even for moderate adjustments to currently available-by-default materials, a user may still want to create their own custom/plugin material to avoid repetition/facilitate sharing.

Finally, a loosely-related comment, and something I know you're aware of from https://github.com/yuxuanzhuang/ggmolvis/issues/24, is that seeing the `Canvas` style workflow sample above and comparing/contrasting to when I'd want to use that vs. `ggmolvis` is a bit confusing. It is what it is I guess--thing are obviously in flux.

BradyAJohnston:
> Materials should eventually be tied to styles (like how it is already in the various style nodes today) and not the whole object, right? I believe this is what you were referring to.

Yes this is what I was referring to. The wrapper class I've got working now could also be used to get & set attributes for individual styles as well. Unsure how they should be accessed though `mol.styles[2].width = 2`?

> One potential direction to explore is to use shader attributes. This will avoid the problem of having several different materials 

I've thought about this & still open to the idea, but that would result in a rather large material that would be hard for users to try and interface with. I think from a GUI perspective as well it would make it harder to create a custom material for a user. Open to more discussion around this.

> The iso positive and iso negative in the density visualization example are such attributes that are passed over from the geometry node tree to the shader node tree

The `Color` attribute is already one that is passed from GN to material, but setting up that bridge to import more attributes I think is worth while as well.

> One other alternative to having individual getters/setters is to pass a json/dict 

I think having an option to injest a json/ dict of values is a good thing to have, but I wouldn't want it to be the only approach. Speaking mostly for myself but I am useless at writing code without auto-complete, especially if using an API for the first time, so being able to do `mol.material.` and press tab to see what options are available is a must I would say. That approach also helps avoid spelling mistakes which are common (for me) in strings for a dictionary.

> I'd suggest considering [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) semantics--if a material is modified in any way, then the object modifying it could point to a copy that is only created as needed upon mutation.

Thanks for pointing me this way, I'll have a look into it and see how it might be applied. I think that would be the way to go as well.

> 2) a way for the user to access the "raw" material object at their own peril for deeper customization. 

The actual `bpy.types.Material` is accessible currently though `mol.material.material`, which a user is then free to change & edit via `bpy`. This is intended just a convenience interface

> I'll again indicate, similar to my PR above, that I'd much prefer to see the infrastructure in raw source rather than binary format wherever feasible, so it is easy to see/regenerate, etc. 

I agree with you on this one, it just requires converting the node trees into python scripts manually which I haven't had the time to do. It's much, _much_ quicker to prototype via the GUI, but storing in a script is better for version control and visibility. This would still be made much simpler by a good scripting API for nodes (such as geometry-script etc have attempted) rather than wrangling all of the connections manually.

> seeing the Canvas style workflow sample above and comparing/contrasting to when I'd want to use that vs. ggmolvis is a bit confusing. It is what it is I guess--things are obviously in flux.

This is kind of the point I was trying to make in that post. I'm focused on developing the scripting API of Molecular Nodes, which needs it's own scene management & rendering tools like the `Canvas`. Still in flux & in initial prototyping stages, but it would then lead to confusion if other packages are wrapping / creating their own scene management tools.

zachcp:
At  a high level I really like where this is going. Just a few comments:

- +1 for small, simple API on top of a small number of classes
- +1 for some sort of immutability (whether copy-on-write or [immutable data classes](https://zachcp.github.io/prettymol/materials.html) or other )
- +1 ultimately getting to a code-only version of the style nodes.

PardhavMaradani:
> I've thought about this & still open to the idea, but that would result in a rather large material that would be hard for users to try and interface with. I think from a GUI perspective as well it would make it harder to create a custom material for a user. Open to more discussion around this.

I think we could keep the node trees created programatically independent of the ones created by users through the GUI with the geometry/shader node editors. I am not sure if we really want users to tinker with the programatically generated node trees in the GUI (as Tyler also pointed out in a different context before) - though we might seem to lose some flexibility here, we might gain in terms of a more user friendly GUI (in a panel) to manage these in one place. These wouldn't replace the existing node trees for styles/materials, so maybe this shouldn't be a problem?

> I think having an option to injest a json/ dict of values is a good thing to have, but I wouldn't want it to be the only approach.

I agree

> This would still be made much simpler by a good scripting API for nodes (such as geometry-script etc have attempted) rather than wrangling all of the connections manually.

My experience with the `Node To Python` extension after having created the required node trees manually in the GUI (for styles, density materials, density node tree etc) is that it works well, but with some caveats, which required some manual fixing after auto-generating the code. They use the literal node names (with the `.001` etc), use indices instead of names (eg: `node[0]` instead of `node["Geometry"]`) etc, but probably not something that can't be fixed in a fork.

BradyAJohnston:
>My experience with the Node To Python extension 

That has been my experience with it as well. So far there still isn't a solution out there that works well. The files that `Node to Python` produce would only be useful for tracking if a change happens, they aren't useful to using as a base to build anything new or modifying via code itself. 

There is certainly nothing currently available that would be feasible to turn the more complex Geometry Nodes node groups (like the style nodes) into python code. One day I'll have time to dedicate towards developing it but not currently.

PardhavMaradani:
Here is a very quick prototype of what I meant by having the material customization at the same place as styles - for others who haven't seen a potential GUI option before: (just a couple of material options for now)

https://github.com/user-attachments/assets/be4a9091-ae3e-4960-b2a5-5c1c5beb9399

Adding this here for future reference. I think it would help to try and keep the API in sync considering a potential GUI mode so they don't become completely independent efforts at a later point.

BradyAJohnston:
Thanks for the example, I think using this simplified wrapper class for accessing attributes of a material or node group is still the way to go. In terms of GUI it could either reach past it or the class could include a method that takes a `UILayout` and sets up the GUI from the node group.

Definitely need to keep them aligned though.

zachcp:
@BradyAJohnston  How are you feeling about the current state of this PR? Almost ready for a merge?

BradyAJohnston:
@zachcp I think it's pretty close. I mostly just want to define some of the default materials with the new changed interface and then I think it is ready to go (plus I have to update PR with changed details)

PR #779, opened 2025-03-14
API for adding & controlling multiple styles

BradyAJohnston:
This changes how styles are added and dealt with inside of MN.

## Old

Previously a default node tree was setup inside of MN on import, with the rest of the customization left to the user, intended to be used via the GUI. This left customization of more complex styles to be difficult via scripting, and templates for more complex styles difficult to show off to the users.

## New

To address this, there are several additions being made in this PR. 

### `add_style_branch()`

This adds a new "branch" of an existing node tree, with a style & potentially a color node on it. This branch is equivalent to a "representation" inside of MolViewSpec (#787). A specific style is created for some subset of atoms, with a color & selection specified. This is expose on the `Molecule` class under `Molecule.add_style()`. Not yet implemented for the `Trajectory`, but can be ported relatively easily as they don't differ in terms of node tree handling, but will differ in terms of selections.

Color & selection specification are currently basic. For colors, `None` doesn't add a _Set Color_ node and the style will source the colors from the underlying `Color` named attribute. If a `str` is given, then a named attribute with that named is attempted to be looked up and used. If a list / vector is given, then those values are applied to the `Color` input of the `Set Color`. For selections a similar approach is used with named attributes or `None` being no selection applied.

More complex color & selection specification will be addressed in another PR.

To add multiple styles, the user can repeatedly call the `mol.add_style()` function which creates a new branch. The existing branches can be accessed via the `mol.styles` list, with newly added styles being prepended to the start of the list. To remove a style, call the `mol.styles[0].remove()` method which removes the associated nodes from the tree. The `add_style()` method returns the molecule object, so multiple `add_style()` calls can be chained together.

```py
import molecularnodes as mn
mol = mn.Molecule.fetch("4ozs")
mol.add_style("cartoon", selection="is_peptide")
```

CleanShot 2025-03-20 at 17 15 52
CleanShot 2025-03-20 at 17 16 36

```py
mol.add_style("spheres", selection="is_nucleic", color=(0, 0.3, 0.8, 1.0))
```

CleanShot 2025-03-20 at 17 17 19

```py
mol.styles[1].remove()
```

CleanShot 2025-03-20 at 17 17 37

```py
(
    mol
    .add_style("cartoon", selection="is_peptide")
    .add_style("surface", selection="is_nucleic")
)
```

CleanShot 2025-03-20 at 17 18 47

To handle the programmatic creation & deletion of nodes, a function for arranging node graphs nicely has been implemented. This was inspired & partially taken from [geometry-script](https://github.com/carson-katri/geometry-script) for node graph traversal and arrangement. Improvements need to be made for getting field nodes better aligned with their geometry nodes, but for now as a method mostly intended for API access (and not visual inspection) this is good enough for keeping things clean & tidy.

## Changing Styles

Similar implementation to in #775 , siloed & programmatic access to only a subset of node tree options is provided via the `NodeTreeInterface` class. For each `Style *` node connected to the `Join Geometry`, a new class is built that exposes all of the options which don't have another node connected (no point changing values of a socket with a connection as the connection has priority) are exposed as class properties. We can index into the style list and change their values via the new API:

```py
mol.styles[0].probe_size = 1
```

CleanShot 2025-03-20 at 17 34 07

```py
mol.styles[0].probe_size = 3
mol.styles[0].scale_radii = 2
```

CleanShot 2025-03-20 at 17 34 33

Similar to other approaches in MN, none of these values are stored as intermediates, we are directly getting & setting the values by looking up the nodes inside of the node trees themselves.

---
There are plenty of improvements to be made, especially with better color & selection specification (such as #760 ). 

Potential current changes to overall design would maybe to have a `StyleManager` class. 

This could then maybe be used for adding and remove rather than a method and a list, but I would appreciate feedback on these ideas.

```py
mol.styles.add("cartoon")
mol.styles[0].width = 3
mol.styles[0].remove()
```


zachcp:
@BradyAJohnston . My main thinking is this is coming along great.  However,  I do have some differences of thinking in terms of implementation. Personally, I think manipulating node trees is something that should be hidden as an implementation detail.   I think the route above is going to end up generating tangled webs over time and I would suggest a modification to allow something like the following:

- 1 standard, paramaterizeable node tree per selection.
- allow the API to create or modify these trees at a high level but don't target connections between nodes.
- create an API for all standard colors and materials that act on the selection. 

Potential additional Benefits: 

- 1 modifier per selection will allow you to extend the UI for selections. e.g. to hide/show selections.
- if nodes are standard and selections are separate you could also imagine adding UI elements like Pyoml - e.g. to change color etc. right now that is done within Geometry Nodes but you can imagine dramatically simplifying the UI.
- 

image

BradyAJohnston:
@zachcp thanks for the feedback. I agree that it would be nice to have one modifier per selection & to keep it clean, but that's not currently possible with Blender & geometry nodes. Each modifier & thus node tree only has access to the result of the previous modifier / node tree. After the first node tree you would have the geometry of the style and not the original `atoms`. With the regular setup it would break after the first modifier.

It would be possible to use `Join Geometry` and `Separate Geometry` to pass the original mesh into the output & then operator on just that on the next modifier and I have thought about this as an approach. I think passing around extra meshes and involving multiple modifiers would also get messy.

I've asked the Blender devs about a node that gives access to the "unmodified" mesh which would allow for this workflow and they have considered it but it's not currently possible and not currently a priority.

There isn't a nice 'clean' approach regardless of what we do, which is why it has taken me so long to even try and approach this programmatically rather than just saying it's up to users to modify node trees via the GUI.

I think the approach of going for a single node tree is ultimately better because it aligns with the same approach via the GUI. If a user runs a script and then opens inside of Blender they see a node tree similar to what they would create manually.

The potential for a "tangled web" is a real valid one. Ensuring that each style is a single 'branch' dramatically simplifies this already. Making improvements to the layout / arrangement algorithm to better structure & arrange everything would be beneficial to this but also for other aspects of programming with nodes so I think its worth putting more effort into.

> allow the API to create or modify these trees at a high level but don't target connections between nodes.

> if nodes are standard and selections are separate you could also imagine adding UI elements like Pyoml - e.g. to change color etc. right now that is done within Geometry Nodes but you can imagine dramatically simplifying the UI.

I agree that having to try and modify links & connections between nodes is a bit of a nightmare through code, but I don't think it should be entirely avoided. A common point of feedback & request from people is to "simplify the UI" and obfuscate the nodes away to a more familiar GUI. 

I understand this, but at the same time, the nodes are kind of _the point_. The whole reason Molecular Nodes can do what it can do so well, is _because of the nodes_. It's modular and extensible and customisable _because of the nodes_. It's a balance between trying to make a familiar UI and trying to onboard new users who haven't worked with node trees before, but I'm firmly of the opinion that some things should be made easier, but ultimately we should be trying to encourage users to 'dive in' to the nodes, because that is where the real power is.

The API will obfuscate any direct manipulation of node trees, and ideally there will be no need for a user to do `import bpy` or `bpy.*`, but I think we should still be manipulating node trees. As long as we are structured about the approach, to ensure things are clean (like one branch per style, joined at the end. We can just delete a whole branch and any orphaned nodes for cleanup), then it shouldn't get too messy internally (at least I hope).

zachcp:
Thanks @BradyAJohnston . One other thought is that you could create selections as special MN objects which use the same data under the hood but has its own node tree.   The below, extremely simple and possible misleading example followed by an illustration. You could have a true "Base Object" plus one or more selections that have the same properties as above.



```python
# Clear existing objects
bpy.ops.object.select_all(action='SELECT')
bpy.ops.object.delete()

# Create our dataset (a simple cube)
bpy.ops.mesh.primitive_cube_add(size=1, location=(0, 0, 0))
original_cube = bpy.context.active_object
original_cube.name = "OriginalCube"

# Create shallow copies (sharing the same mesh data)
scaled_cube = bpy.data.objects.new("ScaledCube", original_cube.data)  #. <------ Same mesh data
scaled_cube.location = (-3, 0, 0)
bpy.context.collection.objects.link(scaled_cube)
```


image

BradyAJohnston:
This was an approach that I considered originally as well. I put forward this approach of having different objects sharing the same mesh data here: https://github.com/yuxuanzhuang/ggmolvis/issues/11

The reason I was ultimately convinced in that thread to instead do one object / node tree with multiple branches is that it is more aligned with how MN works via the GUI. Introducing different objects and different node trees means more code to deal with all of that, as well as the disconnect between that setup and how a user would work creating different styles in a single node tree.

I think this approach still has merit though, and should even maybe still be used somewhere but I'm not entirely sure where or how. My feeling is that it will be easier to implement once Geometry Nodes has better support for passing data back and forth between objects / node trees. Certainly even from a GUI perspective having different objects (and being able to animate / keyframe them) makes sense for some approaches.

I've been pretty set on ensuring there is "one object for each molecule", but I suppose what we really need is only "one mesh per molecule", with the potential for multiple objects. The underlying MN code isn't setup for handling multiple objects currently (just `mol.object`) but this could change in the future with multiple objects representing the same Molecule (sharing the same mesh data).

zachcp:
Roger that!

PR #783, opened 2025-03-17
Fix Selection from AtomGroup after refactoring Selection

yuxuanzhuang:
Currently, the `add_selection_from_atomgroup` function is broken due to a refactor of the `Selection` class.

After the fix, its behavior remains different from before: it now only stores booleans for filtering, whereas previously, the selection mask was simultaneously applied to the Trajectory. I am wondering what is the current API for applying the mask  @BradyAJohnston


BradyAJohnston:
The method `selection.set_selection()` stores the selection as an attribute on the trajectory / object. This is called on every frame update, but should also be called whenever you change / set the `selectrion_str`. It is also being called at the end of the `add_selection_from_atomgroup()`. 

Is it not currently working?

yuxuanzhuang:
The issue arises because `updating=False` is set when selecting from an atomgroup and it prevents the selection from being set https://github.com/BradyAJohnston/MolecularNodes/blob/487621d3c593b3ee7914a6261637c0b6a2db3016/molecularnodes/entities/trajectory/selections.py#L110

Manually calling `set_boolean` resolves the problem, and everything is now working fine.

BradyAJohnston:
Thanks for fixing this!

Issue #792, opened 2025-03-28
Dev Feature Request: Use of a command runner


zachcp:
**Is your feature request related to a problem? Please describe.**
Difficulty for new developers to get started with MN

**Describe the solution you'd like**
Consolidating common build and test scripts in a single, standard place.

Hi Brady,

I'm wondering if you'd be interested in consolidating around a command runner to assist potential developers/ contributors run common development tasks.  For other projects I have used [pixi](https://pixi.sh/latest/), [just](https://just.systems/man/en/) or [make](https://www.gnu.org/software/make/manual/make.html).  If you have a preference I would be happy to use any of them. Here are a set of commands I think would be useful.

```sh
MolecularNodes % just -l

Available recipes:
    build              # Builds the MN Extension
    document           # Creates the Website. NOte: requires `quarto`
    install-in-blender # Installs MN as a local Blender Extension
    run-jupyter        # run Blender with with a Jupyter connecttion
    test               # Runs the testsuite
    update-deps
```

```Just
# Core commands for Developing Molecular Nodes
#
# try `just -l` in your shell
#

# Builds the MN Extension
build:
    echo 'Building Python Packages for Molecular Nodes Extension'
    uv run build.py


# Creates the Website. NOte: requires `quarto`
document: update-deps
    echo 'Creating the UV docs'
    cd docs && \
    uv run generate.py && \
    uv run quartodoc build ** \
    cd .. && \
    uv run quarto render docs

# Installs MN as a local Blender Extension
install-in-blender:
    echo 'Creating the UV docs'
    blender -b -P tests/python.py -- -m pip install ".[test]"
    blender -b -P tests/run.py -- -vv tests --cov --cov-report=xml

# run Blender with with a Jupyter connecttion
run-jupyter:
    uv run jupyter lab # needs tweaking to work

update-deps:
    echo 'Updating Molecular Nodes Dependencies'
    uv sync --all-extras

# Runs the testsuite
test:
  echo 'This is another recipe.'
  uv run pytest
```

BradyAJohnston:
I would be happy to organise around something. 

There is a project `blext` which shows a lot of promise for this exact purpose, but is still a bit early on in development: https://codeberg.org/so-rose/blext

I can certainly see that taking on the role of managing development tasks.

Unsure if we should use another system in the mean time, or try adopting that & attempting to use and improve it.

scientistnobee:
[...]

PR #793, opened 2025-03-28
Justfile

zachcp:
Implementation for #792 .

For illustration.


zachcp:
Seems like blext is a better choice for this project. Closing.

PR #794, opened 2025-03-30
Try BLEXT

zachcp:
Initial thoughts:

- rough around the edged.
- if it matures as advertized it will dramatically simplify development.


zachcp:
Superceded by #814 and #798

Issue #796, opened 2025-04-01
AttributeError: Writing to ID classes in this context is not allowed


PardhavMaradani:
This is a strange issue. This happens frequently enough in one set of conditions, but not necessarily consistent enough to narrow down exactly what causes it.

I have a `bpy.props` property that represents the frame of a trajectory. In the set callback of it, I just call the trajectory's `.frame` setter. I have the `Update with Scene` turned off for this trajectory. Keyframing the property works fine - both in the 3D viewport and in renders - reflecting the updated positions. When rendering the animations, occasionally (at different frames each time), the following error is thrown:

```
Traceback (most recent call last):
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/handlers.py", line 51, in update_entities
    entity.set_frame(frame_to_set)
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 590, in set_frame
    self._frame = self.frame_mapper(frame)
    ^^^^^^^^^^^
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 498, in _frame
    self.object.mn.frame_hidden = value
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Writing to ID classes in this context is not allowed: u1, Object datablock, error setting MolecularNodesObjectProperties.frame_hidden

```

I've ran into a similar error in the past when trying to update a property in the `draw` context (when drawing the UI layout). I suspect something similar is happening here.

Looking through the code, it looks like the `_frame` setter is for the `mn.frame_hidden` property that is only shown in the UI and that too only when the `Update with Scene` option is enabled. So, the following change resolved the above errors for my case:

```
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -578,7 +578,8 @@ class Trajectory(MolecularEntity):
         Update the positions, selections and calculations for this trajectory, based on
         frame number of the current scene, not the frame number of the Universe
         """
-        self._frame = self.frame_mapper(frame)
+        if self.update_with_scene:
+            self._frame = self.frame_mapper(frame)
         self._update_positions(frame)
```

[...]

PR #797, opened 2025-04-01
- Set frame_hidden property only when 'Update with Scene' is enabled

PardhavMaradani:
- Fixes BradyAJohnston/MolecularNodes#796

### DIFF
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -578,7 +578,8 @@ class Trajectory(MolecularEntity):
         Update the positions, selections and calculations for this trajectory, based on
         frame number of the current scene, not the frame number of the Universe
         """
-        self._frame = self.frame_mapper(frame)
+        if self.update_with_scene:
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
I think we need to overall have this a bit clearer. With the current setting, calling the `set_frame()` will have different results depending on whether `traj.update_with_scene` is `True`. The result could be quite different because the frame mapping could give drastically different results. I'm not sure how best to to approach this but I think we need to be as explicit and clear as possible, potentially exposing the option in the `set_frame` call itself?

PardhavMaradani:
Hi Brady, the `_frame` class property points to the `mn.frame_hidden` blender object property, which is only used when the `update_with_scene` option is `True` (to display in the UI panel). So, the current fix should not change any existing behaviour? Am I interpreting this incorrectly? Thanks

BradyAJohnston:
I'm happy with this fix :) I am meaning that in the future I would rather the class itself wasn't using an obfuscated property, but that at the time I couldn't figure out a better way to do it. Something to consider going forward.

PR #798, opened 2025-04-01
Ruff: Rules and Initial Lints

zachcp:
Changes:

1. Add Ruff rules to pyproject.toml. I used biotite's rules as a baseline
2. Add Ruff check to CI tests
3. Fix baseline suggestions. Mostly: imports, unused variables
4.  Put all per-file exceptions in the pypriject.toml

Discussion:

- there are a few style decisions that @BradyAJohnston  will need to make relative to common "errors". The repeat Ruff lints are related to variable naming and relative imports. We can either change those or disable those lints globally or on a per-file basis (current status)


BradyAJohnston:
Thanks for going through and setting this up. As this touches quite a lot of files, I might try and get the materials PR #775 merged as it refactors some file locations and overall setup. Then I can merge and potentially resolve some linting issues. 

I haven't used the linting actions before, but the action that it's currently in runs daily rather than on PRs. Wouldn't it make more sense to have it run on PRs? It could be it's on `.yml`.

zachcp:
Yes on all counts! I can move to its own lint

BradyAJohnston:
I've gone through and merged and cleaned up. The only linting related change was I added something to the ignore which is related global / relative imports (for Blender extensions it needs to be relative imports).

I'm still not as familiar with linting as a whole so if you wanted to be specific with questions around particular lints then we can discuss and decide on them.

BradyAJohnston:
OK I've gone through and addressed remaining lints.

Issue #799, opened 2025-04-02
Errors when creating trajectories without filenames


PardhavMaradani:
It is very common in MDAnalysis use cases to create universes on the fly (with `mda.Merge`, etc). Such universes will not have topology and trajectory filenames and when creating those objects, MN throws errors.

Here is a simple example from the [Aligning a structure to another](https://userguide.mdanalysis.org/stable/examples/analysis/alignment_and_rms/aligning_structure_to_another.html) MDAnalysis user guide:

```
import MDAnalysis as mda
import molecularnodes as mn
from MDAnalysis.analysis import align
from MDAnalysis.tests.datafiles import CRD, DCD, DCD2, PSF

warnings.filterwarnings("ignore")

adk_open = mda.Universe(CRD, DCD2)
adk_closed = mda.Universe(PSF, DCD)

adk_open.trajectory[-1] # last frame
merged = mda.Merge(adk_open.atoms, adk_closed.atoms)

trajectory = mn.entities.Trajectory(universe=merged)
trajectory.create_object(name="merged", style="spheres")
```

This throws the following errors:

```
File ~/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py:465, in Trajectory.create_object(self, name, style)
    462     self.object["atom_type_unique"] = self.atom_type_unique
    464 self.object.mn.entity_type = self._entity_type.value
--> 465 self.save_filepaths_on_object()
    466 bpy.context.view_layer.objects.active = self.object
    468 return self.object

File ~/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py:406, in Trajectory.save_filepaths_on_object(self)
    404 def save_filepaths_on_object(self) -> None:
    405     obj = self.object
--> 406     obj.mn.filepath_topology = str(path_resolve(self.universe.filename))
    407     obj.mn.filepath_trajectory = str(
    408         path_resolve(self.universe.trajectory.filename)
    409     )

File ~/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/blender/utils.py:25, in path_resolve(path)
     23     return Path(bpy.path.abspath(str(path)))
     24 else:
---> 25     raise ValueError(f"Unable to resolve path: {path}")

ValueError: Unable to resolve path: None
```

[...]

PR #800, opened 2025-04-02
Allow trajectories with no filepaths (topology and trajectory)

PardhavMaradani:
- Fixes BradyAJohnston/MolecularNodes#799


BradyAJohnston:
I get the feeling that maybe we should try and solve the usage of `resolve_path` to begin with in a nicer way in the future as well. I'm happy with this in the meantime.

PR #803, opened 2025-04-03
Small NodeGroup typing fix

Andrej730:
I'm not really familar with the codebase, just found this by grepping Github to figure when actually NodeGroup is used.

Changes:

1) custom_iswitch - actually returns NodeTree, not NodeGroup. Added more specific type that actually is used - GeometryNodeTree

2) node_group - also returning NodeTree, not NodeGroup

3) ui_from_node - added more specific GeometryNodeGroup. Oddly enough, in Blender GeometryNodeGroup and doesn't inherit from NodeGroup and typing would complain about passing GeometryNodeGroup though it is the one that's intended here.

### DIFF
--- a/molecularnodes/blender/nodes.py
+++ b/molecularnodes/blender/nodes.py
@@ -780,7 +780,7 @@ def custom_iswitch(
     offset: int = 0,
     panels: Optional[List[str]] = None,
     panels_open: int = 1,
-):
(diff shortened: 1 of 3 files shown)


BradyAJohnston:
Thanks for finding this! It's something that I've slowly making sure I am properly consistent with so I appreciate it.

Issue #812, opened 2025-04-07
Out of bound errors when frame exceeds trajectory frame length


PardhavMaradani:
There are a couple of places where a frame can exceed the number of frames in a trajectory as there are no checks in place. Both these don't show errors in the UI, but show them on the Blender console and in notebook cells.

Here is the setup:

```
import warnings

import MDAnalysis as mda
import molecularnodes as mn
from MDAnalysis.tests.datafiles import DCD, PSF

warnings.filterwarnings("ignore")
u1 = mda.Universe(PSF, DCD)
trajectory = mn.entities.Trajectory(universe=u1)
trajectory.create_object(name="u1", style="spheres")
```

- First case:

Go to `Scene > Molecular Nodes > Object`. `Update with Scene` should be enabled by default. Open up timeline and set the current frame to something like 150. The following error is thrown:

```
Traceback (most recent call last):
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/handlers.py", line 48, in update_entities
    entity.set_frame(frame_to_set)
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 581, in set_frame
    self._update_positions(frame)
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 663, in _update_positions
    self.position = self._position_at_frame(uframe_current)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 587, in _position_at_frame
    self.uframe = frame
    ^^^^^^^^^^^
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 202, in uframe
    self.universe.trajectory[value]
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/tmp/blender-4.3.2-linux-x64/4.3/python/lib/python3.11/site-packages/MDAnalysis/coordinates/base.py", line 835, in __getitem__
    frame = self._apply_limits(frame)
[...]

PR #813, opened 2025-04-07
- Clamp frame to n_frames to prevent oob errors

PardhavMaradani:
Fixes BradyAJohnston/MolecularNodes#812

### DIFF
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -536,7 +536,7 @@ class Trajectory(MolecularEntity):
 
     def _frame_range(self, frame: int):
         "Get the trajectory frame numbers over which we will average values"
-        return frames_to_average(frame, self.average)
+        return frames_to_average(frame, self.n_frames, average=self.average)
 
     def _cache_ordered(self) -> np.ndarray:
         "Return the cached frames as a 3D array, in chronological order"
@@ -576,8 +576,6 @@ class Trajectory(MolecularEntity):
         Update the positions, selections and calculations for this trajectory, based on
         frame number of the current scene, not the frame number of the Universe
         """
-        if self.update_with_scene:
-            self._frame = self.frame_mapper(frame)
         self._update_positions(frame)
         self._update_selections()
         self._update_calculations()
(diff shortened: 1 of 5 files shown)


BradyAJohnston:
I'm a little hesitant to hide away error messages for out of bounds completely. There should be some kind of feedback for a user as to why their trajectory might not be animating how they expect.

PardhavMaradani:
Hi Brady, I have added a label in the UI in the Object panel to show the number of trajectory frames. The `frame` and `frame_hidden` properties will show the actual trajectory frame for the `update_with_scene` `False` and `True` scenarios respectively.

Looking deeper, I noticed several other out of bound cases and also an error in `_update_positions` which is using the `subframes`, `average`, etc values even when the `update_with_scene` is turned off. The frame mapper also doesn't seem to have any validations, especially for the case when a mapping array is passed. I fixed the out of bounds errors in the updated commit and verified it with both MDA and oxDNA trajectories (having multiple frames) - with `Update with Scene` turned off and on with all the sub options of subframes, averaging, offset and interpolation.

BradyAJohnston:
Thanks for diving deeper on this. I'm happy with these changes for now. I'm a little bit still of the opinion that we should be more verbose about if a user (especially via the API) is trying to animate through frames of a trajectory but they are ultimately outside of the range of them. I suppose could just be up to the user to ensure they know about their frame range or not.

BradyAJohnston (approved):

PardhavMaradani:
Hi Brady, I agree. For the API, it looks like there are two different entry points for changing the trajectory frame today, which are mutually exclusive (for the recursion reasons you outlined earlier): The `.frame` property and the `set_frame` method. The first one changes the bpy prop (`mn.frame`) and internally calls the `set_frame` method - this seems to be the preferred way when `update with scene` is disabled (It is easy to add validations and raise exceptions here). The second one gets called in the frame handler and seems more like an internal function. This needs to however be called for the `update with scene` enabled to support the sub frames, averaging, offset etc. There are also a couple of pending issues like updating the `mn.frame_hidden` property which could cause problems and the lack of validation in the frame mapper. So, this definitely needs a slightly larger refactoring, which might be a bit more disruptive (I am also not very clear how this could be done at this point) - something we could possibly take up later?

BradyAJohnston:
It's not something I'm clear on how we should be doing it either, I just know that we should be doing it better than we are currently doing it. I did spend a whole day a week ago trying to refactor it and just ended up going nowhere. Something to deal with in the future and maybe to try and do some higher-level planning beforehand.

PR #814, opened 2025-04-07
Try Blext: Attempt #2

zachcp:
Related to #792 See also:

- [updated issue](https://codeberg.org/so-rose/blext/issues/57)
- [single file example](https://codeberg.org/so-rose/blext/src/commit/f89488ff5652246ae902ead2e63ee383f4031f82/examples/simple_script.py)

```txt
An update: I've merged a rather large batch of changes to main - please expect instability!

In my limited testing, no extraneous wheels are present any more. This is easiest to see when building for a single platform with --bl-platform.

Here's an example of an extension that uses jupyter-core, which only includes cffi and pywin32 when a Windows-based platform is targeted: https://codeberg.org/so-rose/blext/src/commit/f89488ff5652246ae902ead2e63ee383f4031f82/examples/simple_script.py
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -51,6 +51,14 @@ BLExtSpec(
 # you need to use the head
 # see https://codeberg.org/so-rose/blext/issues/57
 uvx --from git+https://codeberg.org/so-rose/blext blext build
+uvx --from git+https://codeberg.org/so-rose/blext blext check # not implemented
+uvx --from git+https://codeberg.org/so-rose/blext blext run
+uvx --from git+https://codeberg.org/so-rose/blext blext show blender_manifest # broken
+uvx --from git+https://codeberg.org/so-rose/blext blext show deps
+uvx --from git+https://codeberg.org/so-rose/blext blext show global_config
+uvx --from git+https://codeberg.org/so-rose/blext blext show profile
+uvx --from git+https://codeberg.org/so-rose/blext blext show path
+
 uv add  git+https://codeberg.org/so-rose/blext
 
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,68 @@
+
+
+## Environment
+
+- [uv](https://docs.astral.sh/uv/)
+- Blender
+- [blext](https://pypi.org/project/blext/)
+
+
+## Starting
+
+
+```python
+# uvx blext show spec
+BLExtSpec(
+    path_proj_root=PosixPath('/Users/zcpowers/Documents/Projects/MolecularNodes'),
+    req_python_version='~=3.11',
+    is_universal_blext=True,
+    bl_platform_pypa_tags={},
+    init_settings_filename='init_settings.toml',
+    manifest_filename='blender_manifest.toml',
+    init_schema_version='0.1.0',
+    use_path_local=False,
+    use_log_file=False,
+    log_file_path=PosixPath('addon.log'),
+    log_file_level=<StrLogLevel.Debug: 'DEBUG'>,
+    use_log_console=True,
+    log_console_level=<StrLogLevel.Debug: 'DEBUG'>,
+    manifest_schema_version='1.0.0',
+    id='molecularnodes',
+    name='BLExt Simple Example',
+    version='4.4.0',
+    tagline='Toolbox for molecular animations with Blender and Geometry Nodes.',
+    maintainer='Brady Johnston <brady.johnston@me.com>',
+    type='add-on',
+    blender_version_min='4.2.0',
+    blender_version_max='4.3.10',
+    permissions={},
+    tags=('Development',),
+    license=('SPDX:GLP',),
+    copyright=('2024 blext Contributors',),
+    website=HttpUrl('https://bradyajohnston.github.io/MolecularNodes'),
+    bl_platforms=frozenset(),
+    wheels=()
+)
+```
+
+
+
+```sh
+# you need to use the head
+# see https://codeberg.org/so-rose/blext/issues/57
+uvx --from git+https://codeberg.org/so-rose/blext blext build
+uv add  git+https://codeberg.org/so-rose/blext
+
+
+# show
+uvx blext show manifest
+uvx blext show init_settings
+uvx blext show spec   # see above
+uvx blext show path_blender # works locally
+
+# build
+uvx blext build
+
+# dev
+uvx blext dev
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,6 @@ dependencies = [
     "biotite>=1.1",
     "mrcfile",
     "starfile",
-    "blext",
 ]
 maintainers = [{ name = "Brady Johnston", email = "brady.johnston@me.com" }]
 requires-python = "~=3.11"
@@ -24,7 +23,7 @@ Documentation = "https://bradyajohnston.github.io/MolecularNodes"
 [project.optional-dependencies]
 bpy = ["bpy>=4.4"]
 jupyter = ["jupyter", "IPython"]
-dev = ["pytest", "pytest-cov", "syrupy", "scipy", "fake-bpy-module", "IPython"]
+dev = ["pytest", "pytest-cov", "syrupy", "scipy", "fake-bpy-module", "IPython", "blext"]
 test = ["pytest", "pytest-cov", "syrupy", "scipy", "IPython", "pytest-xdist"]
 docs = [
     "quartodoc",
@@ -34,10 +33,26 @@ docs = [
     "IPython",
     "jupyter-cache",
 ]
+blender4_3 = [
+    "autopep8==2.3.1", # # ⭳⭳⭳ MANAGED BY BLEXT ⭳⭳⭳
+    "certifi==2021.10.8",
+    "charset_normalizer==2.0.10",
+    "cython==0.29.30",
+    "idna==3.3",
+    "numpy==1.24.3",
+    "pip==24.0",
+    "pycodestyle==2.12.1",
+    "requests==2.27.1",
+    "setuptools==63.2.0",
+    "urllib3==1.26.8",
+    "zstandard==0.16.0", # # ⭱⭱⭱ MANAGED BY BLEXT ⭱⭱⭱
+]
+
 
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = ["setuptools>=61.0"]
+
 
 [tool.setuptools]
 include-package-data = true
@@ -56,7 +71,6 @@ ignore = [
     "E501",
     # Due to constants and class placeholders defined in functions
     "N806",
-
     # 'molecularnodes' when used as a Blender extension may not actually be named
     # 'molecularnodes' (potentially user_default.molecularnodes for example) so we have to
     # use relative imports:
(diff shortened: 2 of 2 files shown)


so-rose:
A few tips that might make life easier (I'll make sure it's documented in `blext` too):
- If you `uv add --dev https://codeberg.org/so-rose/blext.git`, then you can always `uv run blext` from within the project, without having to get people to use a specific version via `uvx`. `--dev` dependencies aren't included in extensions, for reference, so you can also add ex. `ruff`, `mypy`, etc. without messing with the wheels that ship.
- If you do a global install with `uv tool install https://codeberg.org/so-rose/blext.git`, then whenever you run just `blext` (without `uv run`) within a project, then `blext` will automatically proxy itself to the version inside of the project. This way you can always just run `blext build`, for instance, and it'll make sure to use the `blext` version that the project itself is using.
    - So basically, it's like automated `uv run`.

Anyone reading, please keep in mind `blext` is very alpha. But I think it's very cool there's a MR already to try it!

BradyAJohnston:
I'm really liking the CLI, all seems to run pretty well. 

For our use case we would always want to generate separate `.zip` files rather than one monolithic one, the same as running:

```
blender --command extension build --split-platforms
```

Secondly, I found it strange that the default build location is in a hidden folder. This doesn't seem practical for people then wanting to manually distribute the `.zip` file.

so-rose:
Glad you like it! And good feedback.

I've been going back and forth on whether it's best to do monolithic zips or separate zips. I think your feedback is the tipping point; I agree, and in practice, one would always be building platform-specific `.zip`s (unless there are no special wheels, of course). I'll need to do a bit of thinking about the exact semantics of this.

I'm mainly hiding the zip to avoid polluting the git repo root (I fear people checking in the `zip`s to `git`), but also because the filename is kinda long. But considering one can always `--output` it wherever, it's a good point that it should just be generated "in one's face" so to speak. I do that anyway for script extensions!

BradyAJohnston:
I think particularly because the Blender Extensions platform expects uploads of split `.zip` files that it should be the default, glad we agree!

zachcp:
Thanks for your comments @so-rose  and @BradyAJohnston . I'm trying to incorporate your suggestions as best I can. 

The first blext-based CI build issue i see is this one:

```sh
https://github.com/BradyAJohnston/MolecularNodes/actions/runs/14335185422/job/40180310263#step:4:168

│   1119 │   │   │   │   raise                                                 │
╰──────────────────────────────────────────────────────────────────────────────╯
FileNotFoundError: [Errno 2] No such file or directory: 
'/Users/runner/Library/Caches/blext/project_cache'
```

I'm also running into some issues in the test setup where bpy is not being seen. Somehow or another the testing environment is not being setup correctly even though there is a bpy install line.  

```sh
https://github.com/BradyAJohnston/MolecularNodes/actions/runs/14335185422/job/40180310263
Run source .venv/bin/activate
/Users/runner/work/MolecularNodes/MolecularNodes/.venv/bin/python: No module named pytest
```

zachcp:
Thank you  @so-rose 

```sh
https://github.com/BradyAJohnston/MolecularNodes/actions/runs/14336408856

 Downloaded pydantic-core
    Updated https://codeberg.org/so-rose/blext.git (541b525ba29aa4c19169aeeda7efb1542bffbf7e)
   Building blext @ git+https://codeberg.org/so-rose/blext.git@541b525ba29aa4c19169aeeda7efb1542bffbf7e
 Downloaded black
 Downloaded sqlalchemy
 Downloaded pygments
```

I am trying to use main but I keep hitting what seems to be a cache - e.g. `@541b525` is a few commits back. Not sure if you know the fix

zachcp:
Yes I used your suggestion. Ill try the upgrade route.

```toml
[tool.uv.sources]
blext = { git = "https://codeberg.org/so-rose/blext.git" }
```

zachcp:
@so-rose  - I would guess that my pyproject.toml tool.blext passes the lint but maybe is off somehow. Any chance you could take a quick look?   ( And thank you for your efforts here!  )

zachcp:
Ok! All tests and builds pass now!

@so-rose  FYI -when I `uv add blext` it added blext as a dependency. But when I move it to a dev dependency I don't get the build issue. Problem solved!


@BradyAJohnston  there are at least 2 important issues left related to the build step. 


1. I had to temporarily override the toml version tag. Theres got to be a better way to get that....
2.  the builds are currently only generating a single file.... eg. [here](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/14337603495)

```toml
      run: |
        uv sync --all-groups
        uv run blext build --platform macos-arm64
        uv run blext build --platform linux-x64
        uv run blext build --platform windows-x64
        mkdir artifact
        # mv ${{ env.ADDON_NAME }}-${{ steps.read_manifest.outputs.value }}-*.zip ./artifact/.
        mv build/*.zip ./artifact/.
```


image

zachcp:
Although:  the artifact does now produce the 3 files as desired. 

image

zachcp:
Note: this also relates to #802

zachcp:
@so-rose  - your `main` works great for generating per-platform ZIP files.

so-rose:
I'm glad to hear it! Took a moment to get the splitting logic right, but it seems like the right decision. Actually, the hairiest part was representing the Blender version range, but I'm hoping the filename format can stabilize now.

FYI if you need to change the output names, `--output` now supports a few macros ex. `--output build/%bl_version/%ext_id__%platform.zip`. See `blext build --help` for details.

But yes, I can see that I should probably check the `uv add blext` case without `--dev`... For now I'm mainly focused on fixing issues for `bpy_jupyter`, and otherwise documenting everything nicely, but it's good to write these smaller issues down so I can make sure to triage before releasing.

BradyAJohnston:
Just tried it out again and it's working super smoothly, big fan of the informative CLI.

I wouldn't want to use it yet though until the built `.zip` file is `<=` the current built size. All 3 are ~40 mb larger than the current build system.

so-rose:
@BradyAJohnston Thank you so much!

And I get ya - any chance you could help me point out which wheels are getting added incorrectly? If wheels are indeed the culprit. Whatever the reason, binary sizes are very important to me as well (I have an addon I'm wanting to port that starts with jaxlib, so that's already heavy as heck; anything to bring that down would be nice).

BradyAJohnston:
The original list of packages I had for purging from the build I just defined manually:

```py
    # Packages to remove
    packages_to_remove = {
        "pyarrow",
        "certifi",
        "charset_normalizer",
        "idna",
        "numpy",
        "requests",
        "urllib3",
    }
```

Looking at the builds from `blext`, the only one that is there is `pyarrow` (30 mb). This isn't a blender thing but came from one of the dependencies though, but I figured out initially that I could exclude it without anything breaking. Being able to provide a list of packages to exclude from the final build would be useful. Lots of potential to shoot oneself in the foot, but so many of these packages I am using only a small subset of their features anyway that you can get away with dropping some of their dependencies.

I think it's doing a pretty spot-on job for resolving dependencies otherwise.

so-rose:
@BradyAJohnston Glad to hear it's otherwise working well! An interesting conundrum indeed.

From what I can tell in the `uv.lock`, `starfile` is pulling in `pyarrow` as a hard dependency. Which raises some deeper questions.

## The Deeper Questions
I find myself **strongly disagreeing** that providing a package exclusion mechanism is a good idea. It's not just that it's a footgun, it's potentially a foot bazooka.

Who's to say that one day:
- `starfile` one day releases a patch-release that relies on `pyarrow` being always available, as `starfile` explicitly requires anyway, causing the extension to break the next time someone bumps the lockfile?
- Someone imports the wrong thing while scripting, based on the functionality `starfile` promises, and now everything breaks?
 
Tl;dr: By overriding `starfile`'s claim that `pyarrow` is _always_ required, every patch release is potentially a landmine. That's a fragile way to live, and I don't think that's something I'd like to promote with `blext`.

## Solutions
So **what to do in practice**? Clearly, `starfile` does not (in fact) _always actually_ require `pyarrow`, even though it is a "hard" dependency.

So, to put it bluntly, we've just bound bug in `starfile`. Those `30MB` may not usually matter, so upstream may consider it a small bug - but in this case where `30MB` matters a lot, it's a big bug to us! So what to do:

- Convince `starfile` to make `pyarrow` (and probably also `polars`, which uses `pyarrow`) an extra, which would also force combing through the code and adding `try/except ImportError` in places where `pyarrow` is about to be included.
- If upstream disagrees, that's their right. Then one runs a soft-fork of `starfile` that does the above, and which otherwise just tracks the `main` of starfile, and publish that on PyPi with a well communicated message.
- Of course, one can also "just" post-process the zipfile to remove the `pyarrow` wheel and remove the corresponding `blender_manifest.toml` entry. If this seems iffy, _that's because it absolutely is incredibly iffy_ - there is sometimes a virtue in hacks like this feeling exactly as iffy as they are.
- Of course, one could just eat the `30MB` bump and treat the original situation as a bug :wink: But that's not for me to decide.

_Note: The exclusion of Blender's vendored `site-packages` is a very careful thing. We presume that `uv` guarantees the compatibility of all vendored `site-packages` of all supported Blender versions with all manually added dependencies. Which is a tall ask, but the specificity is what makes the exclusion a good idea._

BradyAJohnston:
Thanks for the extra write-up, you've certainly convinced me that what the current approach was is a pretty bad idea to continue with.

It's unfortunate that we have be so conscious of this for distributing the extensions from the official platform. Would be nice to just have some kind of dependency management to solve these issues.

so-rose:
It's my pleasure. The Python ecosystem is kind of a mess.

I too have enormous criticisms of Blender's mechanisms for dealing with pydeps. The problems are very deep - see [this unfortunate comment](https://projects.blender.org/blender/blender/src/commit/5a2a6da0a27f313c67cd6e800a1da6c32b55deaf/scripts/modules/_bpy_internal/extensions/wheel_manager.py#L580), for instance. _"Fairly unlikely" is quite a controversial statement_...

But for now, my hope is that `blext` can at least automate these gymnastics. Once some tooling is in place, and the problems are more well described, maybe upstream would be more receptive to a well though out proposal for non-vendored dependencies (maybe even down to an open letter with a round of signatures among extension makers, if it comes down to it?)

BradyAJohnston:
I think we keep the PR open, and plan to use `blext` going forward.

Once `v0.4` is out I'll go through and double check everything in PR for it.

so-rose:
Awesome! And inspiring for me. Some life stuff and a vacation got in the way, but I'm back now.

I'll make a comment when things feel stable enough (perhaps I'll do a pre-release) `blext`-side, that way you can get any last opinions in before I put `0.4.0` into the world. _Even though it's not 1.0 yet, I'm trying to be very careful about predictable backwards compatibility breaks._

zachcp:
There hasn't been much motion on this PR or on a blext release.  I am unlikely to see this through so I am closing.  When a stable blext is released it can be incorporated.

so-rose:
Sorry for the delay with `blext` - university exams required my attention. The good news is I will soon have a BSc degree.

I'll write here again when there's more on a stable release of `blext`; I'd be happy to be a part of making sure it fits all needs at that time.

BradyAJohnston:
Looking forward to it @so-rose :)

PR #843, opened 2025-04-28
Testing for Used-Defined Materials

zachcp:
- test material passing to 'add_style` 
- test for passing custom material to add_style


zachcp:
@BradyAJohnston I was adding these tests because something was wrong in my environment. But it turns out the code works great. Might be useful to have them there. Please let me know if you want any updates.

BradyAJohnston:
LGTM. Always happy for more tests!

PR #844, opened 2025-04-29
Update tutorial documentation

BradyAJohnston:
- **rename doc files**
- **barebones simulation docs**


zachcp:
Note: I would like to move most of the efforts I put into prettymol into this repo.  Currently I am refactoring a bit and will need to make some changes to adapt to the selection system. I raise this issue here as I would be happy to take on a few simple viz/animation examples.

PR #847, opened 2025-04-29
turn on compositing

zachcp:
- ported code from prettymol
- when using scene this will turn on compositing by default. needed for nice emissive rendering


BradyAJohnston:
Enabling and customising compositing is something we should be working towards, but it needs to be a more general approach than this currently (to support outlines, depth effects etc) rather than just an option for glow.

I'm not sure exactly how to approach it. What I have been thinking is that we create a node tree setup with a bunch of compositing effects ready to go (which covers the range of potential compositing we directly support), then we can enable / disable them and change their values as needed via scripting / UI.

zachcp:
Yep. This sets up the node tree. What are the other compositing options that are not supported - basically this just activates compositing and creates the nodes, and does that on canvas creation.

zachcp:
Based on Brady's comments above, I am closing this.

PR #849, opened 2025-04-29
Add a materialcreator.

zachcp:
This is a WIP PR that adds a MaterialCreator class that can generate Materials. The problem I am trying to address is to be able to programatically generate and explore material creation and make it very easy to use nice presets as well as to explore the full capability of Blender's material system.

This PR introduces 2 classes - Material, Material Creator - where the usage is as follows:

```
MaterialCreator --> Material. --> bpy.types.Material
```

```python
## Material Creator 

# preset materials
MaterialCreator().pearl()  # <- pearly material

# base starting material for altering
MaterialCreator().new()  # <- base material

# updates properties
MaterialCreator().velvet(). update_properties(base_color=[1,0.5,0.2], alpha = 0.5)

# create emissive green material that can be passed o MN
gfp_mat = MaterialCreator().green_glow().blenderize("GFP-Like")
mol = mn.Molecule.fetch("4ozs").add_style("ribbon", material= gfp_mat)

```

I am open to all sorts of modifications on how this should work and where the code should sit etc. But I think something like this would be useful in-package.


BradyAJohnston:
I'm certainly in favor of an approach like this, but I'm a little unsure about the idea of creating the material and having to call `.blenderize()` on it to turn it into an actual material.

Additionally why are we having to call the `.update_properties()` method instead of just updating properties on the object like this? 

```py
mat = mn.material.Velvet(alpha = 0.5)
mat.base_color = (0.1, 1.0, 0.3)
```

I'm not against this design but I'm a little confused by it's approach.

zachcp:
Okay. cool call. my whole precious approaches tried to be immutable as possible - hence the immutable dataclasses. I can definitely refactor ( or create another PR ) that will adhere to your style which I agree is a better fit with the codebase.

BradyAJohnston:
Happy with either a new PR or this one.

zachcp:
## Update 

- Single `create_material` function that takes a set of args and returns a bpy.types.Material tree.
- Single `MN_Materials` class that defines a set of presets as data.

I think this is much more consistent with the codebase although to be truly the same we would need to expose the material sockets as you do above.

One additional note is that I think you could probably also encode the existing defaults in a similar way. This would let you load them using function calls and remove the need to save them in the `blend file`

## Repro Example

```python
import bpy
import molecularnodes as mn
from molecularnodes.nodes.material import MN_Materials
from molecularnodes.scene.base import Canvas

can = Canvas()
mn.register()
can.scene_reset()

try:
    cube_object = bpy.data.objects.get("Cube")
    bpy.data.objects.remove(cube_object, do_unlink=True)
    print("Default Cube deleted successfully!")
except:
    pass
    
pearl = mn.material.MN_Materials.pearl()
structure = mn.Molecule.fetch("1EMA", centre="centroid")
structure.add_style("ribbon", material=pearl)
can.frame_object(structure)
can.snapshot()
```

image

```python
# note this one looks flat as compositing is not enabled. 
# see #847  for a possible solution.

gfp = mn.material.MN_Materials.green_glow()
structure = mn.Molecule.fetch("1EMA", centre="centroid")
structure.add_style("ribbon", material=gfp)
can.frame_object(structure)
can.snapshot()
```
image

zachcp:
( ping )

BradyAJohnston:
I think I am happy enough with this approach currently. 

The material creator is however limited to just the `Principled BSDF` and the potential input values for building a new material. More complex materials like the default ones that are included with MN are more complex node trees. It runs into the same problem again that there isn't a nice way to build node trees via code.

BradyAJohnston:
Unsure if it's something we try to do in this PR, but also we'll need to support the `MN Color Input` node, to access colors from geometry nodes: 
image

BradyAJohnston:
The `.qmd` file will execute with building the docs if you put the make it a python executable block instead of just a python code block. Then it can become part of the docs & will be executed on render to also serve as a test.

zachcp:
@BradyAJohnston  I think adding complicated nodes is out of scope for this PR.  This opens up a new way to define nodes but does not preclude using nodes from the MN or other  blend files, or for creating more complacated node treed via something like GeometryScript.  I recommend accepting.

molecularnodes/nodes/material.py
```diff
@@ -187,3 +189,249 @@ def __init__(self, **kwargs):
 class TransparentOutline(MaterialConstructor):
     def __init__(self, **kwargs):
         super().__init__("MN Transparent Outline", **kwargs)
+
+
+class MN_Materials:
```
BradyAJohnston, line 194: This shouldn't need `MN_` prefix as it is coming from the MN package.

molecularnodes/nodes/material.py
```diff
@@ -187,3 +189,249 @@ def __init__(self, **kwargs):
...
+    - FlatOutlineq
+    - Squishy
+    - TransparentOutline
+
+    """
+
+    @staticmethod
+    def glass() -> bpy.types.Material:
+        return create_material(
+            name="MN_Glass",  # Add a descriptive name
```
BradyAJohnston, line 210: For the individual material names, overall I have been thinking of removing the `MN ` prefix from the other materials, which I have done for all of the user-facing nodes inside of Geometry Nodes.

For now I think we stick with the prefix, but I think it should use space instead of underscore. Could also be something that is prepended by the `create_material()` function rather than having to define it in the actual name argument.

tests/test_select.py
```diff
@@ -40,3 +41,17 @@ def test_select_multiple_residues(selection):
...
     assert (bob.named_attribute("res_id", evaluate=True) == selection[1]).all()
+
+
+def test_gfp_selection():
+    structure = mn.Molecule.fetch("1EMA")
+
+    # select AA and create a named selection
+    structure.select.is_amino_acid().store_selection("Protein")
+    structure.add_style("cartoon", selection="Protein")
+    assert structure.bob.named_attribute("Protein").sum() == 1771
```
BradyAJohnston, line 52: you can just call `structure.named_attribute(` rather than explicitly having to do `bob` first, as `bob` is actually a mix-in for the `MolecularEntity`.

docs/animations/GFP_MN.qmd
```diff
@@ -0,0 +1,106 @@
+## GFP Code From Prettymol Ported to MN
+
+Note: WIP. Still some issues to work out
+
+
+```python
```
BradyAJohnston, line 6: What is happening with this document? Is it intended to be part of the documentation?

BradyAJohnston (changes requested):
I think I am happy with the overall design currently. Some nomenclature changes I think are required first, and something needs to be done with the `.qmd` file. 

If you want it to become part of the documentation then it needs some more write-up.

molecularnodes/nodes/material.py
zachcp: Updated .

molecularnodes/nodes/material.py
zachcp: Ok. Converted to space.s

tests/test_select.py
zachcp: Addressed.

docs/animations/GFP_MN.qmd
zachcp: removed the QMD it was useful for testing but I wouldn't render an entire animation in the docs.  If you want, you can assign me a separate issue to update some of the scripting docs.

BradyAJohnston (approved):
I'll double check again in the morning but I think this LGTM otherwise.

Issue #850, opened 2025-04-29
Spheres Are Initialized as Points Not Mesh


zachcp:
Using blender/MN 4.4.1 when I load spheres I get points. But I think the mesh option is a better default.

```python
# reproducible example
import molecularnodes as mn

structure = mn.Molecule.fetch("1EMA")
# select AA
structure.select.is_amino_acid().store_selection("Protein")
structure.add_style("cartoon", selection="Protein")
assert structure.bob.named_attribute("Protein").sum() == 1771

# select Chromophore
structure.select.res_name("CRO").store_selection("GFP")
structure.add_style("sphere", selection="GFP")
assert structure.bob.named_attribute("GFP").sum() == 22

```

## Defaults to Point 
![Image](https://github.com/user-attachments/assets/48c0d435-072d-4716-a82e-23ed87e12a7a)

## Using mesh
![Image](https://github.com/user-attachments/assets/3ff88930-d9dd-491b-b1d7-0cc94da2ff10)

BradyAJohnston:
For larger structures, importing as mesh will cause a hard-crash of Blender because the memory limit will be reached quickly. Point is the default because it is the fastest and the safest, so I don't think changing the default is a good idea.

What I would like to be possible though is to initialise the style with the mesh already selected as default if requested. 

My thought process would be to have the styles as classes similar to the default materials, so you can initialise like this:

```py
mol.add_style(mn.StyleSpheres(geometry="mesh"), selection = "Protein")
```

zachcp:
When I re-implement the materials as discussed on #849 , I'll take a stab at this as well.

zachcp:
Note this is fixed in 4.4.3:


```python
# reproducible example
import molecularnodes as mn
from molecularnodes.nodes.styles import  StyleSpheres


spheres = StyleSpheres(geometry="Mesh").   # <---- override default here
structure = mn.Molecule.fetch("1EMA")


# select 
structure.select.is_amino_acid().store_selection("Protein")
structure.add_style("cartoon", selection="Protein")

# select Chromophore
structure.select.res_name("CRO").store_selection("GFP")
structure.add_style(spheres, selection="GFP")

```

PR #851, opened 2025-04-30
Add Configurable Styles Classes

zachcp:
This PR adds support for modifying styles when adding styles to MN. Some initial discussion in #850 

```python
%load_ext autoreload
%autoreload 2


import bpy
import molecularnodes as mn
from molecularnodes.scene.base import Canvas
from molecularnodes.nodes.styles import StyleBallandStick, StyleRibbon, StyleSticks


can = Canvas()
mn.register()
can.scene_reset()

try:
    cube_object = bpy.data.objects.get("Cube")
    bpy.data.objects.remove(cube_object, do_unlink=True)
    print("Default Cube deleted successfully!")
except:
    pass

structure = mn.Molecule.fetch("1EMA", centre="centroid")
# structure.add_style("spheres")
structure.add_style(StyleRibbon())
can.frame_object(structure)
can.snapshot()
```
image

```python
# same as above except
structure = mn.Molecule.fetch("1EMA", centre="centroid")
structure.add_style(StyleRibbon(backbone_radius=3, quality=5))
```

image

### DIFF
--- a/molecularnodes/nodes/styles.py
+++ b/molecularnodes/nodes/styles.py
@@ -40,39 +40,27 @@ class StyleBallandStick(StyleBase):
     # fmt: off
     portdata: PortDataList = [
         { "name": "quality", "blendername": "Quality", "type": int, "default": 2 },
-        { "name": "as_mesh", "blendername": "As Mesh", "type": bool, "default": True },
-        { "name": "sphere_radii", "blendername": "Sphere Radii", "type": float, "default": 0.3 },
+        { "name": "sphere_radii", "blendername": "Sphere Radii", "type": float, "default": 0.30000001192092896 },
         { "name": "bond_split", "blendername": "Bond Split", "type": bool, "default": False },
-        { "name": "bond_find", "blendername": "Bond Find", "type": bool, "default": True },
-        { "name": "bond_radius", "blendername": "Bond Radius", "type": float, "default": 0.3 },
+        { "name": "bond_find", "blendername": "Bond Find", "type": bool, "default": False },
+        { "name": "bond_radius", "blendername": "Bond Radius", "type": float, "default": 0.30000001192092896 },
         { "name": "color_blur", "blendername": "Color Blur", "type": bool, "default": False },
         { "name": "shade_smooth", "blendername": "Shade Smooth", "type": bool, "default": True }
     ]
-
-    # Quality 2
-    # Sphere Geometry Instance
-    # Sphere Radii 0.30000001192092896
-    # Bond Split False
-    # Bond Find False
-    # Bond Radius 0.30000001192092896
-    # Color Blur False
-    # Shade Smooth True
     style="ball_and_stick"
     # fmt: on
 
     def __init__(
         self,
         quality: int = 2,
-        as_mesh: bool = True,
         sphere_radii: float = 0.3,
         bond_split: bool = False,
-        bond_find: bool = True,
+        bond_find: bool = False,
         bond_radius: float = 0.3,
         color_blur: bool = False,
         shade_smooth: bool = True,
     ):
         self.quality = quality
-        self.as_mesh = as_mesh
         self.sphere_radii = sphere_radii
         self.bond_split = bond_split
         self.bond_find = bond_find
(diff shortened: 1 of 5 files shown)


BradyAJohnston:
Thanks for having a crack at this. I think it mostly does what I was thinking. 

I don't know what the best approach would be, but I'm conscious of having "multiple sources of truth". Ideally I think we would generate these classes, or at least the `.py` that defines the classes from the data within the `.blend` file, so that if something is updated for the styles then these definitions stay in sync. Doesn't have to be this way, open to what you think about best approaches.

Would be more ideal I know to not have styles defined in a `.blend` but that's not really an option at the moment.

zachcp:
A couple of things come to mind:

1.Lightest weight is just a test suite that tests for deviations both ways. E.g. in fields or default values change

2. Codegen is possible but adds some build complexity. Let me think about it overnight.

BradyAJohnston:
The styles should remain static for patch releases, so they should only have to be updated every few months if at all. I think the test that you have setup should be sufficient. I'll have a more proper look over the PR in a few days when I've got some time, but I think this is basically the approach we want to take.

zachcp:
( ping )

BradyAJohnston (changes requested):
I think we should change the node definitions to use dataclasses for individual sockets rather than dictionaries.

zachcp:
- i've changed the names from `Port` to `Socket`
-  I am providing examples on how tab completion works. 
    - this method  allows you to complete on the function parameters. This is how you overload during creation as per your original design request.
    - because the return value is an instance of the MN nodeset from the MN file, you also get normal tab completion.
    

```python
import molecularnodes as mn
from molecularnodes.nodes.styles import StyleRibbon

# the function call parameters can be auto completed
StyleRibbon(backbone_radius=2)

# the return value is the same as a MN instance from the blend file
# so you can get all those completions as well
StyleRibbon().
```
Screenshot 2025-05-17 at 9 43 14 AM
Screenshot 2025-05-17 at 9 44 20 AM

zachcp:
Upon reflection I think the second image is slightly misleading.  The primary reason I needed to use dictionaries is to allow me to map between the python names which don't have spaces and the internal blender property names which do. The mapping and port data is not intended to be used / visible to end users. But I don't think the second completion style will update the values where the first one definitely will (e.g. on fn call we override the values in the MN base ). 

So I think this is the lightest touch implementation.  However,  would a  Data Class improve the behavior? You still woudl only be able to cleanly override/set the node properties at initialization. But maybe you could set the properties more easily before you do?  For a DataClass I would Imagine it would look like this:


```python

# the call chain for creating the style nodes

@dataclass
class StyleRibbon:
    backbone_radius: 1
    ...


# MN add_style
    def add_style(
        self,
        style: bpy.types.GeometryNodeTree | str = "spheres",
        color: str | None = "common",
        selection: "str | MoleculeSelector | None" = None,
        assembly: bool = False,
        material: bpy.types.Material | str | None = None,
    ):

# add style branch 
def add_style_branch(
    tree: bpy.types.GeometryNodeTree,
    style: str | bpy.types.GeometryNodeTree,
    color: str | None = None,
    selection: str | None = None,
    material: bpy.types.Material | str | None = None,
    frames: bpy.types.Collection | str | None = None,
) -> None:
```


So the implementation more or less depends on when you convert to the BPY Geometry Node. In the current PR, I am converting in the add_style call. Then we are passing a `GeometryNodeTree` to subsequent build calls. If you want to use DataClasses I think you have two options -> convert in the `add_style` call or a little further down in the add_style_branch call.  I think that could work. 



```python
ribbon_style =. StyleRibbon()
ribbon_style.backbone_radius = 3

mol = mn.fetch("1FAP").add_style(style= ribbon_style.init()).  // <-- realize here? not ergonomic

mol = mn.fetch("1FAP").add_style(style= ribbon_style) //   // <-- realize lower in the stack
```

If you like the last style I could: 

1. convert the classes to data classes.
2. allow both functions to accept style classes in their method call
3. handle the conversion to MN nodes deeper in the call stack.

BradyAJohnston:
The current approach of delaying converting to an actual node group until the `add_style_branch()` is the way to go. Outside of working inside the particular node tree we are working in we can't customise it otherwise so keeping as a custom python class until last minute when we are actually adding the node is likely the best approach.

zachcp:
Ok. great . Now its clear whats needed. I've updated accordingly.

molecularnodes/nodes/styles.py
```diff
@@ -0,0 +1,281 @@
...
+        for input in node_style.inputs:
+            if input.type != "GEOMETRY":
+                for arg in self.portdata:
+                    if input.name == arg.blendername:
+                        input.default_value = getattr(self, arg.name)
+
+
+class StyleBallandStick(StyleBase):
+    style = "ball_and_stick"
+    portdata: SocketInfo = [
```
BradyAJohnston, line 69: rename from port to socket

molecularnodes/nodes/styles.py
```diff
@@ -0,0 +1,281 @@
...
+        Socket(name="quality", blendername="Quality"),
+        Socket(name="radius", blendername="Radius"),
+        Socket(name="color_blur", blendername="Color Blur"),
+        Socket(name="shade_smooth", blendername="Shade Smooth"),
+    ]
+
+    def __init__(
+        self,
+        quality: int = 3,
+        radius: float = 0.20000000298023224,
```
BradyAJohnston, line 237: I know inside of blender it's technically the floating point version but in code we can just have it be 0.2

BradyAJohnston (changes requested):
Just some minor things but then I think we are good to go!

molecularnodes/nodes/styles.py
BradyAJohnston: this should be renamed very time it shows up

zachcp: done

molecularnodes/nodes/styles.py
zachcp: good catch

BradyAJohnston (approved):
Again will review and merge when in the morning after some sleep but LGTM!

zachcp:
Wonderful!  Think you can issue a new release?

Issue #861, opened 2025-05-14
oxDNA: resname parsing for T


elija-feigl:
**Describe the bug**
For oxDNA import: res_name T is not parsed properly and results in a -1 value.

**To Reproduce**
1. Loading any oxDNA trajectory file (containing T bases)
2. Click on Geometry Spreadsheet, with NamedAttributes
3. All bases that should be T are shown with res_name = -1
4. Colouring of those bases does not work based on the color select input.

**Expected behaviour**
As oxDNA appears to be using RNA types 40-43, the value should be 43.

**Error Codes**
-

**Screenshots**
<img width="702" alt="Image" src="https://github.com/user-attachments/assets/279fde5c-299b-4227-a33b-12bc72b49bcd" />

**Desktop (please complete the following information):**
 - OS: MacOS
 - Hardware:  M1 Mac
 - Blender Version: 4.3.2
 - MolecularNodes Version: 4.4.1

BradyAJohnston:
Thanks for finding this and providing the fix. If you have other oxDNA related problems & requests, please do let me know. I'm not really in the field so I'm not sure on what people are in most need of!

elija-feigl:
Thank you for the amazing tool and the great support!

PR #862, opened 2025-05-14
fix(oxDNA):  add resname T to RNA for oxDNA parsing

elija-feigl:
oxDNA uses the RNA data assets which to not include T (but U). this causes unknown resname for all T.

#861


BradyAJohnston:
Thanks for for the fix! LGTM

PR #864, opened 2025-05-17
Styles using Dataclasses

zachcp:
- Modified version of #851
- Separate PR so we can compare implementations.


zachcp:
Resuming previous PR after discussion.

PR #875, opened 2025-06-04
Main branch for GSoC 2025 Changes

BradyAJohnston:
example MDA to docs

### DIFF
--- a/docs/api/_sidebar.yml
+++ b/docs/api/_sidebar.yml
@@ -8,6 +8,9 @@ website:
     - api/materials.qmd
     - text: '---'
     - contents:
+      - api/mda/example_mda.qmd
+      section: MDAnalysis
+    - contents:
       - api/reference/index.qmd
       - contents:
         - api/reference/entities.Molecule.qmd
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -55,6 +55,10 @@ website:
     - tutorials/trajectories.qmd
     - tutorials/cryoem.qmd
     - tutorials/simulation.qmd
+    - text: "---"
+    - section: "MDAnalysis"
+      contents:
+      - "mda/example_mda.qmd"
 
   page-footer:
     left: "This site is built with ❤️ and [Quarto](https://quarto.org/)."
@@ -94,6 +98,7 @@ quartodoc:
       contents:
       - "{{ contents }}"
   # css: api/_styles-quartodoc.css
+
 
   sections:
   - title: Entity
--- a/docs/mda/example_mda.qmd
+++ b/docs/mda/example_mda.qmd
@@ -1,0 +1,26 @@
+---
+title: "Example Scripting"
+description: "An example of using MDAnalysis with MolecularNodes."
+author: Brady Johnston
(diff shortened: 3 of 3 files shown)


docs/api/_sidebar.yml
```diff
@@ -7,6 +7,9 @@ website:
     - api/styles.qmd
     - api/materials.qmd
     - text: '---'
+    - contents:
+      - api/mda/example_mda.qmd
```
PardhavMaradani: `example_mda.qmd` in only in the `docs/mda` folder and not in the  `api/mda` folder, so this will be an error?

BradyAJohnston: yeah sorry I moved it out to top level. Sidebar was being problematic. Everything can be moved around later without too much trouble.

BradyAJohnston:
I haven't looked back into previous PRs, but there is currently a rendering error that we are getting when rendering the docs (on the GitHub runner, which is a self-hosted linux PC with an RTX 3090). Unsure what might be causing it, nothing is showing up when I run locally on my M2 macbook pro.

```
Cell 4/5: ''...EGL Error (0x3009): EGL_BAD_MATCH: Arguments are inconsistent (for example, a valid context requires buffers not supplied by a valid surface).
EGL Error (0x3009): EGL_BAD_MATCH: Arguments are inconsistent (for example, a valid context requires buffers not supplied by a valid surface).
Done
```

https://github.com/BradyAJohnston/MolecularNodes/actions/runs/17161939095/job/48695498724

Render times seem to have also ballooned out, so might be erroring and switching to CPU rendering or something?

@PardhavMaradani

PardhavMaradani:
Hi Brady, here are some data points:

It seems like this is the first `build-docs` run where we are seeing this error. It shows up for the other qmd files as well and not just the newly added `rendering.qmd`. The `Rendering Docs` part took around 25 mins (normally it is around 4 mins).

The `build-docs` run when the last PR (#940) got merged into the `gsoc-2025` branch ran fine as can be seen [here](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/17161531912/job/48691789733).

Even the previous run on this branch (after PR #938 merge) ran fine as can be seen [here](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/17153869560/job/48666061562).

This seems to indicate that this is not a code issue (as there are no changes that are render related post the above PRs). A quick search seemed to  point at driver issues. Maybe this could be an environment issue - is there a way to re-run just this build again?

Update: The `build-docs` of the latest PR (#941) is also showing the same [issue](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/17172433360/job/48723840943?pr=941) and I suspect every new build will show the same till the local setup is reset?

BradyAJohnston:
OK I'll have a look at the linux machine, I might need to change something on it. Thanks for looking into it.

BradyAJohnston:
It seems the server had restarted and lost connection to it's GPU. I updated drivers and restarted and all seems to be working again.

PardhavMaradani:
Thanks, Brady

PR #877, opened 2025-06-04
Basic universe management APIs and GUI

PardhavMaradani:
* Base class for MDAnalysis visualization (`MDAVis`)
* APIs for Universe addition/listing/retrieval/deletion
* Subscriptable and iterable universes for easy data access
* GUI for universe listing and management (addition/deletion) and Trajectory info
* Example notebook with basic universe management calls
* Tests for all the new functionality

### DIFF
--- a/molecularnodes/mda/base.py
+++ b/molecularnodes/mda/base.py
@@ -1,0 +1,105 @@
+"""
+Base class for MDAnalysis visualization
+"""
+
+import bpy
+import MDAnalysis as mda
+from MDAnalysis.core.groups import AtomGroup
+from ..session import get_session
+from .universes import Universe, Universes
+from . import utils
+
+
+class MDAVis:
+    """
+    Base Class for MDAnalysis visualization
+
+    Attributes
+    ----------
+    universes: Universes
+        Subscriptable and iterable list of Universe objects
+    session: MNSession
+        MolecularNodes session object
+    """
+
+    def __new__(cls) -> None:
+        """This is a singleton class"""
+        # try to retrieve object from MNSession
+        try:
+            session = get_session()
+        except AttributeError:
+            from .. import register
+
+            register()
+            session = get_session()
+        if hasattr(session, "MDAVis"):
+            return session.MDAVis
+        instance = super().__new__(cls)
+        session.MDAVis = instance  # add instance to MNSession
+        instance._initialized = False
+        instance.session = session  # link to MNSession (for deletes)
+        return instance
+
+    def __init__(self) -> None:
+        if self._initialized:
+            return
+        self._initialized = True
+        self._universes = {}  # dict to hold Universe objects created
+        self.universes = Universes(self._universes)
+        utils.subscribe_to_active_object()  # subscribe to active object changes
+
+    def add_universe(
+        self,
+        universe: mda.Universe | AtomGroup,
+        style: str = "spheres",
+        name: str = None,
+    ) -> Universe:
+        """Add a Universe or AtomGroup to Blender"""
+        u = Universe(universe, style, name)
+        self._universes[u.key] = u  # add to dict
+        print("Added universe", u.name)
+        return u
+
+    def delete_universe(self, universe: Universe | str) -> None:
+        """Delete MDA universe from Blender"""
+        if isinstance(universe, str):
+            # lookup universe object based on name
+            u = self.get_universe(universe)
+            if u is None:
+                return
+            object = u.object
+            key = u.key
+        else:
+            object = universe.object
+            key = universe.object.mn.mda.universe_key
+        sprop = bpy.context.scene.mn.mda  # scene pointer property
+        index = sprop.universes.find(key)  # find universe property in collection
+        if index == -1:
+            print(universe, "not found")
+            return
+        sprop.universes.remove(index)  # remove from universes collection
+        sprop.active_index = len(sprop.universes) - 1
+        del self.session.entities[object.uuid]  # remove from MNSession
+        object_name = object.name  # object / universe name being deleted
+        bpy.data.objects.remove(object, do_unlink=True)  # remove Blender object
+        del self._universes[key]  # remove universe object from dict
+        print("Deleted universe", object_name)
+
+    def list_universes(self) -> None:
+        """List all universes"""
+        if len(self._universes) == 0:
+            print("No universes")
+            return
+        for u in self._universes.values():
+            print(u.name, u.trajectory)
+
+    def get_universe(self, name: str) -> Universe | None:
+        """Get universe based on name"""
+        if name not in bpy.data.objects:  # check name in blender objects
+            print(name, "universe not found")
+            return None
+        object = bpy.data.objects[name]
+        if not utils.is_mda_universe(object):  # check object is mda universe
+            print(name, "is not a mda universe")
+            return None
+        return self._universes[object.mn.mda.universe_key]
--- a/molecularnodes/mda/universes.py
+++ b/molecularnodes/mda/universes.py
@@ -1,0 +1,88 @@
+"""
+Universes
+"""
+
+import bpy
+import MDAnalysis as mda
+from MDAnalysis.core.groups import AtomGroup
+from ..entities.trajectory import Trajectory
+
+
+class Universe:
+    """
+    Base Universe class
+
+    Attributes
+    ----------
+    name: str
+        Name of the universe
+    object: bpy.types.Object
+        Corresponding Blender object
+
+    """
+
+    def __init__(
+        self,
+        universe: mda.Universe | AtomGroup,
+        style: str = "spheres",
+        name: str = None,
+    ) -> None:
+        """Initialize by adding the universe"""
+        sprop = bpy.context.scene.mn.mda  # scene pointer property
+        uprop = sprop.universes.add()  # add new universe property to collection
+        uprop.name = "u" + str(sprop.next_index)  # unique key name
+        sprop.next_index += 1
+        sprop.active_index = len(sprop.universes) - 1
+        object_name = name
+        if object_name is None:
+            object_name = uprop.name  # use unique name if no name passed
+        # create the trajectory
+        if isinstance(universe, AtomGroup):
+            self.trajectory = Trajectory(universe.universe)
+            # TODO: Add style only from the AtomGroup selection
+        else:
+            self.trajectory = Trajectory(universe)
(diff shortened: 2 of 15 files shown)


PardhavMaradani:
Hi Brady,

Couple of notes:
- I disabled the execution of the example `basic-universe-management.ipynb` for now because the example uses datafiles from `MDAnalysisTests` and it looked like it isn't in the dependency list
- The  `docs/api/_sidebar.yml` file seems to get generated on the docs build (`uv run quartodoc build`), but I checked in the current version after I verified the docs build

BradyAJohnston:
I'll have a more detailed look at this later this evening, but I'm happy to add `MDAnalysisTests` to one of the optional dependencies. The old tests that use current data should also be moved over to use trajectories from there as well probably, but something for another project.

yuxuanzhuang:
I will also take a look sometime this week.

Quick thought: there are a lot of `print`, should a logging mechanism be introduced to MN?

molecularnodes/mda/base.py
```diff
@@ -0,0 +1,105 @@
...
+        instance._initialized = False
+        instance.session = session  # link to MNSession (for deletes)
+        return instance
+
+    def __init__(self) -> None:
+        if self._initialized:
+            return
+        self._initialized = True
+        self._universes = {}  # dict to hold Universe objects created
+        self.universes = Universes(self._universes)
```
yuxuanzhuang: this is out of sync to the change of `_universes` right now; maybe set `universes` as a class property instead.

molecularnodes/mda/base.py
```diff
@@ -0,0 +1,105 @@
...
+        utils.subscribe_to_active_object()  # subscribe to active object changes
+
+    def add_universe(
```
yuxuanzhuang: If you add the same universe twice, you will get two separate MN universes but they are both linked back to the same mda universe? Is that something we want?

molecularnodes/mda/universes.py
```diff
@@ -0,0 +1,92 @@
...
+Universes
+"""
+
+import bpy
+import MDAnalysis as mda
+from MDAnalysis.core.groups import AtomGroup
+from ..entities.trajectory import Trajectory
+
+
+class Universe:
```
yuxuanzhuang: The naming is getting really confusing now...maybe we should make them more explicit on what they actually are.

molecularnodes/mda/base.py
PardhavMaradani: The `Universes` class is currently a wrapper to get subscriptable and iterable access (along with auto complete of universe names) of the added universes. Because it is a dict, it seems to show updated values during my testing and hence I didn't run into a problem. Could you please let me know how this can be changed if you still see an issue.

molecularnodes/mda/base.py
PardhavMaradani: From my testing, both these MN universes (created from the same mda universe) seemed to behave independently of the trajectory frame changes, etc and have separate entries in MNSession. I believe this is what we want. Please let me know if you guys think otherwise. We discussed about having a single MN universe object with support for multiple representations (styles), but not this particular case of trying a single mda universe to a single MN universe.

molecularnodes/mda/universes.py
PardhavMaradani: Yes, this could get confusing. The universes here are the ones in MN/Blender, which will have methods to add styles, annotations, etc later. Should we change them to something like `MNUniverse` or `BUniverse`? Do you guys have any recommendations on naming this?

PardhavMaradani:
> I will also take a look sometime this week.
> 
> Quick thought: there are a lot of `print`, should a logging mechanism be introduced to MN?

We could use `loguru` like in `ggmolvis`. Brady, adding a dependency seems to be more than adding to the `dependencies` list in `pyproject.toml` due to `wheels` that I don't fully understand yet. Could you please help.

molecularnodes/mda/universes.py
yuxuanzhuang: I’m leaning toward a name like `TrajectoryView` or `TrajectoryCanvas` since it conveys that this class controls how the underlying entity `Trajectory` is displayed and rendered.

On a related note, several of its potential methods are quite generic and could apply to other Entity types as well. It might make sense to extract those common methods into an abc, with each specific `EntityView` (e.g. for volumes) subclassing it and implementing only the entity-specific details. (ofc you don't need to implement all of them)

I’m also open to any other suggestions.

PardhavMaradani: Hi Yuxuan, do we want to rename just this class or others as well to not reference universes? For example, this is how we have it so far:

```
import molecularnodes as mn

vis = mn.MDAVis() # to create the visualization object
vu = vis.add_universe(...) # to add an MDAnalysis universe
# vu is the added `Universe` in blender which will have `add_styles` etc later
# vis.universes[... will also return the `Universe` object
vis.delete_universes(...) # to delete based on `Universe` object or name
vis.list_universes() # to list universes
```

Are you suggesting we rename the `add_universe` and other methods along with the `universes` class property in the main vis class? I chose this naming because it seemed more natural to me from the user perspective (ie., when we add an mda universe using the `add_universe` api call, I would expect the added universe object returned to further configure it etc), but I am open to renaming all these classes as you guys see fit. @BradyAJohnston could you share your thoughts as well.

PardhavMaradani:
I refactored the universe classes and also renamed them as `UniverseManager` and `UniverseView` to be more explicit of what they do. I am open to renaming them again if needed.

molecularnodes/mda/base.py
PardhavMaradani: This should not be an issue anymore in the refactored code

molecularnodes/mda/universes.py
PardhavMaradani: Could you guys please review the refactored and renamed code and let me know if this still is an issue? I will leave this unresolved for now.

BradyAJohnston:
I know you posted about this initially in the discussion thread before starting, but I have some concerns that it seems like we maybe duplicating work and creating multiple classes for basically the same purpose. 

I can't currently see a strong reason why we need a second manager of universes when the `MNSession` already exists. If there is lacking functionality, should it not just be added directly to that? The session should already be a record of what universes have been loaded into Blender, it seems bad practice to have an additional location where we are tracking this information and potentially getting out of sync. Additional categories for annotations and the like could just be added to the session.

Similarly with the `UniverseView` class. I think I would want some stronger justification for why this shouldn't be done with the existing `Trajectory` class. The point has already been raised about naming and potential confusion with so many things called universe / trajectory. Trying to keep the number of different names and classes as low as possible I think should be a priority. I'm not against the idea entirely, but I'm not currently seeing things that couldn't be done by improving the existing classes.

BradyAJohnston:
In the past few months I started to introduce the system for adding styles to the `Molecule` class, but haven't yet ported it to the `Trajectory` class (I was leaving it for this GSoC project for potential design changes etc) (PR: https://github.com/BradyAJohnston/MolecularNodes/pull/779) : https://bradyajohnston.github.io/MolecularNodes/api/styles.html

From my perspective it would just be a matter of making this style management available to the `Trajectory` as well, then exposing some of that via the GUI. Ideally everything (universes, molecules, ensembles etc) all share essentially the same style & node tree management code so that improving it for any of them improves it for all of them.

PardhavMaradani:
I thought we were starting out with the objective to provide a clear and carefully crafted API interface along with an associated GUI for MDAnalysis users who would need better visualization capabilities. But if that is not the case and we're leaving that to other visualization tools like `ggmolvis` etc we could definitely expose `MNSession` and the `Trajectory` entity directly. To me, it seems a bit awkward to expose these directly but I agree that `MNSession` is already holding the entity details and `Trajectory` is what is already responsible for the view. I don't quite see the relevance of the `MDAnalysis` named GUI tab in the n-panel in this model.

Given that there is already a way to add trajectories (by instantiating a `Trajectory` object followed by a `create_object`), this isn't really a blocker for me. If you want to have APIs like `add_trajectory`, `get_trajectory` etc to `MNSession`, could you please help define clearly what these would look like (along with if any GUI support for them is needed and how they'd exactly look like). Given that this is how all users would interact with this, it would greatly help if you guys can outline this based on your experience and looking at it from the longer run. I could then help with implementing those as you guys see fit. For now, I could start working on the more independent aspects like the style branch support, annotations, etc.

BradyAJohnston:
> I thought we were starting out with the objective to provide a clear and carefully crafted API interface along with an associated GUI for MDAnalysis users who would need better visualization capabilities. But if that is not the case and we're leaving that to other visualization tools like ggmolvis etc we could definitely expose MNSession and the Trajectory entity directly. To me, it seems a bit awkward to expose these directly but I agree that MNSession is already holding the entity details and Trajectory is what is already responsible for the view. I don't quite see the relevance of the MDAnalysis named GUI tab in the n-panel in this model.

I'm not sure what is meant here. The goal is still the same, to improve the API. I don't see that duplicating code & responsibilities across the code base as doing that. I'm not understanding how improving and extending the existing `Trajectory` would work against these goals either? Improving `Trajectory` with the proposed abilities would result in the same but without introducing a new class and potentially & duplicating code and responsibility. 

I'm not 100% opposed to the new class and the changes in this current PR, but I currently don't understand how implementing them on the existing `Trajectory` is limiting in any way?

PardhavMaradani:
As I see it, `Trajectory` and `MNSession` already have a lot of internal details. I'm not sure if we want to expose all of that to users outside - as opposed to a simpler and cleaner interface (as a top down approach). We will have things like `add_style` etc within `Trajectory` (as we are discussing in #879 ). Any external interface provided will still use these internally, so I'm not sure why this would be considered duplicating code or effort. Maybe we could get all the specifics done and then get back to how these could be exposed. I would again defer to you guys to help define what this needs to be and I can fill in accordingly.

BradyAJohnston:
I think go forward with implementing around the existing `Trajectory`. Other particulars that don't need to be exposed we could hide. There is certainly lots overall that could be improved & cleaned up on the `Trajectory` class.

PardhavMaradani:
I have re-opened the discussion #870 to document the changes. We can close this PR. I will open a new PR once I get a go ahead with what is outlined in that discussion

PardhavMaradani:
Closing this as per the [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/879#discussioncomment-13467304) in discussion #870 . A new PR #882 should address the new requirements.

Issue #787, opened 2025-03-20
Support MolViewSpec import


BradyAJohnston:
With improvements to defining and creating node trees progrommatically in #779 & #775 , we can aim to work towards supporting `MolViewSpec`: https://molstar.org/mol-view-spec-docs/

This could be either through drag & drop of `.mvsj` into the viewport or via a script. Similarly specification of relevant values in a `.yaml` for batch processing.

Support for labels is likely further off, but molecular import & customisation is close.

zachcp:
I am interested in implementing this and would like to support MVSJ directly.  With th e New scripting additions of MN4.4 this should now be possible. The main aspect that I would need some assistance on are related to globals (translation/rotation of camera and objects. Maybe you have a way you were thinking of implementing these?) and label (e.g. text etc. Maybe its possible to build on Jan's work with typst/text? )

BradyAJohnston:
I haven't put any thought into other approaches for transformations, but Blender does already have a pretty robust setups for this it would just be a matter of wrapping them nicely.

I think for MVP we don't think about text yet, and just focus on molecular import, styling & coloring to match up with MVSJ.

Text and labels is something that is relevant to more than just MVSJ, which needs a bit more thinking and design work before we start implementing.

BradyAJohnston:
Also yes if you'd like to have a crack at it, please do!

zachcp:
Hi @BradyAJohnston , 

MolViewSpec often uses fully fleshed out URLs instead of codes.  If I wanted to extent MN to accept URLs for download where would you recommend adding them?  The most natural place to me would be: 

- `mn.download.StructureDownloader`
- the CACHE should probably be longer than only the code - maybe the  full URL string? 

I'm open to better ideas.

BradyAJohnston:
If we are just caching the structure, then the structure is no different to other structures downloaded via other methods so should just be cached the same with code & filetype.

For handling URLS, are you meaning just for downloading or for parsing as well?

zachcp:
Just the download itself and how we want to handle the cache name. I will put together a PR and you can comment.

PR #880, opened 2025-06-11
Add ability to download files by full .URL

zachcp:
# Motivation

I'm working on implementing MoViewSpec (#787) and that spec uses URLs or local files as opposed to codes.

# Approach

- update the `StructureDownloader.download()` fn to check for URL
- if URL then divert to  StructureDownloader.download._download_url()
- otherwise use the same cache.
- currently can access this mechanism via mn.Molecule.fetch()

```python
# currently I can imagine using `fetch`. 
# we could also add a top level `fetch_url` or similar
import molecular nodes as mn
mol = mn.Molecule.fetch(code="https://files.rcsb.org/download/4ozs.cif", format="cif")

```


zachcp:
ping @BradyAJohnston

BradyAJohnston:
I think this would be better suited under Molecule.load() that can just potentially load from a url.

molecularnodes/entities/molecule/base.py
```diff
@@ -167,6 +170,9 @@ def _read(
         file_path : Union[Path, io.BytesIO]
             The file path to the file which stores the atomic coordinates.
         """
+        if StructureDownloader.is_valid_url(file_path):
```
zachcp, line 173: Note: this largely replicates approach of mn.Molecule.fetch() so that could potentially be consolidated into mn.Molecule.load() the same way.

molecularnodes/download.py
```diff
@@ -29,6 +30,71 @@ def __init__(self, cache: str | Path | None = CACHE_DIR):
...
+    @staticmethod
+    def is_valid_url(url):
+        try:
+            result = urlparse(url)
+            return bool(result.scheme and result.netloc)
+        except Exception:
+            return False
+
+    @staticmethod
+    def download_url(
```
zachcp, line 42: I added this as its own method for conceptual clarity. There is a lot of shared code with the `download` method

zachcp:
@BradyAJohnston Updated and added a few comments for your consideration

zachcp:
I am going to close this and roll it into a MolViewSpec PR

PR #882, opened 2025-06-18
Add basic trajectory API

PardhavMaradani:
* Added basic trajectory API as per this [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/870#discussioncomment-13466894) in the discussion thread #870 
* Added tests for the new API
* Added doc generation for `MNSession` class

cc: @yuxuanzhuang

### DIFF
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -2,13 +2,16 @@ import os
 import pickle as pk
 from typing import Dict, Union
 import bpy
-from bpy.app.handlers import persistent
-from bpy.props import StringProperty
-from bpy.types import Context
+import MDAnalysis as mda
+from bpy.app.handlers import persistent  # type: ignore
+from bpy.props import StringProperty  # type: ignore
+from bpy.types import Context  # type: ignore
 from databpy.object import LinkedObjectError, get_from_uuid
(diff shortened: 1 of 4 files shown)


molecularnodes/session.py
```diff
@@ -154,6 +162,110 @@ def clear(self) -> None:
         """Remove references to all molecules, trajectories and ensembles."""
         self.entities = {}
 
+    def remove(self, uuid: str) -> None:
```
BradyAJohnston, line 165: Should this just be combined with `remove_entity()`? Maybe it should be a single `remove()` method that has an argument to potentially not delete an object?

PardhavMaradani: `remove` is what is getting called from the UI operator today. I added the `remove_entity` method so that there is just one place where entities get deleted from the dictionary - as opposed to each one doing their own `del self.entities[...`. This is to complement the `register_entity`, which is also the single place to add to the dictionary. I made this change so that if we get to add corresponding blender properties, then there is just one place to add / delete them. Does this sound ok?

BradyAJohnston: Yep seems OK.

BradyAJohnston:
The tests that are failing are unrelated and being fixed elsewhere.

PR #887, opened 2025-06-19
Add UI for entities in 3D Viewport

PardhavMaradani:
* Add blender properties corresponding to entities that hold `uuid` and entity type
* Link creation/deletion of session entities to corresponding blender properties
* Add panels for entities listing and trajectory details
* Add tests for all new functionality

### DIFF
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -9,7 +9,6 @@ from bpy.types import Context  # type: ignore
 from databpy.object import LinkedObjectError, get_from_uuid
 from MDAnalysis.core.groups import AtomGroup
 from .entities import Molecule
-from .entities.base import EntityType
 from .entities.ensemble.base import Ensemble
 from .entities.trajectory.base import Trajectory
 from .nodes.nodes import styles_mapping
@@ -80,33 +79,10 @@ class MNSession:
     def register_entity(self, item: Union[Molecule, Trajectory, Ensemble]) -> None:
         """Add entity to the dictionary"""
         self.entities[item.uuid] = item
-        # add entity to blender properties if it doesn't exist
-        props = bpy.context.scene.mn
-        entities = props.entities  # entities collection
-        if entities.find(item.uuid) == -1:
-            entity = entities.add()  # add new entity to collection
-            entity.name = item.uuid  # immutable name that allows find to get index
-            # EntityType is not available here in most cases as it is set after __init__
-            # In other cases, it is reset later like in case of MD_OXDNA
-            # So, use a simple check based on supported entity types here
-            if isinstance(item, Molecule):
-                entity.type = EntityType.MOLECULE.value
-            elif isinstance(item, Trajectory):
-                entity.type = EntityType.MD.value
-            elif isinstance(item, Ensemble):
-                entity.type = EntityType.ENSEMBLE.value
-            props.entities_active_index = len(entities) - 1
 
     def remove_entity(self, uuid: str) -> None:
         """Remove entity from the dictionary"""
         del self.entities[uuid]
-        # remove entity from blender properties
-        props = bpy.context.scene.mn
-        entities = props.entities
-        index = entities.find(uuid)
-        if index != -1:
-            entities.remove(index)  # remove entity from collection
-            props.entities_active_index = len(entities) - 1
 
     def match(self, obj: bpy.types.Object) -> Union[Molecule, Trajectory, Ensemble]:
         return self.get(obj.uuid)
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -135,12 +135,12 @@ class MNSession:
         # remove any properties that don't exist in session
         props = bpy.context.scene.mn
         entities = props.entities
-        pruned = False
-        for i, entity in enumerate(entities):
-            if entity.name not in self.entities:
-                entities.remove(i)  # remove entity from collection
-                pruned = True
-        if pruned:
+        remove_indices = [
+            i for i, entity in enumerate(entities) if entity.name not in self.entities
+        ]
+        if remove_indices:
+            for i in remove_indices:
+                entities.remove(i)
             props.entities_active_index = len(entities) - 1
 
     @property
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -9,6 +9,7 @@ from bpy.types import Context  # type: ignore
 from databpy.object import LinkedObjectError, get_from_uuid
 from MDAnalysis.core.groups import AtomGroup
 from .entities import Molecule
+from .entities.base import EntityType
 from .entities.ensemble.base import Ensemble
 from .entities.trajectory.base import Trajectory
 from .nodes.nodes import styles_mapping
@@ -79,10 +80,33 @@ class MNSession:
     def register_entity(self, item: Union[Molecule, Trajectory, Ensemble]) -> None:
         """Add entity to the dictionary"""
         self.entities[item.uuid] = item
+        # add entity to blender properties if it doesn't exist
+        props = bpy.context.scene.mn
+        entities = props.entities  # entities collection
+        if entities.find(item.uuid) == -1:
+            entity = entities.add()  # add new entity to collection
+            entity.name = item.uuid  # immutable name that allows find to get index
+            # EntityType is not available here in most cases as it is set after __init__
+            # In other cases, it is reset later like in case of MD_OXDNA
+            # So, use a simple check based on supported entity types here
+            if isinstance(item, Molecule):
+                entity.type = EntityType.MOLECULE.value
+            elif isinstance(item, Trajectory):
+                entity.type = EntityType.MD.value
+            elif isinstance(item, Ensemble):
+                entity.type = EntityType.ENSEMBLE.value
+            props.entities_active_index = len(entities) - 1
 
     def remove_entity(self, uuid: str) -> None:
         """Remove entity from the dictionary"""
         del self.entities[uuid]
+        # remove entity from blender properties
+        props = bpy.context.scene.mn
+        entities = props.entities
+        index = entities.find(uuid)
+        if index != -1:
+            entities.remove(index)  # remove entity from collection
+            props.entities_active_index = len(entities) - 1
 
     def match(self, obj: bpy.types.Object) -> Union[Molecule, Trajectory, Ensemble]:
         return self.get(obj.uuid)
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -126,11 +126,22 @@ class MNSession:
         """
         Remove any entities that no longer exist in Blender
         """
+        # remove any entities that don't have linked objects
         for uuid in list(self.entities):
             try:
                 _ = self.entities[uuid].name
             except LinkedObjectError:
                 self.remove_entity(uuid)
+        # remove any properties that don't exist in session
+        props = bpy.context.scene.mn
+        entities = props.entities
+        pruned = False
+        for i, entity in enumerate(entities):
+            if entity.name not in self.entities:
+                entities.remove(i)  # remove entity from collection
+                pruned = True
+        if pruned:
+            props.entities_active_index = len(entities) - 1
 
     @property
     def n_items(self) -> int:
@@ -378,4 +389,14 @@ class MN_OT_Session_Create_Object(bpy.types.Operator):
         return {"FINISHED"}
 
 
-CLASSES = [MN_OT_Session_Remove_Item, MN_OT_Session_Create_Object]
+class MN_OT_Session_Prune(bpy.types.Operator):
+    bl_idname = "mn.session_prune"
+    bl_label = "Session Prune"
+    bl_description = "Prune session entities by removing ones that no longer exist"
+
+    def execute(self, context: Context):
+        get_session().prune()
+        return {"FINISHED"}
+
+
+CLASSES = [MN_OT_Session_Remove_Item, MN_OT_Session_Create_Object, MN_OT_Session_Prune]
--- a/molecularnodes/blender/utils.py
+++ b/molecularnodes/blender/utils.py
@@ -1,6 +1,5 @@
 from pathlib import Path
 import bpy
-from bpy.app.handlers import persistent  # type: ignore
 
 
 def path_resolve(path: str | Path) -> Path:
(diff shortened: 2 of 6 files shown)


BradyAJohnston:
Thanks for putting this together, visually and layout it looks good and I agree with your comments in the other thread. 

For displaying of the properties, rather than creating a new property group, we could potentially just use the existing `bpy.data.objects`. UILists have an option for filtering, so if we just filter to only return those objects which have a corresponding entity registered with the session, then we have one less place to try and keep in sync when changing data.

The `MN_UL_EntitiesList` would essentially remain the same but you can add a filter method which would reduce the list to just the same number of items (as we are having only 1 object per entity): https://docs.blender.org/api/current/bpy.types.UIList.html#advanced-uilist-example-filtering-and-reordering

What do you think?

PardhavMaradani:
Hi Brady, thanks for the pointer. I tried this out and here are my observations:

The `filter_items` function itself in `MN_UL_EntitiesList` is pretty simple:

```
    def filter_items(self, context, data, propname):
        objects = getattr(data, propname)
        helper_funcs = bpy.types.UI_UL_list
        # Default return values.
        filtered = []
        ordered = []

        filtered = [self.bitflag_filter_item] * len(objects)
        session = context.scene.MNSession
        for i, object in enumerate(objects):
            if object.uuid not in session.entities:
                filtered[i] &= ~self.bitflag_filter_item

        ordered = helper_funcs.sort_items_by_name(objects, "name")
        return filtered, ordered
```

and the `bpy.data.objects` is passed as follows in the main panel draw:

```
        row.template_list(
            "MN_UL_EntitiesList",
            "entities_list",
            bpy.data,
            "objects",
            props,
            "entities_active_index",
            rows=3,
        )
```

However, there are are a few issues:

- The `index` param in the `draw_item` is the index of the filtered item in the whole list, so there didn't seem to be any way to do a numbered list starting from 1 (like in the example) - this is probably not a big deal if we are ok to get rid of the numbering
- The `active_index` property is and index into the whole list, so keeping it up to date during additions / deletions is pretty messy - not just updates in our control, but even when regular objects get added / deleted. I couldn't figure out any reliable way to get this working after trying out for a bit
- We cannot use the object's `hide_viewport` directly as this property is not animatable. We'd probably need another property in the object (which has a `set` and a `get`) to deal with visibility - same way as it is done by the `visible` property for entity in this PR (both viewport and render). This is again probably not a big deal

It is the `active_index` issue that worries me and I didn't find many examples that handles filtering and this together correctly. If you are ok with the selected item in the entities list changing randomly (to something else, or completely deselected) due to other object related changes by the user (not just additions/deletions, even renaming etc) - then we can change over to this approach. Please let me know. Thanks

BradyAJohnston:
Hmm it does seem like the filtering of the list of items doesn't seem to be the best. The problem with maintaining a second list of items in a PropertyGroup is that if a user deletes an object via the Blender UI (`x` to delete in the 3D viewport) instead of the minus button on the entities UIList, then the data is out of sync. I think we need to try and come up with a solid way of ensuring the data is in sync or maybe trying to handle this filtering.

PardhavMaradani:
I do like the filtering approach because it is less code and maintains data integrity. I think we could do our best for things that we control (like when adding / deleting) entities to ensure the active index is correctly updated. Given the index is of `bpy.data.objects`, which is fairly changeable, we have this unique issue. Maybe this is the small price we pay for data consistency - is is just the numbering and the retaining of the selection. I can try a bit more to see what can be done to minimize these issues.

BradyAJohnston:
Trying to think of a good way to handle it. Maybe we create a custom text property that stores the name of the currently selected object / entity, then part of the poll function is checking where in the filtered list that object is and setting the selection index from that?

PardhavMaradani:
If I remember correctly, changing props in the `poll` or `draw` context could lead to some exceptions - we ran into some of them in the past.

PardhavMaradani:
Hi Brady, I added the commit which has the filtering based changes and removed the property collection. I made changes to keep the selection in sync with the active object change and vice versa. With these changes, things seem to work reasonably well in my testing. Could you please take a look. I can always undo and go back to the previous changes if you see any issues with this. Please let me know.

PardhavMaradani:
Just a quick summary of both the approaches:

custom property group: Predictable behavior with respect to selections from the ui list. Entities deleted from Blender (using `x` or from some other API, not ours) lead session entities to get out of sync - this is an existing problem and not something this PR introduces. These session entities get pruned today in `update_entities`.

filtering: Can use `bpy.data.objects` directly and not need separate props for session entities. Has challenges with keeping the selection predictable. Entities deleted from Blender still lead to out of sync in our session object. The ui list will be in sync, but this is something that can be achieved in the above case too.

While writing the above, it seems like we should just stick with the custom property group and the simpler and predictable approach - though with a bit more property handling code. Please let me know if you agree - the current code has the filtering changes due to the most recent commit, so I can bring them back to the old approach. Thanks

ps. If we are to go back to the old approach, we could `filter_items` to exclude invalid entities so that the ui list is in sync. Note that we cannot change props or call other operators from within the `poll` or `draw` context due to the `AttributeError: Writing to ID classes in this context is not allowed` and `RuntimeError: Calling operator "___" error, can't modify blend data in this state (drawing/rendering)` errors.

yuxuanzhuang:
I got an error when I tried to run this branch with Blender Development addon. It seems to be related to the `register_msgbus_subscriptions`.

Did I do anything wrong here?

<details>

```
  Exception in module register(): /Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/__init__.py
    File "/Applications/Blender.app/Contents/Resources/4.4/scripts/modules/addon_utils.py", line 515, in enable
      mod.register()
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/ui/addon.py", line 69, in register
      register_msgbus_subscriptions()
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 62, in register_msgbus_subscriptions
      _subscribe_to_msgbus(None)
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 56, in _subscribe_to_msgbus
      _subscribe_to_active_object_changes()  # subscribe to active object changes
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 46, in _subscribe_to_active_object_changes
      owner=bpy.context.scene.MNSession,
            ^^^^^^^^^^^^^^^^^
  AttributeError: '_RestrictContext' object has no attribute 'scene'
  ADDONS_TO_LOAD [{'load_dir': '/Users/scottzhuang/git_rego/MolecularNodes_GSOC/molecularnodes', 'module_name': 'molecularnodes'}]
  INFO: Removing old link: /Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes
   * Serving Flask app 'Blender Server'
   * Debug mode: on
  Sending: {'type': 'setup', 'blenderPort': 7240, 'debugpyPort': 7713, 'blenderPath': '/Applications/Blender.app/Contents/MacOS/blender', 'scriptsFolder': '/Applications/Blender.app/Contents/MacOS/4.4/scripts', 'addonPathMappings': [{'src': '/Users/scottzhuang/git_rego/MolecularNodes_GSOC/molecularnodes', 'load': '/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes'}]}
  Waiting for debug client.
  Debug client attached.
  Warning: User Default: [Errno 2] No such file or directory: '/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/user_default'
  register_class(...): already registered as a subclass 'MN_PT_Scene'
  register_class(...): already registered as a subclass 'MN_UL_EntitiesList'
  register_class(...): already registered as a subclass 'MN_PT_Entities'
  register_class(...): already registered as a subclass 'MN_PT_trajectory'
  register_class(...): already registered as a subclass 'MN_OT_Add_Custom_Node_Group'
  register_class(...): already registered as a subclass 'MN_OT_Residues_Selection_Custom'
  register_class(...): already registered as a subclass 'MN_OT_Assembly_Bio'
  register_class(...): already registered as a subclass 'MN_OT_iswitch_custom'
  register_class(...): already registered as a subclass 'MN_OT_Change_Color'
  register_class(...): already registered as a subclass 'MN_OT_Node_Swap'
  register_class(...): already registered as a subclass 'MN_OT_Import_Fetch'
  register_class(...): already registered as a subclass 'MN_OT_Import_OxDNA_Trajectory'
  register_class(...): already registered as a subclass 'MN_OT_Import_Trajectory'
  register_class(...): already registered as a subclass 'MN_OT_Reload_Trajectory'
  register_class(...): already registered as a subclass 'MN_OT_Import_Map'
  register_class(...): already registered as a subclass 'MN_OT_Import_Star_File'
  register_class(...): already registered as a subclass 'MN_OT_Import_Cell_Pack'
  register_class(...): already registered as a subclass 'MN_OT_Import_Protein_Local'
  register_class(...): already registered as a subclass 'MN_OT_Import_Molecule'
  register_class(...): already registered as a subclass 'MN_FH_Import_Molecule'
  register_class(...): already registered as a subclass 'MolecularNodesObjectProperties'
  register_class(...): already registered as a subclass 'MolecularNodesSceneProperties'
  register_class(...): already registered as a subclass 'TrajectorySelectionItem'
  register_class(...): already registered as a subclass 'MN_UL_TrajectorySelectionListUI'
  register_class(...): already registered as a subclass 'MN_OT_Universe_Selection_Add'
  register_class(...): already registered as a subclass 'MN_OT_Universe_Selection_Delete'
  register_class(...): already registered as a subclass 'MN_OT_Template_Install'
  register_class(...): already registered as a subclass 'MN_OT_Template_Uninstall'
  register_class(...): already registered as a subclass 'MolecularNodesPreferences'
  register_class(...): already registered as a subclass 'MN_OT_Session_Remove_Item'
  register_class(...): already registered as a subclass 'MN_OT_Session_Create_Object'
  register_class(...): already registered as a subclass 'MN_MT_Node'
  register_class(...): already registered as a subclass 'MN_MT_Node_Animate'
  register_class(...): already registered as a subclass 'MN_MT_Node_Assembly'
  register_class(...): already registered as a subclass 'MN_MT_Node_Bonds'
  register_class(...): already registered as a subclass 'MN_MT_Node_Color'
  register_class(...): already registered as a subclass 'MN_MT_Node_Density'
  register_class(...): already registered as a subclass 'MN_MT_Node_DNA'
  register_class(...): already registered as a subclass 'MN_MT_Node_Style'
  register_class(...): already registered as a subclass 'MN_MT_Node_Select'
  register_class(...): already registered as a subclass 'MN_MT_Node_Topology'
  register_class(...): already registered as a subclass 'MN_MT_Node_Attributes'
  register_class(...): already registered as a subclass 'MN_MT_Node_Geometry'
  register_class(...): already registered as a subclass 'MN_MT_Node_Simulation'
  register_class(...): already registered as a subclass 'MN_MT_Node_Curves'
  register_class(...): already registered as a subclass 'MN_MT_Node_Utils'
  register_class(...): already registered as a subclass 'MN_MT_Node_Fields'
  Exception in module register(): /Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/__init__.py
    File "/Applications/Blender.app/Contents/Resources/4.4/scripts/modules/addon_utils.py", line 515, in enable
      mod.register()
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/ui/addon.py", line 69, in register
      register_msgbus_subscriptions()
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 62, in register_msgbus_subscriptions
      _subscribe_to_msgbus(None)
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 56, in _subscribe_to_msgbus
      _subscribe_to_active_object_changes()  # subscribe to active object changes
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/scottzhuang/Library/Application Support/Blender/4.4/extensions/vscode_development/molecularnodes/blender/utils.py", line 46, in _subscribe_to_active_object_changes
      owner=bpy.context.scene.MNSession,
            ^^^^^^^^^^^^^^^^^
  AttributeError: '_RestrictContext' object has no attribute 'scene'
  Error: '_RestrictContext' object has no attribute 'scene'
    File "/Users/scottzhuang/.vscode-insiders/extensions/jacqueslucke.blender-development-0.0.25/pythonFiles/include/blender_vscode/load_addons.py", line 222, in load
      bpy.ops.preferences.addon_enable(module=addon_name)
    File "/Applications/Blender.app/Contents/Resources/4.4/scripts/modules/bpy/ops.py", line 109, in __call__
      ret = _op_call(self.idname_py(), kw)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  RuntimeError: Error: '_RestrictContext' object has no attribute 'scene'
```
</details>

PardhavMaradani:
Hi Yuxuan, I will take a look the error if we decide to go with this approach (filtered `bpy.data.objects`) vs the custom properties. If it is the later, this code won't be involved. I'll make a note to debug  this anyway - seems like there is no `scene` in the context.

@yuxuanzhuang , it seems like there is a concept of [Restricted Context](https://archive.blender.org/wiki/2015/index.php/Extensions:2.6/Py/API_Changes/#Restricted_Context) during `register()` and `unregister()`. Since I ran the `register()` manually from the notebook, I didn't run into this issue, but would probably have when enabling the extension. I'll take a closer look at how to fix this if we go down this path. Thanks

PardhavMaradani:
Hi Brady, after quite a bit of testing both approaches, I moved back to the old approach (props based) and added further validations, cleanup and filtering to ensure the list is in sync with what shows up in Blender. All of this works fairly well now. Could you please review. Thanks

BradyAJohnston:
I've tested it out and it works well. I'm happy to go with your choice here as you've spent the most time with it. An initial look of the code looks good to me. Will have a proper review tomorrow but should be good to go.

molecularnodes/session.py
```diff
@@ -102,11 +126,22 @@ def prune(self) -> None:
...
                 _ = self.entities[uuid].name
             except LinkedObjectError:
                 self.remove_entity(uuid)
+        # remove any properties that don't exist in session
+        props = bpy.context.scene.mn
+        entities = props.entities
+        pruned = False
+        for i, entity in enumerate(entities):
+            if entity.name not in self.entities:
+                entities.remove(i)  # remove entity from collection
```
yuxuanzhuang: You are modifying the list that is being iterated. It's probably gonna lead to unexpected errors.

yuxuanzhuang: is it being tested?

yuxuanzhuang (changes requested):
I am able to install this version and everything runs smoothly without errors. Really love this UI for entities.

Just two nitpicking
- Would it be possible to select the entity object when clicking on it in the list?
- The filter function in the UI doesn’t seem to be working.

molecularnodes/session.py
PardhavMaradani: I did test this, but redid this to eliminate any uncertainty in the latest commit

PardhavMaradani:
> * Would it be possible to select the entity object when clicking on it in the list?

The latest commit should address this. However, the reverse - ie., updating the entity selected in our ui based on what's selected/active in the outliner will require the msgbus subscription - for the active object changes etc, so that doesn't exist.

> * The filter function in the UI doesn’t seem to be working.

We don't have access to the object name in the entity property collection - just the uuid, to lookup the entity from the session. It is for this reason, I didn't find a way to do the filtering based on name in the `filter_items` for this ui list. I'll make a note of this for the future.

BradyAJohnston:
This all looks good for now. Some minor tweaks can be made as we develop this more moving forward.

PR #888, opened 2025-06-22
Add style branches support for Trajectories

PardhavMaradani:
* Refactor common methods from Molecue to Entity base
* Add GUI for styles - supports both Molecule and Trajectory entities
* Selection can be either MDAnalysis selection phrase or AtomGroup
* Tests for new functionality - both API and GUI operator calls

cc: @yuxuanzhuang 

@BradyAJohnston , this should not change any existing functionality with one minor exception. The node tree that is setup in `_setup_modifiers` was re-using the same node tree after deleting and adding a new entity, because of which style branches are added to the older tree. The change I made is to pass `fallback=False` to ensure a new tree is always set up when creating a new object.

### DIFF
--- a/molecularnodes/entities/trajectory/base.py
+++ b/molecularnodes/entities/trajectory/base.py
@@ -10,6 +10,7 @@ from ...blender import coll, path_resolve
 from ...nodes.geometry import (
     add_style_branch,
 )
+from ...nodes.nodes import styles_mapping
 from ...nodes.styles import (
     StyleBase,
 )
@@ -732,6 +733,11 @@ class Trajectory(MolecularEntity):
         Trajectory
             Returns self for method chaining.
 
+        Raises
+        ------
+        ValueError
+            If an unsupported style string is passed
+
         Notes
         -----
         If a selection is provided, it will be evaluated and stored as a new
@@ -739,6 +745,11 @@ class Trajectory(MolecularEntity):
         """
         if style is None:
             return self
+
+        if isinstance(style, str) and style not in styles_mapping:
+            raise ValueError(
+                f"Invalid style '{style}'. Supported styles are {[key for key in styles_mapping.keys()]}"
+            )
 
         if selection is not None:
             attribute_name = "sel_0"
--- a/molecularnodes/entities/molecule/base.py
+++ b/molecularnodes/entities/molecule/base.py
@@ -15,6 +15,7 @@ from ...nodes import nodes
 from ...nodes.geometry import (
     add_style_branch,
 )
+from ...nodes.nodes import styles_mapping
 from ...nodes.styles import (
     StyleBase,
 )
@@ -333,6 +334,11 @@ class Molecule(MolecularEntity, metaclass=ABCMeta):
         Molecule
             Returns self for method chaining.
 
+        Raises
+        ------
+        ValueError
+            If an unsupported style string is passed
+
         Notes
         -----
         If a MoleculeSelector is provided, it will be evaluated and stored as a new
@@ -340,6 +346,11 @@ class Molecule(MolecularEntity, metaclass=ABCMeta):
         """
         if style is None:
             return self
+
+        if isinstance(style, str) and style not in styles_mapping:
+            raise ValueError(
+                f"Invalid style '{style}'. Supported styles are {[key for key in styles_mapping.keys()]}"
+            )
 
         if isinstance(selection, str) and selection not in self.list_attributes(
             drop_hidden=False


PardhavMaradani:
I will fix the test errors in `test_dna.py`
update: This is fixed now with the latest commit

yuxuanzhuang:
These all test really great! Just two suggestions for future improvements (as a possible TODO item, I know you have a lot going on)

1. Would it be possible to temporarily hide the style? (I assume a simple switch could do the trick.)
2. When hovering over the text, a placeholder text appears: “Input value used for unconnected socket.” It might be nicer if it could directly reuse the text from the corresponding geometric node input instead.

molecularnodes/entities/trajectory/base.py
```diff
@@ -681,3 +690,82 @@ def _update_positions(self, frame):
...
+        -------
+        Trajectory
+            Returns self for method chaining.
+
+        Notes
+        -----
+        If a selection is provided, it will be evaluated and stored as a new
+        named attribute on the trajectory with an automatically generated name (sel_N).
+        """
+        if style is None:
```
yuxuanzhuang, line 746: Add a helpful error message when an invalid style is provided, including a list of valid options.

PardhavMaradani:
> * Would it be possible to temporarily hide the style? (I assume a simple switch could do the trick.)

This will bring us back to some of the items in discussion #879 . The current model revolves around a single style node, so there is no logical grouping of nodes that comprise a given style branch (unless we walk back the node tree from that style node). I didn't add any new properties to keep track of anything extra (which also has the potential to get invalidated with manual changes) and am relying on the current API that returns the node details directly from the node tree. A few additional things missing (compared to what I had in the prototype) in addition to the visibility are: color selection, selection phrase update, style switching. A node group approach would add these possibilities, but we decided to go with this approach based on this [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/879#discussioncomment-13467304)

> * When hovering over the text, a placeholder text appears: “Input value used for unconnected socket.” It might be nicer if it could directly reuse the text from the corresponding geometric node input instead.

Unfortunately, this doesn't seem to be possible as of now. A request for this feature in Blender can be found [here](https://blender.community/c/rightclickselect/16J3/) and [here](https://blender.stackexchange.com/questions/214240/is-it-possible-to-give-a-custom-description-to-a-native-blender-property-in-my-u) is other users experiencing the same

I will make a note of these issues.

molecularnodes/entities/trajectory/base.py
PardhavMaradani: Added the validations which show the supported values as part of the latest commit. Thanks

yuxuanzhuang:
> A few additional things missing (compared to what I had in the prototype) in addition to the visibility are: color selection, selection phrase update, style switching.

I think the selection phrase update would be something essential to make it really useful.

I tried to use the "walk back the node tree" approach and it seems fine, maybe there's caveat I missed?

```python
if input.name == "Selection":
    obj = context.active_object
    if not input or not input.is_linked:
        # or create one here.
        continue
    from_node = input.links[0].from_node
    if from_node.bl_idname == "GeometryNodeInputNamedAttribute":
        sel_attr = from_node.inputs["Name"].default_value
    else:
        continue
    if sel_attr in entity.selections:
        item = obj.mn_trajectory_selections[
            obj.mn.trajectory_selection_index
        ]
        # add a text field for selection string
        row.prop(item, "selection_str", text="Selection String")
    continue
```

PardhavMaradani:
I see a few issues:

- We cannot use `context.active_object` because that could be pointing to anything based on user selection in the outliner (note that we are not updating our selection based on the active object changes). This is not a big deal, because we could use `entity.object` here that is already pointing to the right object instead
- We cannot use the `trajectory_selection_index` because then the value here depends on the selection in that UI - this will change if that is changed. We would instead have to do a lookup based on the selection name to get the index, which is possible
- The `mn_trajectory_selections` properties and the `selections` attribute are only for Trajectories, so this has to be guarded - Molecules use the implicit selections, so we should probably expose the direct attribute name
- As you already noted in the comment `or create one here` - the named attribute node might or might not exist based on its initial creation - because the nodes are only selectively created today
- `AtomGroup` selections currently use the selection name as the selection phrase in the property - we should probably make it non-editable like it is in the current MN UI
- This will currently not make it exposable to the API, unless we do a similar walk back in the `GeometryNodeInterFace` - even if we did, we have to expose the options of allowing a selection phrase (string) or an `AtomGroup` through the API and hence need separate handling for it

If you guys want me to explore exposing these additional options as well - while retaining the current node tree model, I could try, but possibly as a separate PR because these will involve nodes beyond the current style node and are all susceptible to user changes etc. This current PR doesn't change any existing behavior (of what was introduced for `Molecule` entity) and stays within the style node bounds.

BradyAJohnston:
I think this is good as a current PR. As it's not going directly into main additional functionality can come from additional PRs.

PR #891, opened 2025-07-06
Add base framework for annotations

PardhavMaradani:
* Base annotation, interface and manager classes
* Dynamic annotation classes support - with corresponding API, GUI and interface
* Geometry Nodes as annotation data stores
* Dynamic operator support to add annotation types from GUI
* Add base annotation support for Trajectory and Molecule entities
* Add pytests for all new functionality
* Add doc entries for new classes

cc: @yuxuanzhuang 

This PR is based on the annotation support discussion in #889 . This PR does not have the actual annotation drawing code yet. It has the framework to support dynamic annotation classes, the corresponding API and GUI. The annotation drawing code will be added as a separate PR and will include the Trajectory specific annotations that we will support by default. Could you please review.


PardhavMaradani:
Hi Brady, the test failures don't seem to be related to changes in this PR. The failures are in blender daily and pass in blender 4.4

BradyAJohnston:
Hi @PardhavMaradani I've had a look through some of the code and importantly had a bit of a play around with the prototype. 

I think it's an interesting idea and you've done well to develop it, but the UX of having a node inside of the node tree that the user is not meant to interact with just breaks the UX too much of using a node tree. The node tree is intended for GUI interaction, and for us to start adding nodes that the user isn't meant to interact with goes against this too much.

BradyAJohnston:
I'll comment and we can have some more discussion around property groups on the discussion #889 to see how we can potentially make that work instead.

PardhavMaradani:
I will change the underlying data layer to use properties instead of nodes. We are already creating dynamic property groups for the annotation inputs from the  GUI. Instead of having all possible properties together (which can lead to name conflicts), I will create annotation type specific inputs as a `PointerProperty` to the corresponding dynamic `PropertyGroup`. I verified that this is feasible. Please let me know if you need any more changes. Thanks

PardhavMaradani:
I updated the PR. The underlying datastore is now Blender properties. Rest of the functionality remains exactly the same. Could you please review. Thanks

PardhavMaradani:
Hi @BradyAJohnston , all the test failures are unrelated to this PR. It looks like this branch needs the databpy update changes merged?

BradyAJohnston:
Yep no worries I'll deal with the databpy errors today and go through the review today as well. Thanks for putting this together!

molecularnodes/ui/panel.py
```diff
@@ -827,11 +827,159 @@ def draw(self, context):
...
+
+    def filter_items(self, context, data, propname):
+        if data is None:
+            return [], []
+        items = getattr(data, propname)
+        helper_funcs = bpy.types.UI_UL_list
+        filtered = []
+        # Filtering by name
+        if self.filter_name:
+            filtered = helper_funcs.filter_items_by_name(
```
BradyAJohnston, line 869: This could just be replaced by  `self.filter_items_by_name(` as the self is the same time, and you can remove the `helper_funcs`

molecularnodes/ui/ops.py
```diff
@@ -869,6 +873,118 @@ def invoke(self, context, event):
         return context.window_manager.invoke_confirm(self, event, title="Remove Style?")
 
 
+def _register_temp_annotation_add_op(entity, annotation_type):
```
BradyAJohnston, line 876: I think we can drop the need to have any kind of popup when a new annotation is added (or at the very least only the first popup for annotation type). 

The particular settings for an annotation will all potentially be changed after the annotation is added, so there is no need for a popup where they will be entered. Once the annotation is added the user can just edit them afterwards. 

This saves us interupting the flow with another popup, and we also don't have to register a temporary operator and we can drop all of this code here.

We just add the annotation, display the properties and the user can change them afterwards which they would anyway.

molecularnodes/annotations/manager.py
```diff
@@ -0,0 +1,315 @@
...
+            annotation_type = annotation_class.annotation_type
+            AnnotationInputs = create_annotation_type_inputs(annotation_class)
+            bpy.utils.register_class(AnnotationInputs)
+            AnnotationProperties.__annotations__[annotation_type] = (
+                bpy.props.PointerProperty(type=AnnotationInputs)
+            )
+        # Re-register the new AnnotationProperties class
+        bpy.utils.register_class(AnnotationProperties)
+        # Re-assign the annotation properties to Object - old data is retained
+        bpy.types.Object.mn_annotations = bpy.props.CollectionProperty(
```
BradyAJohnston, line 122: I'm a bit unsure about this. As this is working on the `bpy.types.Object`, this changes it for every object in the scene all at once? Every time this is invoked this would remove all existing collection properties?

BradyAJohnston (changes requested):
I've left some questions and comments. 

I'm wondering if for the sake of simpler code, we just have one large property group that is re-used for all potential annotations, rather than trying to create new property groups dynamically based different annotations.

In terms of maintainability going into the future, the simplest solution is best rather than over-complicating creation. Interested to get your thoughts on it.

PardhavMaradani:
> I'm wondering if for the sake of simpler code, we just have one large property group that is re-used for all potential annotations, rather than trying to create new property groups dynamically based different annotations.
> 
> In terms of maintainability going into the future, the simplest solution is best rather than over-complicating creation. Interested to get your thoughts on it.

Hi Brady, it seems like we are on different pages on why this is being done. Let me explain. Right from the first comment in discussion #889 , I mentioned that I wanted to do this in such a way that allows end users to be able to create their own custom annotations for their own analysis needs. I wanted this to be extensible the same way that MDAnalysis has `MDAnalysis.analysis.base.AnalysisBase` to write custom analysis classes, without having to modify MDA code.

Here is a clipped version from the video in discussion [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/889#discussioncomment-13630978) that shows what I mean:

https://github.com/user-attachments/assets/18f5f5ba-6983-4039-a08b-2c844e1ffa80

At the beginning of the video, you see that there is only one annotation type - 'atom_info' in the GUI and API. Think of this as what is packaged with MN. The user then creates a custom annotation in their own notebook as you see in the next cell. This is the custom 'bond_info' annotation that shows up later in the GUI along with that annotation specific inputs - all of which is user defined. So, if someone wants to draw custom annotations for say [Helix analysis](https://userguide.mdanalysis.org/stable/examples/analysis/structure/helanal.html) from the MDAnalysis user guide, they should be able to do so without changing MN code - that is what I am shooting for.

So, with this, there **cannot** be one large property group that has all potential annotation properties as we just don't know what users could have. It is the same framework that we will use for the annotations that we package (built-in) with MN. Note that the exploratory `bmda` module had all properties bundled together as seen [here](https://github.com/PardhavMaradani/bmda/blob/main/bmda/ui.py#L337) because there was no custom annotations support - you can also see how messy it gets with that property group.

So, the larger question here is whether or not we want to support user extensible annotations?

If the answer is yes, we need the ability for dynamic properties. I am pretty certain that the only two approaches that exist are either using Geometry Nodes (which are extensible via inputs) or dynamic blender properties.

If the answer is no, and we only need support for built-in annotations, we can hard code them all into one property group and don't need all this dynamic stuff.

Please let me know your thoughts.

molecularnodes/ui/panel.py
PardhavMaradani: Replacing with `self.filter_items_by_name` shows the following error:

```
molecularnodes/ui/panel.py", line 869, in filter_items
    filtered = self.filter_items_by_name(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MN_UL_AnnotationsList' object has no attribute 'filter_items_by_name'
```

`self` and `helper_funcs` are of different types:

```
helper_funcs = bpy.types.UI_UL_list

class MN_UL_AnnotationsList(bpy.types.UIList)
```

This is also how the UIList [example](https://docs.blender.org/api/current/bpy.types.UIList.html) uses this. So, we can retain this as is?

molecularnodes/ui/ops.py
PardhavMaradani: This goes back to my [comment](https://github.com/BradyAJohnston/MolecularNodes/pull/891#issuecomment-3055300943) about user defined annotations. Users can define custom inputs for their annotation type. Some of these inputs could be required and others optional. The dynamically added API call `add_<annotation_type>` does validations that the required inputs are passed before actually adding the annotation - else it throws a `ValueError` exception.  The GUI also uses the exact same API call to add the annotations, so these inputs have to be read from the user - especially if they are required params as indicated by the user. We could potentially drop the dialog if there are no mandatory params - we already do this for the case when there are no inputs at all in the annotation. Note that once an annotation property is added to the collection, it immediately becomes part of what's executed during the `draw` code - it might not be ideal to add defaults to user required details to just bypass a dialog - it would still be a two step process.

molecularnodes/annotations/manager.py
PardhavMaradani: No, it doesn't. As mentioned in the comment, old data is retained. It was for these uncertainties that I didn't know before that I chose the Node based approach as it seemed easily extensible and could come in handy in the future. I have since verified that this approach with dynamic blender properties works fine. Blender retains properties as ID props when the corresponding classes are not registered and converts them to `API defined` (as seen in the GUI) when the property classes are registered. You can see a similar approach [here](https://github.com/McManning/DynamicProperties)

molecularnodes/ui/panel.py
BradyAJohnston: Sorry that was my mistake on misreading it.

molecularnodes/ui/ops.py
BradyAJohnston: Having two popups one after the other isn't great UX, so I think we can drop the second one with options. While it will immediately start drawing, we can have sensible defaults that won't do anything too drastic.

BradyAJohnston:
Okay thanks for the extra info and explanation. You are right that we should stick with dynamic properties to try and support user definer annotations.

We can definitely skip the dynamic operator registration though.

molecularnodes/ui/ops.py
PardhavMaradani: We can definitely merge the two popups into one - ie., the annotation type selection drop down followed by the annotation type specific inputs - this would need to be a dynamic operator. That I think would be better for validations and initial error handling than guessing defaults. If you are ok with this, I can make these changes as part of the rest of the changes for annotations in later PRs. Please let me know if there are any other changes you need in this PR. Thanks

BradyAJohnston: Yep I'm happy with that.

PR #897, opened 2025-07-10
Remove micrograph tests

BradyAJohnston:
In Blender 5.0 the daily tests are failing specifically for the micrograph loading of .star files. 

I haven't been happy with how this is handled for a while, given it's done in the background all on one node. We should have a single node for displaying an ensemble, then another node that can potentially display a micrograph:

CleanShot 2025-07-10 at 18 29 05@2x

This current node could essentially be split into two, and the image loaded when the additional node is added rather than subscribing to depsgraph changes and trying to update then.

### DIFF
--- a/molecularnodes/entities/ensemble/star.py
+++ b/molecularnodes/entities/ensemble/star.py
@@ -32,7 +32,6 @@ class StarFile(Ensemble):
         self.object = blender_object
         self.data = self._read()
         self._create_mn_columns()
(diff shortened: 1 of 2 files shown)


BradyAJohnston:
Ping @jojoelfe you'll be the main user of this. Haven't yet added back an operator to re-add the image node with the correct image but should be relatively simple.

jojoelfe:
Thanks! I know this was always a bit fragile. I'll try to find some time to look into it.

PR #900, opened 2025-07-11
Add custom validation support for annotations

PardhavMaradani:
* Single add annotations dialog

### DIFF
--- a/molecularnodes/annotations/base.py
+++ b/molecularnodes/annotations/base.py
@@ -12,7 +12,8 @@ class BaseAnnotation(metaclass=ABCMeta):
     manager and '__init__' to pass the entity to annotations.
 
     Entity annotations will have to implement the 'draw' method that specifies
-    how to display the annotations. An optional 'defaults' method can be provided
+    how to display the annotations An optional 'validate' method can be provided
+    to validate annotation inputs An optional 'defaults' method can be provided
     to set default values to the annotation.
 
     Attributes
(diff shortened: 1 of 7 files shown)


PardhavMaradani:
Hi Brady, this is a small PR. It makes the add annotation dialog a single one - with the option to select the type and provide that type specific inputs (as agreed upon in the last PR). It also adds basic validation support for annotation inputs - common for both from the API and GUI. Any errors during the draw calls are not seen in the notebook cells or blender console because they run in a different context, so we need robust error handling during additions and updates. The basic framework will exist with this PR. Thanks

BradyAJohnston:
It's mostly looking good to me. I might just not be able to find it because I'm less familiar, but is there somewhere that the temporary operator is unregistered after we are done with it?

PardhavMaradani:
Hi Brady, we currently remove it before creating a new one. We can unregister the temp operator at the end of execute as well (ie., right before returning `FINISHED` - I verified this), but there are a couple of cases where it will still remain: 1) When the operator is cancelled (using the cancel button instead of ok in the invoke dialog) - we don't get any control back 2) When there is an exception in the operator execute - if we try to unregister in the `finally` block, we cannot raise the exception, which is not ideal. Note that there is only only registered operator by that id. It is for these above reasons, I left the remove before creating a new one. I'll make a note to see if there are ways around this. Thanks

BradyAJohnston:
Thanks for the details @PardhavMaradani as always. Current solution seems fine!

PR #901, opened 2025-07-12
Annotations draw handler base

PardhavMaradani:
* Lifecycle of the viewport draw handler
* Base for annotations draw handler code
* Basic 3D and 2D text drawing
* Working AtomInfo annotation
* Better handling of temp annotation add operator classes

cc: @yuxuanzhuang

### DIFF
--- a/molecularnodes/entities/trajectory/annotations.py
+++ b/molecularnodes/entities/trajectory/annotations.py
@@ -77,15 +77,13 @@ class AtomInfo(TrajectoryAnnotation):
         params = self.interface
         universe = self.trajectory.universe
         # check if selection phrase is valid - mda throws exception if invalid
-        _ag = universe.select_atoms(params.selection)
+        self.atom_group = universe.select_atoms(params.selection)
         return True
 
     def draw(self) -> None:
         params = self.interface
-        universe = self.trajectory.universe
-        atom_group = universe.select_atoms(params.selection)
         # iterate over each atom in the atom group
-        for atom in atom_group:
+        for atom in self.atom_group:
             text = atom.name
             if params.show_resid:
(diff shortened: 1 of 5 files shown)


PardhavMaradani:
Hi Brady, the actual drawing code (for text, lines, etc) will change as it evolves - especially with support for rendering etc. I want to layer these changes in small steps rather than one big change. Thanks

molecularnodes/entities/trajectory/annotations.py
```diff
@@ -63,22 +63,32 @@ class AtomInfo(TrajectoryAnnotation):
...
-        # check if selection phrase is valid - mda throws exception
+        # check if selection phrase is valid - mda throws exception if invalid
         _ag = universe.select_atoms(params.selection)
         return True
 
     def draw(self) -> None:
         params = self.interface
-        print(self.trajectory, params.selection, params.show_resid, params.show_segid)
+        universe = self.trajectory.universe
+        atom_group = universe.select_atoms(params.selection)
```
BradyAJohnston: Creating the atom group inside of the draw code might not be ideal. Creation of the new AtomGroup is relatively slow, so if this can be done once (creation / whenever selection changes) then re-used for the actual draw call that would be ideal.

BradyAJohnston (changes requested):

molecularnodes/entities/trajectory/annotations.py
PardhavMaradani: Hi Brady, this should be fixed now with the new commit. Thanks

BradyAJohnston:
LGTM this works well! I'll note here that the annotation and the 3D model get out of sync if any averaging / subframes are applied, but if we change the average / subframes to be applied as a Universe transformation then it would solve that problem.

PR #903, opened 2025-07-13
Add basic line drawing support for annotations

PardhavMaradani:
* Pointer support for 3D text
* 2D text drawing support with normalized viewport coords
* 3D and 2D line drawing support (with arrow ends)
* Trajectory - Center-of-Mass annotation
* Trajectory - Distance between COMs annotation

### DIFF
--- a/molecularnodes/annotations/manager.py
+++ b/molecularnodes/annotations/manager.py
@@ -96,7 +96,7 @@ class BaseAnnotationManager(metaclass=ABCMeta):
 
         # Add a dynamic wrapper to ensure custom signature is retained
         def dynamic_wrapper(cls, annotation_class, **kwargs):
-            cls._create_annotation_instance(annotation_class, **kwargs)
+            return cls._create_annotation_instance(annotation_class, **kwargs)
 
         method = functools.partialmethod(dynamic_wrapper, annotation_class)
         parameters = [


PardhavMaradani:
cc: @yuxuanzhuang 

Continued work on annotations. Basic lines support. Here is a brief video with the progress so far:

https://github.com/user-attachments/assets/10dc22e2-fc44-44f5-b8cd-921e1a5c21e0

PardhavMaradani:
About the Codecov report above: The `draw` code will only get executed in Blender's viewport draw handler - which will also not get called in background mode. Some parts of it will get called later when we have rendering support for annotations (to PIL image), but there will always be a coverage gap - similar to how it is for `panel.py`.

PardhavMaradani:
Hi @BradyAJohnston , I have new changes that support using `AtomGroup`s in Annotation API calls in addition to selection phrases and they work seamlessly with the GUI - while using the same attribute, etc. Do you  want me to add those to this PR or wait till this is merged? They are based on top of these changes - there is also a bit of cleanup so the diff might look noisy if added to this PR, hence checking. Thanks

BradyAJohnston:
You can open a new PR and once this one is merged it'll be cleaner on the new PR as well.

Sorry haven't had a chance to tea this out yet hopefully this evening.

BradyAJohnston (approved):
LGTM, good to see it keep coming along.

PR #906, opened 2025-07-17
Add depth support for annotation text

PardhavMaradani:
Hi @BradyAJohnston and @yuxuanzhuang , this PR adds support for text depth in annotations. This is currently disabled by default and the option is available at an annotation level. So, there is no change of functionality by default with respect to the previous PR. There is some code refactoring, but it is useful refactoring.

If the text depth option is enabled (at annotation level), the size of the text varies depending on its  position from the virtual viewport camera, creating a perception of depth. Without this, everything is of the same size and could look pretty flat. Currently only the text size is impacted with this option, but we could also include the alpha and even add a clip param to control how much gets displayed - especially in the case of any dense text.

Here is a short video that shows the difference with this option enabled and disabled:

https://github.com/user-attachments/assets/375b1aac-9ae9-4e7f-b96f-b14571931367

This is also not an expensive computation. Only the object bounding box coords are used to get the min and max viewport distance and the reduction factor is determined based on it. If you think this looks better, we could make this the default later after more testing.

### DIFF
--- a/molecularnodes/annotations/base.py
+++ b/molecularnodes/annotations/base.py
@@ -101,20 +101,7 @@ class BaseAnnotation(metaclass=ABCMeta):
             Text to display. '|' as multi-line separator
 
         """
-        params = self.interface
-        text_pos = pos_3d
-        if params.pointer_length > 0:
-            # draw a pointer to the 3D position
-            nv = -Vector(pos_3d)
-            nv.normalize()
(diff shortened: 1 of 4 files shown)


BradyAJohnston:
I think this all seems reasonable. I think it would make a good default setting if it proves to work well.

PR #907, opened 2025-07-19
Add canonical dihedrals trajectory annotation

PardhavMaradani:
* Add support for drawing circles and arcs in 3D space (with arrows)
* Use arcs with arrows to indicate dihedral angle direction

### DIFF
--- a/molecularnodes/entities/trajectory/annotations.py
+++ b/molecularnodes/entities/trajectory/annotations.py
@@ -218,3 +218,75 @@ class COMDistance(TrajectoryAnnotation):
             v1_arrow=True,
             v2_arrow=True,
         )
+
+
+class CanonicalDihedrals(TrajectoryAnnotation):
+    """
+    Canonical Dihedrals of a given residue
+
+    Attributes
+    ----------
+    resid: int
+        The residue number
+
+    show_atom_names: bool
+        Whether or not to show the individual atom names in the residue
+
+    show_direction: bool
+        Whether or not to show the arc indicating the angle direction
+
+    """
+
+    annotation_type = "canonical_dihedrals"
+
+    resid: int
+    show_atom_names: bool = True
+    show_direction: bool = True
+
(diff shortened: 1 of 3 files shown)


PardhavMaradani:
Hi @BradyAJohnston and @yuxuanzhuang , this PR adds the canonical dihedrals annotation for trajectories.

Here are a couple of examples of how this looks for resid 4 of our sample:

canonical-dihedrals-r-4-example-1

canonical-dihedrals-r-4-example-2

I added support to draw circles and arcs in 3D space, which was used above to show the direction of the angles.

molecularnodes/entities/trajectory/annotations.py
```diff
@@ -218,3 +218,75 @@ def draw(self) -> None:
...
+                v2 = a2.position
+                text = None
+                if ai == 1:
+                    text = f"{symbols[si]} = {selections[si].dihedral.value():1.2f} °"
+                    if params.show_direction:
+                        center = (v1 + v2) / 2
+                        radius = 0.1 * self.distance(v1, v2)  # 10% of distance
+                        # draw arc indicating direction of the angle
+                        self.draw_circle_3d(
+                            center, radius, (v2 - v1), angle=270, c_arrow=True
```
BradyAJohnston: Would it make more sense to have the arc size match the angle of the dihedral that has been computed?

PardhavMaradani: Hi Brady, yes, makes sense. Even better, I made the arc start in the first plane rather than an arbitrary point - this way the arc goes from the first plane to the second plane.

Here's a quick verification of this: (planes are drawn only for illustration)

This image shows the two planes and the arc:

resid-4-phi-example-1

This image shows the N and CA aligned to ensure the arc is indeed between these two planes:

resid-4-phi-example-2

BradyAJohnston: great improvements!

PR #909, opened 2025-07-24
Rendering support for annotations

PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds support for rendering annotations to an image.

I haven't added any automatic compositing, because that should be independent of this and something we could add later during other rendering related changes. With this PR, an annotations image that can be composited using a simple alpha over is generated.

Here is a short video that shows this:

https://github.com/user-attachments/assets/d447194f-13ef-4942-a483-23ba0d43ddb1

This works for lines as well as can be seen in this example:

annotations-render-lines-example

Text with depth also works as expected.

I haven't added the test case for this yet because the render call seems to cause pytest to hang. A quick look at it seemed to because of the shader compilation involved in the `bpy` module. I will take a closer look at it later and see if there is a way around it. (There is currently no test with a render call - even the Canvas.snapshot() hangs in my setup)


PardhavMaradani:
The docs build failure is related to this. I will take a look and fix it - this is fixed now.

PardhavMaradani:
I am unable to call the render op in the tests after having tried several different things (it hangs with eevee, works on ubuntu with cycles, fails saving image in windows and fails to render at all on macos, etc). There is also no other render call in tests for reference, so I tried to see if there was a way. I removed the test and the PR is now ready for review. Thanks

BradyAJohnston:
For the failing render tests, we will need to look into this more in the future to get something working for testing but also just for usability for someone using this in an automated pipeline across platforms. 

Can you add in the failed tests but just skip them for for now? That way I can potentially have a look at them and we can use them as the yardstick for when it starts working.

PardhavMaradani:
Hi Brady, I added the test back and marked it to be skipped for now. I will also take a look at the failure again at a later time. The single windows failure after this commit seems to be a transient download error. Thanks

BradyAJohnston:
I think we'll want to definitely try and get tests for this before final merge, but happy for this as is for merge into the GSOC branch.

PR #910, opened 2025-07-26
Trajectory Universe Info, generic 2D and 3D label annotations

PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds a universe info annotation, which displays the universe info details in the viewport / renders. The positioning of it can be adjusted and all aspects of what gets displayed can be controlled.

This PR also adds support for showing the annotations correctly scaled in the 3D viewport camera view. The universe info annotation is displayed with normalized coords, so it shows up at the intended location in viewport, viewport camera view and in the final renders as well.

Here is a short video that shows this:

https://github.com/user-attachments/assets/f991f97d-7a74-4c87-bcee-d9fa67e2c23a

@BradyAJohnston , this PR is on top of the changes in #909, which is not merged yet. Once that is merged, I will resolve any merge conflicts as I believe there will be some because of changes in common lines. Thanks

PR #913, opened 2025-07-28
Use property indexing instead of reference to avoid crashes

PardhavMaradani:
This PR is on top of PR #910 

This fixes a nasty bug because the created property interface uses references to the blender properties. The problem shows up after multiple annotations, which causes a re-alloc that invalidates the old data.  Similarly, undo also invalidates all references (as learnt later). This is documented in Blender [gotchas](https://docs.blender.org/api/current/info_gotchas_crashes.html). As recommended, we now use indexed access to the properties and all these problems and crashes go away.


PardhavMaradani:
The windows failure is unrelated to this change.

BradyAJohnston (approved):

PR #916, opened 2025-07-30
Density entity - style interface, gui, annotations

PardhavMaradani:
* Fix density entity to work with style interface
* Add GUI support for Density entity in 3D viewport
* Add annotations support for Density entity
* Fix bug in AlphaFold import prop
* Set 3D text depth to True by default
* Fix bug in checking for invalid entities in draw handler

### DIFF
--- a/molecularnodes/annotations/manager.py
+++ b/molecularnodes/annotations/manager.py
@@ -399,6 +399,8 @@ class BaseAnnotationManager(metaclass=ABCMeta):
         self._remove_annotation_instance(self._interfaces[uuid])
 
     def _draw_handler_add(self):
+        if bpy.app.background:
+            return
         if self._draw_handler is not None:
             return
         self._draw_handler = bpy.types.SpaceView3D.draw_handler_add(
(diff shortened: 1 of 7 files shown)


PardhavMaradani:
Hi Brady, these are just preliminary changes to bring the density entity inline with `Molecule` and `Trajectory` so that it has the same style interface, shows up in the viewport GUI, has annotations support, etc.

density-entity-gui-annots

There are also a few minor fixes I found along the way. Thanks

PardhavMaradani:
I will add more details about the test failures on 4.4 tomorrow
Update: The previous test [errors](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/16623497319/job/47033568579) in 4.4 was because `nanobind` was showing leak errors for the `grid` that we are saving in the entity (This is currently not being used, but I didn't remove it because it will be useful later for example when we have some density specific annotations that can reference the `grid` or for the API in general). Other versions don't seem to have this test being run. The issue was because the draw handler was getting registered, which has a reference to the entity (`density` here), which in turn has a `grid` object. In the regular GUI case, the draw handler gets automatically removed for invalid entities. For the background mode (`bpy`), the handler doesn't have to be registered at all because there is no `gpu`/`blf` available and there is no drawing involved. The render case still works fine through the PIL image route. The fix was to just skip registering the draw handler in the background mode.

BradyAJohnston (approved):

PR #917, opened 2025-08-02
Use MDAnalysis GridDataFormats for density grids

PardhavMaradani:
* Remove MRC class - MRC files are supported in gridData
* Add a new Grids class that supports all grids of gridData library
* Support for different grid types based on the grid data type
* Existing support for inverting and centering a grid
* Supports both pyopenvdb (pre 4.4) and openvdb (4.4+)
* Allow threshold min, max and type specification for node input
* Add water.dx and apbs.dx test files (gzipped)

### DIFF
--- a/molecularnodes/nodes/nodes.py
+++ b/molecularnodes/nodes/nodes.py
@@ -343,6 +343,7 @@ def create_starting_nodes_density(
     # make the node tree of this node independent (single user)
     # to allow separate configuration of min, max and default threshold values
     node_tree_copy = node_density.node_tree.copy()
+    node_tree_copy.name = f"{styles_mapping[style]}.{object.name}"
     node_density.node_tree = node_tree_copy
 
     items_tree = node_density.node_tree.interface.items_tree
--- a/molecularnodes/nodes/geometry.py
+++ b/molecularnodes/nodes/geometry.py
@@ -201,6 +201,8 @@ class GeometryNodeInterFace(TreeInterface):
                 .removeprefix("ball_and_stick_")
                 .removeprefix("cartoon_")
                 .removeprefix("spheres_")
+                .removeprefix("density_surface_")
+                .removeprefix("density_wire_")
             )
             if isinstance(input, bpy.types.NodeSocketMaterial):
                 prop = getset_material(input)


PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR is on top of PR #916 and implements what is described in this [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/914#discussioncomment-13957946) of discussion #914. With this, we can support all the grid formats that MDAnalysis [GridDataFormats](https://github.com/MDAnalysis/GridDataFormats) supports - including the existing `MRC` file support without any loss of existing functionality.

PardhavMaradani:
I resolved the merge conflicts after PR #916 is merged and this PR is now ready for review. Thanks

molecularnodes/nodes/nodes.py
```diff
@@ -333,7 +340,24 @@ def create_starting_nodes_density(object, threshold=0.8, style="density_surface"
     node_output.location = [800, 0]
 
     node_density = add_custom(group, styles_mapping[style], [400, 0])
+    # make the node tree of this node independent (single user)
+    # to allow separate configuration of min, max and default threshold values
+    node_tree_copy = node_density.node_tree.copy()
```
BradyAJohnston, line 345: I don't know if this is necessary. I think we can just stick with the same general max min (which are soft thresholds anyway) and the default node group value won't really matter as we'll set the value manually anyway. 

I think it's not worth duplicating node groups just for min / max thresholds. The thresholds are soft anyway and only limit a user from dragging above / below. A user can always manually put in values above below.

PardhavMaradani: Hi @BradyAJohnston , I added this after I found the defaults and the no bounds to be not that usable. For example, the way to reset a value to default in the UI (by hovering over the item and using backspace) will revert it back to 0.8 and not the 99.5 quantile value etc - even though the user can change this, this didn't seem ideal. Besides, if we are to support different grid types - not just floats (as is in the code), we would have to change the type, which again, cannot be done without making it independent. Even though these were soft thresholds, it seemed like better ux to kind of see the bounds.

Also, for the next part which has the slice ranges as you see in the example below:

image

Not having the bounds makes it completely unusable. The min / max values of the grid (as seen in the grid axes annotation) are needed for this. The alternative to not duplicating the node group would be to pass all these as inputs or atttributes, both of which didn't seem ideal. Please let me know what you think. Thanks

BradyAJohnston: Thanks for the extra info and you've convinced me we should duplicate the node tree for each instance. For visual feedback to the user, in case they delete and have to re-add the node, you should change the `name` of the node tree to reflect which density this is for and potentially the `label` as well.

PardhavMaradani: Hi Brady, I added the object name to the duplicated node tree name so this can be used again. I also fixed the API part that wasn't correctly exposing the density style values. The test failures below seem to be due to alpha fold download errors. Thanks

PR #918, opened 2025-08-07
Density Annotations and additions

PardhavMaradani:
* Add density info and density grid axes annotations
* Add custom font configuration for text (works for both viewport and renders)
* Add molecule info annotation
* Separate annotation types per entity (to avoid any name collisions)
* Add vertical spacing support for multiline text
* Tests for new molecule and density annotations

### DIFF
--- a/molecularnodes/annotations/base.py
+++ b/molecularnodes/annotations/base.py
@@ -196,9 +196,9 @@ class BaseAnnotation(metaclass=ABCMeta):
         if not isinstance(pos, Vector):
             pos = Vector(pos)
         pos_2d = pos
+        params = self.interface
         if is3d:
             text_pos = pos
-            params = self.interface
             # check if pointer is required
             if params.pointer_length > 0:
                 # draw a pointer to the 3D position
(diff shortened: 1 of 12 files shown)


PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds the changes described in this [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/914#discussioncomment-14013711) of discussion #914

PardhavMaradani:
All the test failures are due to alpha fold download errors like in the previous PR. However, note that the same alpha fold download test runs fine on my local machine.

BradyAJohnston:
Seems to be an error with the alphafold server causing all of these failing tests. I'm going to update to skip the tests for now so we can get at least passing tests for merging these PRs.

BradyAJohnston:
Local tests are passing because they are using cached files instead of re-downloading (something to improve for the tests)

PardhavMaradani:
I noticed daily builds passing. I didn't see an option in my GitHub UI to re-run the tests, so made an empty commit to trigger the tests. The  two windows failures are due to the presence of the `MolecularNodesCache` folder (known) and the one ubuntu failure seems to be a temporary alpha fold download error.

BradyAJohnston:
Okay good to see that AlphaFoldDB seems to have fixed itself.

PR #920, opened 2025-08-10
Density ISO Surface Style

PardhavMaradani:
* Separate colors for positive and negative ISO values
* Contours support in 3D viewport
* Slicing along the 3 axes in both directions

Hi @BradyAJohnston , @yuxuanzhuang , this PR is on top of PR #918 and adds the new density iso surface style.

Here is a screenshot of this:

style-density-iso-surface


PardhavMaradani:
Hi @BradyAJohnston , just wanted to note one observation about a material setting:

All the MN materials seem to have the backface culling setting turned on. For the above new style in this PR, I am re-using the same MN materials. In the prototype, I had used separate materials for the positive and negative colors (and the default is backface culling off) and hence noticed the difference. With backface culling on, especially for surfaces (both regular and when sliced), this doesn't seem to work well.

Here is a side by side with backface culling enabled on the left and the defaults on the right (I enabled the show contours option to highlight the actual surface):

density-backface-culling

I didn't find an easy way to expose this option (`bpy.data.materials[...].use_backface_culling`) in our style interface. Even if we did, this will impact all the others using this material. Just wanted to bring this to your attention.

PardhavMaradani:
I added a few more options for contours and added support for handling node input panels in the most recent commit. Here are the current options available from both API and GUI for this new style:

image

molecularnodes/entities/density/annotations.py
```diff
@@ -151,7 +155,7 @@ def draw(self) -> None:
             return
         origin = grid.origin.copy()
         if grid.metadata["center"]:
-            origin += -np.array(grid.grid.shape) * 0.5 * grid.delta
+            origin = -np.array(grid.grid.shape) * 0.5 * grid.delta
```
BradyAJohnston, line 158: This is no longer updating the old origin - I'm assuming because the origin is always at 0? Not sure if we ever need to take into account non-zero origin.

molecularnodes/nodes/nodes.py
```diff
@@ -339,25 +343,44 @@ def create_starting_nodes_density(
...
-    # to allow separate configuration of min, max and default threshold values
-    node_tree_copy = node_density.node_tree.copy()
-    node_tree_copy.name = f"{styles_mapping[style]}.{object.name}"
-    node_density.node_tree = node_tree_copy
+    if style == "density_iso_surface":
+        key = "ISO Value"
+        node_density = group.nodes.new("GeometryNodeGroup")  # type: ignore
+        node_density.name = styles_mapping[style]
+        node_density.location = [400, 0]
+        node_tree = style_density_iso_surface_node_group()
```
BradyAJohnston: I would change the name of `node_tree` to instead just `tree`. The `node_` prefix is being used through to signify it's an actual node whereas this is the tree. Could just be `tree` or something else.

BradyAJohnston:
Overall looks good. Up until this point I have avoided using the "Node to Python" to generate scripts for node groups, mostly because of how ridiculous verbose it makes the final scripts.

I'd _prefer_ to have a much smaller script for a node tree, so that looking at version control of it is easier to figure out what is going on. I think I'm happy to accept this as is as there isn't really a better alternative at the moment, but in the future I'm hoping to come up with a better node -> code solution.

molecularnodes/nodes/nodes.py
PardhavMaradani: I changed this to `tree` in the latest commit

molecularnodes/entities/density/annotations.py
PardhavMaradani: Yes, noticed the bug with the `+=` when testing out the slicing with the centered option. The object location was hard coded to `(0, 0, 0)` in `create_object` in the previous `mrc.py` file too, which is what I am using in the `grids.py` file too.

PardhavMaradani:
I found the `Node To Python` script quite useful here. If you look through the commit history, there were a few times I needed to update the node tree and in all cases, I updated this in the UI (and verified there) before exporting back, which made it simpler - including checking the diff. I used explicit node names to keep this readable and see what is being done. The only complaint I have is with the use of indices in the `inputs` and `outputs`, which would've been nice with the actual names. I agree we could stick with this for now till we have a better option.

PardhavMaradani:
Hi Brady, any thoughts about this [comment](https://github.com/BradyAJohnston/MolecularNodes/pull/920#issuecomment-3172323727) above? The workaround currently is to change this in the material, which works, but is something that would have to be addressed later. Thanks

BradyAJohnston:
I think having the backface culling disabled by default on pretty much all materials unless they are required specifically is the way to go, so changing that in a future change is good.

PR #926, opened 2025-08-14
Automatic compositor setup to handle annotations overlay in renders

PardhavMaradani:
* User configurable auto setup option for compositor
* Option for manual compositor setup from n-panel in composite editor
* New composite node group to add annotations render image as overlay on final renders
* Object visibility checks in scene before drawing annotations
* Minor updates in arrange_tree to work for composite node trees


PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds the compositor setup that is described in discussion #925.

molecularnodes/nodes/mn_compositor_node_group.py
BradyAJohnston, line 1: No need to prefix this filename with `mn_` as we are in the MN repo / project after all. Just rename to `compositor.py` I think.

molecularnodes/nodes/mn_compositor_node_group.py
```diff
@@ -0,0 +1,85 @@
+# Auto-generated by 'Node To Python' extension
+# https://extensions.blender.org/add-ons/node-to-python/
+import bpy
+
+
+# initialize MN Compositor node group
+def mn_compositor_node_group():
```
BradyAJohnston: same with prefix here

BradyAJohnston:
All looks and works good but I think just some naming changes.

molecularnodes/nodes/mn_compositor_node_group.py
PardhavMaradani: I renamed the file above, but this is within the auto-generated code (which uses the Node name, which in our case is `MN Compositor` to generate this method), so I left it as is to prevent updating it every time there is a change. Please let me know if you think otherwise and I can change it too. Thanks

PR #932, opened 2025-08-18
Add rendering animation support to Canvas

PardhavMaradani:
* Add support to temporarily override any object properties (context manager)
* Add support to suppress outputs (context manager)
* Fix frame_object to work for all entities
* Refactor snapshot method
* Add animation method to render animations


PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds support for rendering animations through Canvas. It also refactors the current `snapshot` method a bit. These rendering methods don't overwrite any of the user configured render settings. Any temporary changes needed for rendering are automatically reverted back.

Here is an example video of this in action:

https://github.com/user-attachments/assets/30a766d0-2047-4b05-9311-3e7dda4c3bd6

As you can see from the right side of the video, the user render settings remain intact after these calls. This works in the background mode as well. Thanks

PardhavMaradani:
The docs build failure is unrelated to this PR and is due to a download error.

BradyAJohnston:
This looks great and I am really happy with the improvements you've made. Do you want to add the example you showed in the video as documentation that is rendered? Becomes part of the testing then and acts as user documentation as well.

PardhavMaradani:
Hi Brady, I added the example from the video above to the docs. I will add other examples in the same way as you asked in this [comment](https://github.com/BradyAJohnston/MolecularNodes/discussions/928#discussioncomment-14146603) of discussion #928.

I have also added the sample topology and trajectory files that are used across most MDAnalysis examples because they aren't big. We could move over to the ones in `MDAnalysisTests` once you add that dependency at a later time. Note that the rendered doc as per this PR won't show the annotations though the code exists for them in the example. Once the compositor setup code from PR #926 is merged, they will automatically show up.

Here is the sample rendered doc:

canvas-rendering-doc

PR #938, opened 2025-08-20
Add frame view support to Canvas

PardhavMaradani:
* Ability to frame the camera to any selection of the trajectory
* Both selection phrases and AtomGroups can be used for selections
* Support for bounding boxes based on trajectory frame
* Utility functions to frame camera to any bounding box region
* Tests for new functionality

### DIFF
--- a/molecularnodes/blender/utils.py
+++ b/molecularnodes/blender/utils.py
@@ -64,10 +64,16 @@ def get_bounding_box(obj: bpy.types.Object) -> list[tuple]:
 
 def look_at_object(obj: bpy.types.Object) -> None:
     """Set camera to look at an object"""
+    # save current selected objects list
     prev_sel = bpy.context.selected_objects
+    # de-select all objects in the scene
+    for o in bpy.data.objects:
+        o.select_set(False)
+    # select this object and set camera view
     obj.select_set(True)
     bpy.ops.view3d.camera_to_view_selected()
     obj.select_set(False)
+    # restore previous object selections
     for o in prev_sel:
         o.select_set(True)
 


PardhavMaradani:
Hi @BradyAJohnston , @yuxuanzhuang , this PR adds support to set to camera to specific regions of the trajectory. The `Canvas` class so far only supports framing the whole object. For documenting the changes made as part of GSoC, I wanted the ability to show specific regions. This PR adds a generic `look_at` that can point to any selection (based on either the selection phrase or an AtomGroup) of the trajectory. The prototype had a lot of additional functionality to look at different regions from different view points, rotate around them, etc, but all those changes are too disruptive at this stage. The changes here are pretty basic and can be easily extended going forward if needed. More importantly, they help showcase the changes made more effectively.

Here is an example to look at resid 4:

look-at-resid-4

I will file a separate issue to outline the clipping planes and focal length issues of the template.

I am open to renaming the trajectory method from `look_at` to `focus` or anything else that you think will convey this better.

PardhavMaradani:
I am not too happy with updating the camera view from within a method of an entity. I will refactor this to something more extensible like having the entities return a bounding box in a common method like `get_view` (with whatever inputs that make sense for that entity - like selection phrase or atom group for Trajectories, etc), which the `Canvas` can then use in a method like `frame_view` (along the lines of `frame_object`) to frame the camera to the required view(s). This way, we can frame across entities and all the camera handling will stay with `Canvas` etc, which would be much cleaner. Converting this to a draft till I make those changes.

BradyAJohnston:
I think ideally we would move away from using the operator `bpy.ops.view3d.camera_to_view_selected()`. We should be able to compute the transformation matrix for a camera with a given field of view for a set of points. Then we wouldn't need to create a temporary object and we could potentially try and take into account other information to 'better frame' what we are after. 

I think the name `look_at` works. Focus I think would apply more for adjusting the focal distance / aperture if DoF is enabled.

PardhavMaradani:
I agree with what you said about `bpy.ops.view3d.camera_to_view_selected()`, but maybe it is something we could consider for later because that math is beyond me at this point. I believe will have to take into account the aspect ratio, the focal length and the bbox coords etc. This operator is at least friendly and doesn't have any poll issues with the context like others and also works well in the background mode.

I refactored the code in the latest [commit](https://github.com/BradyAJohnston/MolecularNodes/pull/938/commits/9653f1343533471700f781dc0c3ed435dac3f063) and this should be much more cleaner and extensible now. I added a basic `Camera` class that helps configure basic camera params from within the canvas (and also addresses the issues outlined in #939 by setting the required defaults), support for different viewpoints so we aren't limited to the single default one in the template. We can now frame the required views across multiple entities as well. Thanks

BradyAJohnston:
Thanks for update, I'm happy with leaving operator for now and changing over later on.

BradyAJohnston (approved):

PR #940, opened 2025-08-22
Add MDAnalysisTests dependency to tests and docs

PardhavMaradani:
Hi Brady, just wanted to check if this is the correct way to add the dependency. I updated the rendering doc to use these files (and verified locally after running `uv pip install ".[docs]"` to install the deps and the regular quarto render). I removed the files added to the test data as well. The download tests were failing locally for me with a `503 Server Error: Service Unavailable for url` - maybe it is a transient issue and shouldn't be related to this. Thanks


BradyAJohnston:
Yep this is good for tests. Seems like the PDB is having some down time.

PR #960, opened 2025-09-06
Add biotite structure converter

PardhavMaradani:
Hi @BradyAJohnston , this PR adds a converter from a Biotite `AtomArray` / `AtomArrayStack` [structure](https://www.biotite-python.org/latest/apidoc/biotite.structure.html) to an MDAnalyisis universe as discussed in #949 .

We cannot pass the `AtomArray` or `AtomArrayStack` directly to `mda.Universe` like in other [converters](https://docs.mdanalysis.org/stable/documentation_pages/converters.html) because there is an issue with how MDAnalysis treats iterable inputs. MDAnalysis supports reading [multiple trajectories](https://userguide.mdanalysis.org/stable/reading_and_writing.html#reading-multiple-trajectories). When an iterable is passed as the topology, each of the iterated value is parsed for coordinates. Biotite's `AtomArray` / `AtomArrayStack` are iterable and this leads to a call to parse the individual Atom objects, which is incorrect. Either the `iterable` method in MDAnalysis `lib/util.py` has to be fixed to avoid iterating these objects (using something like a `type(obj).__name__` to prevent importing these modules) or we have to handle this at our end. I had to create a simple `BiotiteWrapper` to workaround this problem for now.

The converter code itself is pretty generic and adds all the Biotite's mandatory and optional annotations as MDAnalysis topology attributes. Only a single `AtomArray` is considered (like how it is in `Molecule` today), but I don't see why an `AtomArrayStack` with trajectory co-ordinates cannot be supported if required.

Some of the computed (like `Color`, `sec_struct`, etc) named attributes and additionally parsed and added object properties (like `obj.mn.biological_assemblies`) can directly be added to the object after we create the universe object if we go down this path. It didn't make sense to pass them on as separate topology attributes in the converter only to read them back and re-create the named attributes or blender properties - this way, the converter too remains generic.

I added a test that verifies the setup of MDAnalysis's standard topology attributes and Biotite's mandatory and optional annotations. There are some minor differences in the computed attributes between the `Molecule` and `Trajectory` case today, which I mentioned in the comments in the test.

Overall, I think having a universal MDAnalysis backend for `Molecule` and `Trajectory` entities (I haven't looked at `Ensemble` and `Density` is entirely different) is definitely possible and this converter could be a first step in that direction. It would be great if you can try these changes out as well - I made these independent so they don't break anything and can be easily used to compare existing `Molecule` entities and the universe versions of them. Thanks


PardhavMaradani:
Hi Brady, the test failures in `daily` are not related to this change. Also, about the `test_martini` in `test_trajectory.py`, we seem to be comparing the named attributes with all the object data attributes. This is problematic for a couple of reasons - when new named attributes are added, this fails (like what happened earlier because this PR added a couple of new named attributes). The snapshot includes all the obj data attributes and not just the named attributes (eg: `['vdw_radii', 'mass', 'b_factor', 'atomic_number', 'res_id', 'segid', 'res_name', 'chain_id', 'atom_types', 'atom_name', 'position', '.select_vert', 'is_backbone', 'is_alpha_carbon', 'is_solvent', 'is_nucleic', 'is_lipid', 'is_peptide', '.edge_verts', '.select_edge', 'sharp_face']` for one of them - notice the `.select_vert`, `sharp_face` etc) because of which I couldn't just add a specific list of attributes. So, I had to skip the newly added attributes because I don't know how to update the snapshot. Just fyi.

BradyAJohnston:
Thanks for putting this together! Should have time to go over it tomorrow. The daily issue is something I should be able to fix separately.

BradyAJohnston:
Re: snapshots, you can generate new snapshots via:
```bash
pytest --snapshot-update
```

How we test attributes is an interesting one. I think iterating all attributes is still worth-while, maybe skipping some of the built-in ones, but we want to make sure that the geometry being generated is what we expect. If something changes in the nodes or in the python code that changes the final geometry in some way we want to know about it. If a new attribute appears it should fail the test, which is why we then update the snapshots.

BradyAJohnston (approved):

BradyAJohnston:
If you wanted to update the snapshot then I think it's good to go.

This PR makes me think even more how we need maybe to re-think how we approach accessing attributes (from a Blender object vs from a Universe). I'll write up those thoughts in another issue - but this is good for an initial implementation of the converter.

PardhavMaradani:
> Re: snapshots, you can generate new snapshots via:
> 
> ```shell
> pytest --snapshot-update
> ```

I'll try updating the snapshots and remove the hack for skipping the new attributes in the `test_martini` - will update this PR. Thanks

PR #961, opened 2025-09-06
Density ISO style contours and prop update fixes

PardhavMaradani:
* Scale input is new in Curve to Mesh node from 4.5 - set default
* Update view layer after style property updates (need for bpy)

### DIFF
--- a/molecularnodes/blender_manifest.toml
+++ b/molecularnodes/blender_manifest.toml
@@ -1,7 +1,7 @@
 schema_version = "1.0.0"
 
 id = "molecularnodes"
-version = "4.5.0"
+version = "4.5.1"
 name = "Molecular Nodes"
 tagline = "A toolbox for molecular import and animation in Blender"
 maintainer = "Brady Johnston<brady.johnston@me.com>"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "molecularnodes"
-version = "4.5.0"
+version = "4.5.1"
 description = "Toolbox for molecular animations with Blender and Geometry Nodes."
 readme = "README.md"
 dependencies = [
--- a/tests/__snapshots__/test_attributes.ambr
+++ b/tests/__snapshots__/test_attributes.ambr
@@ -17901,288 +17901,15 @@
   NamedAttributeError("The selected attribute 'entity_id' does not exist on the mesh.")
 # ---
 # name: test_bond_attributes
-  [-0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5
-   -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5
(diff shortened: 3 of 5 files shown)


PardhavMaradani:
Hi Brady, I noticed the following discrepancies in the solvent density images under:
https://bradyajohnston.github.io/MolecularNodes/api/trajectory/solvent.html

1. The contours are not showing up
2. The last image is not framed correctly

The contours showed until the latest docs deployment and I guess this is when the bpy version changed from 4.4 to 4.5.x. A new `Scale` input was [added](https://developer.blender.org/docs/release_notes/4.5/geometry_nodes/#miscellaneous) to Curve to Mesh node. Without this default set, the value was 0 and hence the contours don't show up. Given the density iso style is in the code, I made it backward compatible. (I still had 4.4 in my setups and hence didn't catch this before)

About the last image not being framed correctly: This doesn't happen in the GUI mode and only happens in the bpy mode. Turns out that the style interface updates too need a view layer update (at least in the background mode) (we already do this for the regular annotation blender props in the setter).

This PR fixes both the above issues. Thanks

molecularnodes/nodes/style_density_iso_surface.py
```diff
@@ -508,8 +508,14 @@ def style_density_iso_surface_node_group():
     # node Curve to Mesh
     curve_to_mesh = style_density_iso_surface.nodes.new("GeometryNodeCurveToMesh")
     curve_to_mesh.name = "Curve to Mesh"
-    # Fill Caps
-    curve_to_mesh.inputs[2].default_value = False
+    if bpy.app.version >= (4, 5, 0):
+        # Scale
+        curve_to_mesh.inputs[2].default_value = 1.0
```
BradyAJohnston: Ideally if you could change these to use the identifier instead of the input index. Future-proofs against changes in node inputs that way.

BradyAJohnston: Which would defend us against the current issue where `False` is being converted to `0` for the node input for the radius

PardhavMaradani: Hi Brady, makes sense. I didn't want to just change this to use the identifier while the rest of the file used indices. We also cannot change all of them to use identifiers because they are not necessarily unique. For example, in the `Math` node, all the inputs have the same name `Value` - this is also why NodeToPython uses indices.

I instead chose to re-generate the file (clearly this is too big for me to rewrite now). I hacked the NodeToPython code to not write out defaults (by creating temporary reference nodes, sockets etc and comparing against them) for inputs, outputs and attributes as well. I also made changes to use name identifiers for inputs / outputs as long as they are unique in that node. With this, a lot of unnecessary stuff is gone (including this bug, because we don't set the default again) and code is no different to what I'd probably have written manually and only has the required stuff. (I'll try to cleanup the NodeToPython changes I made and add to a fork when I next find some time so we don't lose them if needed again)

BradyAJohnston (approved):

molecularnodes/nodes/style_density_iso_surface.py
BradyAJohnston: That's a good point about Math inputs. Improving and even creating our own nodes -> python module would be a worthwhile endeavor as well.

Issue #965, opened 2025-09-09
Crash on Render from Compositing


BradyAJohnston:
The `render_pre` handler seems to be causing crashes on render. 

I'm yet to get a minimum example to reproduce, but in some different files with more complex compositing node setups on render there is a hard crash. Removing the handler via below fixes the issue completely.

```py
bpy.app.handlers.render_pre.remove(bpy.app.handlers.render_pre[0])
```

On a simple setup with a default import everything is fine but building a more complex setup can create problems.

BradyAJohnston:
@PardhavMaradani

PardhavMaradani:
Hi Brady, I can help with debugging the crash, but would need a bit more info and hopefully a reproducible case if you already have one - doesn't have to be minimal - I can narrow it down. You mentioned removing the handler fixes the issue, so I am assuming you have some form of a reproducible case. If so, that would really help.

If not,

- Is this consistent?
- Does this happen in the background mode or GUI mode? ie., when running the `canvas.snapshot(...)` or the render image from the GUI (I am assuming this is with a render image and not an animation as the later indicates a deeper Blender issue)
- Did Blender generate a crash file in the temp folder?
- You mentioned that this is in 'complex compositing node setup' - does this mean that the compositor is already setup? (There is an option to turn off our auto compositor setup - worth a quick try if this seems to be the case)
- Do any of the entities have any annotations added?

BradyAJohnston:
Here is a file that crashes for me in 4.5. I haven't tested in headless rendering just using the GUI. 

Hard crash happens immediately upon hitting F12 / render. No annotations are set up. A protein has been imported from the PDB and a bunch of changes made in the compositor but I haven't tinkered at all with the annotations.

[double_outlines.blend.zip](https://github.com/user-attachments/files/22249562/double_outlines.blend.zip)

PardhavMaradani:
Hi Brady, I am unable to reproduce the issue with this file.

This renders fine for me as can be seen below:

<img width="1850" height="1053" alt="Image" src="https://github.com/user-attachments/assets/7eb1cc48-fea2-41ba-884f-0c9e78e97647" />

[...]

PR #968, opened 2025-09-12
Disable "Auto Setup" to fix crashing on Render

### LINKED ISSUE #967: Render crash guide through Compositor.
Please look at my report ticket this link.
https://projects.blender.org/blender/blender/issues/146109#issuecomment-1698987

I do macintosh.
Blender 4.5.3

BradyAJohnston:
We can't currently track down the cause of crashes on render and it seems to be inconsistent across platforms and machines. 

This disables the `Auto Setup` setting to false by default which will stop the crashing by default. Users can still manually enable and setup themselves if desired.

- **fixes #967**
- cause of  **#965** but not a final fix for it yet
- **bump databpy**


BradyAJohnston:
Failed tests are related to the RCSB PDB that keeps going down and falling to server files properly.

Issue #969, opened 2025-09-12
Annotations not rendering in headless mode


BradyAJohnston:
Due to annotation compositor setup being now disabled by default #968 the annotations in headless mode do not render. 

It was more important to get a fix out for users to stop the crashing. We need to just enable annotation setup when an annotation is added for the first time.

Related to #965

PardhavMaradani:
Running the `bpy.ops.mn.setup_compositor()` operator manually in the `Canvas` constructor will fix this - bug only for those using `Canvas`.

BradyAJohnston:
If you wanted to put together a PR I could merge and make another release.

PardhavMaradani:
Ok, I'll create a PR

PR #971, opened 2025-09-12
Setup compositor when creating a Canvas

PardhavMaradani:
This should address the issue in #969

### DIFF
--- a/molecularnodes/scene/base.py
+++ b/molecularnodes/scene/base.py
@@ -10,6 +10,7 @@ from tqdm.auto import tqdm
 from .. import assets
 from ..blender import utils as blender_utils
 from ..entities.base import MolecularEntity
+from ..scene.compositor import setup_compositor
 from ..session import get_session
 from ..ui import addon
 from ..utils import suppress_stdout, temp_override_properties
@@ -72,6 +73,7 @@ class Canvas:
         self.resolution = resolution
         self.camera = Camera()
         self.transparent = transparent
+        setup_compositor(self.scene)
 
     @property
     def scene(self) -> bpy.types.Scene:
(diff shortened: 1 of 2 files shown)


PardhavMaradani:
Seems like things changed a bit in the daily builds. Normally, till now, a scene has a `node_tree` attribute, which is `None` when the use nodes is not enabled. Looks like this attribute itself isn't present going forward. I'll try adding this check to see if the corresponding daily test failures pass.

PardhavMaradani:
Seems like there are more issues that need to be resolved for the failures in blender daily. I'll take a closer look at them tomorrow.

BradyAJohnston:
Related to the daily builds, 5.0 is changing for the compositor to remove the `use_nodes` and instead it'll will behave much like the material / geometry nodes where you can use different node groups. I haven't looked into what the API changes will be. 

https://projects.blender.org/blender/blender/pulls/138560

https://developer.blender.org/docs/release_notes/5.0/compositor/#removed

PardhavMaradani:
Hi Brady, thanks for the info.

Given this, I made the change in this PR to just skip setting up the compositor for 5.x as it won't work in its current form. As you filed a separate issue, we can add support for the 5.x compositor too (keeping it backward compatible) as part of that issue in a separate PR. That way, you will have this PR if you wanted to fix the current release as it is a bit broken (with doc images not having annotations, APIs not working as advertised, etc)

The reason we didn't catch this earlier is because we didn't have any render test (I see you have added a separate issue for handling those in the future). The one we added had to be skipped because it didn't quite work on all platforms - that would've gone through the pre render handler, invoking the compositor setup etc. The docs build is only for the current one and hence doesn't catch this.  Independent of the render test, we can add an explicit compositor setup test when we fix this with 5.x setup, so that we catch any such issues.

PardhavMaradani:
I have the changes for the compositor setup that work with both 5.x and 4.x tested and ready. I will create a separate PR and will let you decide if you want to include those as well. Thanks

BradyAJohnston:
Closed as implemented in #975

PR #973, opened 2025-09-12
Fix reading of .mrc and .ccp4

BradyAJohnston:
Some `.mrc` and `.ccp4` files were not being properly read by GridDataFormats. This adds a fallback parser using mrcfile to attempt with.

@PardhavMaradani if you wanted to have a look at this.

Error message when attempting to load the `.map.gz` downloaded from here: https://www.ebi.ac.uk/emdb/EMD-48397

```
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/vscode_development/molecularnodes/ui/ops.py", line 615, in execute
    density.load(
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/vscode_development/molecularnodes/entities/density/io.py", line 14, in load
    density = Grids(
              ^^^^^^
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/vscode_development/molecularnodes/entities/density/grids.py", line 32, in __init__
    self.file_vdb = self.grid_to_vdb(
                    ^^^^^^^^^^^^^^^^^
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/vscode_development/molecularnodes/entities/density/grids.py", line 166, in grid_to_vdb
    gobj = Grid(file, file_format=file_format, metadata=metadata)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/core.py", line 241, in __init__
    self.load(filename, file_format=file_format)
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/core.py", line 548, in load
    loader(filename)
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/core.py", line 559, in _load_mrc
    mrcfile = mrc.MRC(filename)
              ^^^^^^^^^^^^^^^^^
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/mrc.py", line 92, in __init__
    self.read(filename)
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/mrc.py", line 100, in read
    raise ValueError(
ValueError: MRC file /Users/brady/Downloads/emd_48397.map.gz is not a volumetric density.
Error: Python: Traceback (most recent call last):
  File "/Users/brady/Library/Application Support/Blender/4.5/extensions/.local/lib/python3.11/site-packages/gridData/mrc.py", line 100, in read
    raise ValueError(
ValueError: MRC file /Users/brady/Downloads/emd_48397.map.gz is not a volumetric density.
```

Import now correctly works.


PardhavMaradani:
> Some .mrc and .ccp4 files were not being properly read by GridDataFormats.

Hi Brady, what is the exception you were seeing for those files? Is it the `Only orthorhombic unitcells are currently supported, ...`? I think I mentioned this in one of the PRs or the status reports about this. I thought there might be some reason why `GridDataFormats` has that check.

`GridDataFormats` also uses `mrcfile` internally, but there is an explicit check for the case above. Ideally changes could go there too under a flag - but I don't know much about that stuff. If done at this end, is there a specific reason to have a separate `_MinimalGrid` class? `GridDataFormats` already provides an easy way to return a `Grid` object from data - see `tests/test_grid.py`. This returned grid object is only used by annotations today (for grid info, axes) so this will work too as long as it has the same content.

BradyAJohnston:
Sorry I did this late last night so description wasn't very useful. I've updated the description to show the error message I was getting. I've updated to just use the `Grid()` and also a test to download and parse a file which was otherwise failing.

Issue #972, opened 2025-09-12
Compositor setup changing in 5.0


BradyAJohnston:
Related to the daily builds, 5.0 is changing for the compositor to remove the `use_nodes` and instead it'll will behave much like the material / geometry nodes where you can use different node groups. I haven't looked into what the API changes will be. 

https://projects.blender.org/blender/blender/pulls/138560

https://developer.blender.org/docs/release_notes/5.0/compositor/#removed

Tests will be failing in daily as highlighted in #971 

We need some handling for it in different versions.

PardhavMaradani:
Hi Brady, this can be closed because #975 is merged?

BradyAJohnston:
Yep should be @PardhavMaradani

PR #975, opened 2025-09-13
Add compositor setup for Blender 5.x

PardhavMaradani:
The following things change from 5.x with respect to the compositor:

- The compositor node trees can be re-used and are created the same way as materials, nodes, etc
- There is no `use_nodes` checkbox anymore
- The node tree is accessible under `scene.compositing_node_group` compared to `scene.node_tree`
- There is no `Composite` node anymore. The output is a regular `Group Output` with an `Image` output socket

This should address #972 . These changes also supersede the ones in #971


BradyAJohnston:
Thanks for putting this together. Were you wanting #971 merged first or you just want this PR instead?

PardhavMaradani:
Hi Brady if you are ok with these changes, just this are fine as they include the compositor setup through the canvas as well (the other can then be closed - not merged) - to fix the background mode annotations missing in docs etc. I had the ones in #971 before I knew how much of an effort this was etc. Thanks

PR #976, opened 2025-09-14
Remove auto setup option for MN compositor

PardhavMaradani:
Removing the `Auto Setup` option as described in this [comment](https://github.com/BradyAJohnston/MolecularNodes/issues/965#issuecomment-3289267587) of #965


PardhavMaradani:
Hi @BradyAJohnston , the test failures are clearly not related to this change. The same tests (`test_session.py`, `test_trajectory.py`) ran fine locally, so maybe it is due to some other upstream change in `databpy` etc?

BradyAJohnston:
Yep sorry about that thought I went through and cleaned out references to it. Fixing in #977

BradyAJohnston:
These biotite deserialisation errors are only happening on Linux - unsure what the cause is but they seem intermittent so I am assuming it is just PDB doing weird things.

PardhavMaradani:
About the biotite deserialisation errors - that tests runs fine on my ubuntu. I specifically tried `tests/test_pdbx.py::test_entity_parsing` - after deleting `6VBU.bcif` from the cache as well and it worked fine. Maybe the problem is with cached files on the test setup? (just guessing)

BradyAJohnston:
It's a really confusing one for me. On a linux server I can replicate the issue but clearing the downloaded files doesn't help. It also seems to _sometimes_ happen on the windows machines. Downloads of `.cif` work fine it's only ever the `.bcif` files. Not sure what to do about it.

PardhavMaradani:
If you are able to replicate the issue, we could probably save the corresponding `.bcif` file for that test and check the size, checksum, parsing etc with other downloaded versions - that way, we could confirm it is a download issue etc. Thanks

BradyAJohnston:
I'll have more of a look into it when I get the chance if it persists, but all of the tests on main just passed so I am confused 🤷

PR #977, opened 2025-09-14
Remove usage of set_boolean

BradyAJohnston:
The `set_boolean()` method was removed upsteam in `databpy==0.3.0`, this removes reference to it and replaces with regular store_named_attribute.

- **bump databpy**
- **change set_boolean to store_named_attribute**


BradyAJohnston:
Failing tests are just the intermittent failures to download from the PDB.

Issue #981, opened 2025-09-18
Add annotation for drawing gromacs simulation box


adamwitmer:
Hello, I am currently working on a project that requires visualization of a gromacs simulation box over multiple frames to observe the displacement of a molecular system in response to a gromacs pull-code. I have been using GGMolvis and VMD to visualize the system, however the former does not include the capability to draw the outline of the simulation box and the latter does not produce the high quality renders that I require.

I would like to be able to use MN to render the simulation box including the outline of the cube so that I can visualize the system as a video (or over multiple frames). I looked through the documentation at https://bradyajohnston.github.io/MolecularNodes/api/ to familiarize myself with the current functionality of `molecularnodes` and to see if there are any existing functions related to drawing boxes as a basis for adding the capability to draw a simulation box. The closest thing I could find there was under the `Trajectory: Solvent Density: From Top with Grid Axes` example (https://bradyajohnston.github.io/MolecularNodes/api/trajectory/solvent.html) which uses an annotation called `add_grid_axes`. I did a `git grep` of `grid_axes` in a local installation of the `molecular nodes` repo and found a reference to the annotation in:

```bash
molecularnodes/entities/density/annotations.py:    annotation_type = "grid_axes"
```

In the docs it states:

> Annotations can be added using the add_<annotation_type> method of the annotations manager of an entity. Each entity can have different annotation types supported.

Is this a good place to start in terms of adding an annotation for something like i.e. "add_box" (with similar functionality as the class `DensityGridAxes` which uses `draw_line_3d` to create the `grid_axes`) or is there a better solution that you might suggest for adding this functionality? Thanks!

tylerjereddy:
@BradyAJohnston and I discussed this in a recent call (Adam is a postdoc at LANL).

One question for Adam might be (which we just discussed in person), even if you knew which part of `MolecularNodes` to use for this functionality, do you understand enough about the box geometry to be able to draw the eight corners at the moment? In other words, is the only blocker that you don't know which part of `MolecularNodes` to use, or is it also possible that you may not know how to get the coordinates of the corners of the box, etc?

If you know how to get the corners of the container (it is likely straightforward for rectangular shapes, but dodecahedra may be harder), you could always use `bpy` to prototype an initial mockup (place a sphere at each corner) and then ask for how you'd do it with `MolecularNodes`--that way you uncouple *understanding* the geometry of the problem from the `MolecularNodes` internals, which are two separate things to some extent.

[...]

PR #983, opened 2025-09-22
Annotation line meshes

PardhavMaradani:
Hi Brady, this PR adds support for showing annotation lines as 3d meshes as described in this [comment](https://github.com/BradyAJohnston/MolecularNodes/issues/981#issuecomment-3314601217) of issue #981

Here is an example of how this looks:

https://github.com/user-attachments/assets/d956a441-5591-4129-97d1-20681b139e4c

This is still a draft PR. Though material shows up in the GUI and API, it is not implemented yet (I might need your help here) and tests and documentation updates are not done yet. Creating this to unblock Adam and Tyler to see if this helps.

### DIFF
--- a/molecularnodes/annotations/base.py
+++ b/molecularnodes/annotations/base.py
@@ -573,6 +573,20 @@ class BaseAnnotation(metaclass=ABCMeta):
         """Internal: Draw line 3D or 2D"""
         if v1 is None or v2 is None:
             return
+
+        if is3d and self.geometry is not None:
+            geometry = self.geometry
+            i = len(geometry["vertices"])
+            # add the ends of line as vertices
+            geometry["vertices"].append(Vector(v1) * self._world_scale)
+            geometry["vertices"].append(Vector(v2) * self._world_scale)
+            # add an edge
+            geometry["edges"].append((i, i + 1))
+            params = _get_params(self.interface, overrides)
+            # add resolved params to be added as attributes
+            geometry["color"].append(params.mesh_color)
+            geometry["thickness"].append(params.mesh_thickness)
+
         self._draw_arrow_line(
             v1, v2, v1_arrow, v2_arrow, is3d=is3d, overrides=overrides
         )
@@ -639,6 +653,9 @@ class BaseAnnotation(metaclass=ABCMeta):
         if v1 is None or v2 is None:
             return
         params = _get_params(self.interface, overrides)
+        if params.line_mesh and not params.line_overlay:
+            # only draw overlays when enabled in mesh mode
+            return
         rgba = params.line_color
         line_width = params.line_width * self._scale
         if self._render_mode:
--- a/molecularnodes/annotations/interface.py
+++ b/molecularnodes/annotations/interface.py
@@ -1,3 +1,4 @@
+import bpy
 import numpy as np
 
 
@@ -15,6 +16,7 @@ class AnnotationInterface:
     # py annotations of common params for code complete
     # these are bound dynamically along with any annotation inputs
     visible: bool
+    text_font: str
     text_color: np.ndarray | list | tuple
     text_font: str
     text_size: int
@@ -29,6 +31,11 @@ class AnnotationInterface:
     line_width: float
     arrow_size: int
     pointer_length: int
+    line_mesh: bool
+    line_overlay: bool
+    mesh_color: np.ndarray | list | tuple
+    mesh_thickness: float
+    mesh_material: str | bpy.types.Material
 
     def __init__(self, instance):
         self._instance = instance
--- a/molecularnodes/annotations/manager.py
+++ b/molecularnodes/annotations/manager.py
@@ -3,11 +3,16 @@ import inspect
 from abc import ABCMeta
 from uuid import uuid1
 import bpy
+import databpy as db
+import numpy as np
 from databpy.object import LinkedObjectError
 from mathutils import Vector
+from ..blender import coll
 from ..blender.utils import get_viewport_region_from_context, viewport_tag_redraw
+from ..nodes.material import append_material
 from .base import BaseAnnotation
 from .interface import AnnotationInterface
+from .node_tree import annotations_node_tree
 from .props import (
     BaseAnnotationProperties,
     create_annotation_type_inputs,
@@ -38,6 +43,9 @@ def _validate_annotation_update(self, context, attr):
         self.valid_inputs = False
         raise exception
     self.valid_inputs = True
+    if instance._ready:
+        # update annotation object
+        entity.annotations._update_annotation_object()
 
 
 class BaseAnnotationManager(metaclass=ABCMeta):
@@ -68,6 +76,8 @@ class BaseAnnotationManager(metaclass=ABCMeta):
         self._draw_handler = None
         self._scene = None
         self._render_mode = False
+        # annotation blender object
+        self.bob = None
         # viewport params
         self._region = None
         self._rv3d = None
@@ -261,6 +271,7 @@ class BaseAnnotationManager(metaclass=ABCMeta):
             self._draw_handler_add()
         else:
             self._draw_handler_remove()
+        self._update_annotation_object()
         viewport_tag_redraw()
 
     @classmethod
@@ -305,6 +316,8 @@ class BaseAnnotationManager(metaclass=ABCMeta):
         # set the interface attribute of the annotation instance so that
         # users can access the inputs and common params using this interface
         setattr(annotation_instance, "interface", interface)
+        # instance is ready only after all the property interfaces are created
+        setattr(annotation_instance, "_ready", False)
         # call the validate method in the annotation class if specified for
         # any annotation specific custom validation
         if hasattr(annotation_instance, "validate") and callable(
@@ -393,7 +406,7 @@ class BaseAnnotationManager(metaclass=ABCMeta):
             prop_path = item.identifier
             if prop_path in ("name", "label", "type"):
                 continue
-            if item.type == "POINTER":
+            if item.type == "POINTER" and prop_path != "mesh_material":
                 continue
             prop_interface = create_property_interface(self._entity, uuid, prop_path)
             setattr(interface.__class__, prop_path, prop_interface)
@@ -402,6 +415,15 @@ class BaseAnnotationManager(metaclass=ABCMeta):
             getattr(annotation_instance, "defaults")
         ):
             annotation_instance.defaults()
+        # material default needs to be set explicitly
+        material = "MN Default"
+        if material not in bpy.data.materials:
+            append_material(material)
+        interface.mesh_material = material
+        # update annotation object
+        self._update_annotation_object()
+        # mark instance ready for object updates from property callbacks
+        annotation_instance._ready = True
         # tag redraw of viewport to refresh values in GUI
         viewport_tag_redraw()
         # return the newly created dynamic annotation interface
@@ -419,6 +441,8 @@ class BaseAnnotationManager(metaclass=ABCMeta):
         viewport_tag_redraw()
         # remove from instances registry
         del self._interfaces[uuid]
+        # update annotation object
+        self._update_annotation_object()
 
     def _remove_annotation_by_uuid(self, uuid: str) -> None:
         """Remove annotation instance by uuid - used by GUI"""
@@ -482,34 +506,44 @@ class BaseAnnotationManager(metaclass=ABCMeta):
     def _disable_render_mode(self):
         self._render_mode = False
 
-    def _draw_annotations(self):
+    def _draw_annotations(self, get_geometry: bool = False):
+        # geometry of the lines
+        geometry = {
+            "vertices": [],
+            "edges": [],
+            "color": [],
+            "thickness": [],
+        }
         # all annotations visibilty
         if not self.visible:
-            return
+            return geometry
         # object visibility
         object = self._entity.object
         if object.hide_get():
-            return
+            return geometry
         # check if object is in current scene
+        if self._scene is None:
+            self._scene = bpy.context.scene
         if object.name not in self._scene.objects:
-            return
-        # calculate the min and max viewport distance of the bounding box verts
-        # find bounding box verts
-        bb_verts = [co for co in bpy.data.objects[object.name].bound_box]
-        # viewport distance of the bounding box verts
-        bb_verts_distances = []
-        for vert in bb_verts:
-            view_matrix = get_view_matrix(self)
-            if is_perspective_projection(self):
-                # perspective
-                dist = (view_matrix @ Vector(vert)).length
-            else:
-                # orthographic
-                dist = -(view_matrix @ Vector(vert)).z
-            bb_verts_distances.append(dist)
-        # min and max
-        min_dist = min(bb_verts_distances)
-        max_dist = max(bb_verts_distances)
+            return geometry
+        if not get_geometry:
+            # calculate the min and max viewport distance of the bounding box verts
+            # find bounding box verts
+            bb_verts = [co for co in bpy.data.objects[object.name].bound_box]
+            # viewport distance of the bounding box verts
+            bb_verts_distances = []
+            for vert in bb_verts:
+                view_matrix = get_view_matrix(self)
+                if is_perspective_projection(self):
+                    # perspective
+                    dist = (view_matrix @ Vector(vert)).length
+                else:
+                    # orthographic
+                    dist = -(view_matrix @ Vector(vert)).z
+                bb_verts_distances.append(dist)
+            # min and max
+            min_dist = min(bb_verts_distances)
+            max_dist = max(bb_verts_distances)
         # iterate over all annotations
         for interface in self._interfaces.values():
             # annotation specific visibility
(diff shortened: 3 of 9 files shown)


PardhavMaradani:
Hi Brady, all the test failures are in `test_download.py` with 'Error: Too Many Requests for url:' error and not related to this change. Thanks

BradyAJohnston:
It might be a couple of days before I can go through code proper, but overall approach of creating an annotation object for world-space geometry / annotations I think is a good idea. 

All of the GHA actions are currently getting denied requests from the PDB. I wonder if maybe they have blacklisted it for some reason.

BradyAJohnston:
I still haven't had a chance to go through in detail - but I would like to avoid using `bmesh` at all. It's much less performant and planned to be removed (no deadline for that though, many years out). I would prefer we just compute vertices and edges and directly create through `databpy`.

PardhavMaradani:
Hi Brady, can you please share the references that state that `bmesh` is planned to be removed or deprecated? From everything I have read, `BMesh` is one of the main [data structures for meshes](https://developer.blender.org/docs/features/objects/mesh/mesh/) that is extensively [documented](https://docs.blender.org/api/current/bmesh.html), very widely used and even the recommended one for more flexible mesh editing from Python at the top of the regular [Mesh (ID)](https://docs.blender.org/api/current/bpy.types.Mesh.html) documentation as well. `bmesh` is also used internally for polygons in Edit-Mode since n-gon support was added and BMesh-faces is mentioned as by far the best way to manipulate geometry in the [gotchas](https://docs.blender.org/api/current/info_gotchas_meshes.html) page too. Did you mean the old `EditMesh` that [`Bmesh`](https://developer.blender.org/docs/features/objects/mesh/bmesh/) replaced?

What we have supported here as part of this PR is more than just simple vertices and edges. Not using `bmesh` and trying to get this done with the `from_pydata` from `databpy` is extremely limiting - it is also unclear if this is at all possible without re-inventing a lot of what `bmesh` already provides. I would personally rather not add a new feature than pick a design choice that definitely looks like a step backwards as it will be a pain to maintain going forward. With that said, I will obviously leave the decision regarding this PR to you. Thanks

ps. All the test failures above are the same that happened on the daily build - biotite bcif parsing failures (now fixed, but not released yet) and new alpha fold download errors.

tests/test_annotations.py
```diff
@@ -387,3 +396,51 @@ def test_annotations_render_image(self, universe, session):
...
+        assert len(color) == 5
+        assert np.all(thickness == 1.0)
+        assert np.all(color == (1.0, 1.0, 1.0, 1.0))
+        # test attribute updates
+        a.mesh_thickness = 0.5
+        a.mesh_color = (0.0, 1.0, 0.0, 1.0)
+        thickness = t.annotations.bob.named_attribute("thickness")
+        color = t.annotations.bob.named_attribute("Color")
+        assert np.all(thickness == 0.5)
+        assert np.all(color == (0.0, 1.0, 0.0, 1.0))
```
tylerjereddy: It would be preferable to use `np.testing.assert_allclose` here -- that way we get a much nicer error message if something breaks (the number of mismatches values, and the tolerance level of the violation, instead of just a boolean check).

tests/test_annotations.py
```diff
@@ -387,3 +396,51 @@ def test_annotations_render_image(self, universe, session):
...
+        assert len(thickness) == 5
+        assert len(color) == 5
+        assert np.all(thickness == 1.0)
+        assert np.all(color == (1.0, 1.0, 1.0, 1.0))
+        # test attribute updates
+        a.mesh_thickness = 0.5
+        a.mesh_color = (0.0, 1.0, 0.0, 1.0)
+        thickness = t.annotations.bob.named_attribute("thickness")
+        color = t.annotations.bob.named_attribute("Color")
+        assert np.all(thickness == 0.5)
```
tylerjereddy: same thing here -- `np.testing.assert_allclose` is likely to be preferable for the same reason as noted below

tests/test_annotations.py
```diff
@@ -387,3 +396,51 @@ def test_annotations_render_image(self, universe, session):
...
+        a.line_mesh = True
+        assert bpy.data.objects[name].data.vertices
+        # test attributes (defaults)
+        thickness = t.annotations.bob.named_attribute("thickness")
+        color = t.annotations.bob.named_attribute("Color")
+        # verify single line with arrows at both ends - 5 edges
+        assert len(thickness) == 5
+        assert len(color) == 5
+        assert np.all(thickness == 1.0)
+        assert np.all(color == (1.0, 1.0, 1.0, 1.0))
```
tylerjereddy: here and above, similar comments apply re: using NumPy testing utilities for more helpful match checking

tests/test_annotations.py
PardhavMaradani: Thanks for reviewing Tyler. Changed.

PardhavMaradani:
The virion example above is indeed a compact rhombic dodecahedron (xy-square). The default view seems to be aligning the bottom lines perfectly. Changing the viewpoint to `front` shows the lines more clearly:

```
...
canvas.frame_view(traj.get_view("all"), viewpoint="front")
...
```

Here is the output with just the above change in the `frame_view` call:

test_virus

Here is confirmation of the shape from the GUI mode as well:

https://github.com/user-attachments/assets/37665db9-3709-4c8a-9ab8-235e664dca95

tylerjereddy:
Cool. I'm certainly in favor of something like `add_simulation_box()` on my end. I probably can't give you a detailed review of the mechanics, but probably Brady will continue to.

BradyAJohnston:
Sorry for the continued delay on getting around to this. I've been wrestling with trying to get the testing infrastructure working. Tests now at least _partially_ pass now that biotite is updated. Still some random crashes on Windows that I simply can't reproduce / track down.

Will be able to have a proper look at this next week though.

PardhavMaradani:
> Still some random crashes on Windows that I simply can't reproduce / track down.

If you meant test failures on windows: I took a quick look. At least the oxdna related attribute errors clearly show the reason in the logs:

```
E             - AttributeArray 'res_name' from NewOXDNAObject('NewOXDNAObject')(domain: POINT, shape: (98,), dtype: int64)
E             + AttributeArray 'res_name' from NewOXDNAObject('NewOXDNAObject')(domain: POINT, shape: (98,), dtype: int32)
```

Given Blender's [Integer](https://docs.blender.org/manual/en/latest/modeling/geometry_nodes/attributes_reference.html#attribute-data-types) attribute is a 32-bit integer, we should probably have this set explicitly in `databpy` so that numpy doesn't pick it based on platform like what happened above when the snapshot was captured leading to the discrepancy. On the daily macos, the oxdna attributes didn't seem to be stored on the object at all. (The exceptions for `store_named_attribute` seem to be passed, so logs don't capture the reason)

BradyAJohnston:
I've been going over these logs for hours and hadn't noticed the int64 / int32 difference 🫠 I had been staring at it trying to figure out what difference it was complaining about but my brain is clearly over looking at it.

I don't have a windows machine to test with currently so have only been going off the GHA logs. 

As for why it is failing occasionally in the other instance I can't reproduce reliably or locally. It seems to be only _some_ GHA runs fail with that error. Some kind of platform error but I can't reproduce to be able to debug it.

tylerjereddy:
I have an old Windows box, I'll make a note to check the integer handling there if it isn't resolved by the weekend.

PardhavMaradani:
Just fyi that I created a draft PR #999 that outlines the reasons for the current test failures (not related to this PR, though). Thanks

BradyAJohnston:
The code I think is mostly looking good - but I think the approach needs to change a bit in terms of properties settings. 

For the properties, we should either re-use the properties between line / mesh (color / width etc) or for the UI when you change between one and the other it changes the visibility of those properties. 

Selecting between mesh / line should also be an enum and not a boolean as well.

PardhavMaradani:
Hi Brady, we cannot re-use properties between regular overlay lines and 3d mesh lines. First, the overlay lines use a thickness, which is a pixel value as they are just overlays in 2d space. The mesh wire thickness cannot use this value - the value is in the world space and the thickness observed will be based on the orientation, how far from camera etc, whereas the 2d overlay line thickness are the same irrespective of the view. Second, the 3d version of lines are great for stuff like box wireframes, grids etc, but they can be obscured for most other use cases (like for showing distances, interactions, etc). Here is a quick example of a distance:

distance-as-only-mesh

The current code allows an option to draw overlay lines on top to kind of better help with the visualization as can be seen below:

distance-as-mesh-with-overlay-line

They are thus not mutually exclusive and I think it is better to keep them independent as they are. Thanks

BradyAJohnston:
I can see having both mesh & line overlay at the same time can be useful - but having that option as a boolean behind another boolean I don't think is nice UX. 

For this it would be better to have a boolean for mesh and a boolean for line that enable & disable their respective set of options. You can independently enable & disable the visibility of both and they are each in their own separate panels. 

For an annotation like this it could also be good maybe then to also have toggles for the text as well? Each component of the annotation could then be independently enabled / disabled.

PardhavMaradani:
Hi Brady, having separate boolean toggles within an annotation to control visibility of text / lines / meshes independently might make it overly complex. Note that users can create their own inputs that control an annotation behavior (what shows up, what doesn't, etc) which now get mixed up with these additional toggles.

Instead, if we really need better organization, we can place these props (for text, lines and meshes) in separate panels as shown below:

annotation-options-ui

This will keep things separated, clean and simple. Let me know if you are fine with this and I will update the PR. Thanks

BradyAJohnston:
The panels look good. In one of your earlier examples you show that you can have the mesh but not the overlay line - is that not being controlled somehow? Would that not be useful to have a boolean to control it? https://github.com/BradyAJohnston/MolecularNodes/pull/983#issuecomment-3430900952

PardhavMaradani:
Hi Brady, this is being controlled. There is no change in functionality. Here are those options in the lines panel:

line-annotation-params

The `Line as Mesh` param controls whether lines have to be displayed as meshes. The `Line Overlay` param controls whether an overlay line has to be drawn as well on top of a line mesh. I have updated the PR with these changes. Thanks

BradyAJohnston:
Having a button that disables one setting but enables another button that then re-enables the original functionality isn't great UX.

This should either be two booleans to enable / disable displaying the line & mesh separately, or it should be a 3-button enum to choose display method (line, mesh, line + mesh).

PardhavMaradani:
Done. This is how the options look now:

line-annotation-params

tests/test_annotations.py
```diff
@@ -341,12 +339,62 @@ def test_trajectory_annotation_canonical_dihedrals(self, universe, session):
...
+                [10, 10, 10, 90, 90, 90],  # cubic
+                [10, 10, 10, 60, 60, 90],  # rhombic dodecahedron xy-square
+                [10, 10, 10, 60, 60, 60],  # rhombic dodecahedron xy-hexagon
+                [10, 10, 10, 70.53, 109.47, 70.53],  # truncated octahedron
+                [10, 10, 20, 90, 90, 60],  # hexagonal cell
+            ]
+        )
+        # add mda transform to set different box vectors for different frames
+        transform = boxdimensions.set_dimensions(box_dimensions)
+        universe.trajectory.add_transformations(transform)
```
BradyAJohnston: This is adding a transform to the universe which is a shared fixture between the tests, potentially changing results in tests run after this. For this specific test where you are chaning the universe you'll want to create a deep copy initially before creating the trajectory / universe.

molecularnodes/annotations/interface.py
```diff
@@ -15,6 +16,7 @@ class AnnotationInterface:
     # py annotations of common params for code complete
     # these are bound dynamically along with any annotation inputs
     visible: bool
+    text_font: str
```
BradyAJohnston, line 19: this is duplicated from two lines down

molecularnodes/annotations/manager.py
```diff
@@ -518,20 +568,206 @@ def _draw_annotations(self):
...
+                    attr_material_index.append(object_material_index)
+                    attr_shade_smooth.append(object_shade_smooth)
+                bm.edges.index_update()
+                # free bmesh
+                bm_to_add.free()
+
+        # update the object mesh
+        bm.to_mesh(self.bob.object.data)
+        # free bmesh
+        bm.free()
```
BradyAJohnston: We should maybe wrap this in a `try: except:` to free the bmesh if something goes wrong inside the creation

molecularnodes/annotations/node_tree.py
```diff
@@ -0,0 +1,168 @@
...
+    # curve_to_mesh.Mesh -> join_geometry.Geometry
+    node_group.links.new(
+        curve_to_mesh.outputs["Mesh"], join_geometry.inputs["Geometry"]
+    )
+    # group_input.Geometry -> join_geometry.Geometry
+    node_group.links.new(
+        group_input.outputs["Geometry"], join_geometry.inputs["Geometry"]
+    )
+    # curve_to_mesh.Mesh -> join_geometry.Geometry
+    node_group.links.new(
```
BradyAJohnston: this is a duplicate of L:156-159

molecularnodes/annotations/props.py
```diff
@@ -123,8 +123,18 @@ def setter(self, value):
             # set a non blender property and validate
             setattr(instance, nbattr, value)
             instance.validate()
+            if instance._ready:
```
BradyAJohnston, line 126:
```suggestion
            if hasattr(instance, '_ready') and instance._ready:
```

molecularnodes/annotations/props.py
```diff
@@ -134,6 +144,11 @@ def setter(self, value):
     return property(getter, setter, doc=getattr(_prop(), "description", None))
 
 
+def _update_annotation_object(self, context):
+    entity = context.scene.MNSession.get(self.id_data.uuid)
+    entity.annotations._update_annotation_object()
```
BradyAJohnston:
```suggestion
entity = context.scene.MNSession.get(self.id_data.uuid)
if entity is not None and hasattr(entity, 'annotations') and entity.annotations is not None:
       entity.annotations._update_annotation_object()
```

molecularnodes/entities/trajectory/annotations.py
```diff
@@ -389,6 +390,70 @@ def draw(self) -> None:
...
+                    params.center_to_origin,
+                    params.show_lattice,
+                )
+            else:
+                # Regular Triclinic cell
+                a, b, c, alpha, beta, gamma = ts.dimensions
+                origin = Vector((0, 0, 0))
+                # move box center to origin if set
+                if params.center_to_origin:
+                    origin = -1 * np.sum(ts.triclinic_dimensions, axis=0) / 2
```
BradyAJohnston: `draw_triclinic_cell` expects a `Vector` and `origin` is a `Vector so we can just not involve numpy here: 

```suggestion
                    origin = Vector(-sum(ts.triclinic_dimensions, axis=0) / 2)
```

molecularnodes/annotations/base.py
```diff
@@ -536,13 +548,500 @@ def draw_circle_3d(
...
+            depth=height,
+        )
+        base_offset = Matrix.Translation(Vector((0, 0, height / 2)))
+        up = Vector((0, 0, 1))
+        rot = up.rotation_difference(axis).to_matrix().to_4x4()
+        loc = Matrix.Translation(Vector(location) * self._world_scale)
+        mat = loc @ rot @ base_offset
+        bm.transform(mat)
+        self.draw_bmesh(bm, overrides=overrides)
+        bm.free()
```
BradyAJohnston: My comments about `bmesh` might be wrong because I've never used it but similarly to other file we might want to have bmesh usage in a `try: except:` to ensure we release it properly if something goes wrong?

tests/test_annotations.py
PardhavMaradani: Fixed by making a copy of the universe

molecularnodes/annotations/interface.py
PardhavMaradani: Fixed

molecularnodes/annotations/manager.py
PardhavMaradani: Instead of a try/except, I added a re-usable context manager that ensures a newly created bmesh is always freed.

molecularnodes/annotations/node_tree.py
PardhavMaradani: Removed the duplicate code

molecularnodes/annotations/props.py
PardhavMaradani: The `_ready` attribute is created right after the instance is created. So, this check is redundant. Did you run into any issue that prompted this check? If so, I could debug that, because that shouldn't happen at all (unless of course some one intentionally deleted that attribute)

molecularnodes/annotations/props.py
PardhavMaradani: The entity is not None check is fine. I added that. This code is in the annotation properties update callback. If the annotations manager doesn't exist for the entity for whatever reason (this shouldn't happen in standard use), I would like to see an exception so we have better debuggability than silently not update the annotation object. If you suggested this because you ran into an issue, I would like to get to the bottom of why it happened in the first place.

molecularnodes/entities/trajectory/annotations.py
PardhavMaradani: The above code as is will not work because there is no axis keyword param in the regular sum method. I removed that and made the change removing numpy dependency.

molecularnodes/annotations/base.py
PardhavMaradani: The new `new_bmesh` context manager should address this as mentioned earlier.

BradyAJohnston (approved):

Issue #985, opened 2025-09-26
ENH: support MDAnalysis Universe without resnames


tylerjereddy:
@BradyAJohnston this is in the context of what we were discussing with @orbeckst earlier today re: rendering of SMILES-based MDAnalysis `Universe` objects. `MolecularNodes`  `4.5.2` currently seems to require `resname` information to create a `mn.Trajectory` object, but of course if you start from a small/raw SMILES string, you may very well not have residue names available, and I don't think MDAnalysis has a residue name guesser at the time of writing (and I'm not sure how one would sensibly work anyway).

I wonder if we could support this somehow? Presumably there are a number of compelling uses cases in the small molecule rendering space. I'll try to think of a sneaky workaround for now, but I'll also place a fairly minimal viable reproducer below. Perhaps there is an `RDKit` argument I can pass through in the interim or something like that, via `rdkit_kwargs`, and if so it may make sense for us to concisely document that somewhere for users if a direct "fix"/enhancement would be painful.

```python
import MDAnalysis as mda
import molecularnodes as mn


ethanol_smiles = "CCO"
u = mda.Universe.from_smiles(ethanol_smiles)
canvas = mn.Canvas(
    mn.scene.Cycles(samples=64), transparent=True, resolution=(1000, 1000)
)
traj = mn.Trajectory(u)
traj.add_style(
    mn.StyleSpheres(), selection=f"all"
)
canvas.frame_view(traj.get_view("all"))
canvas.snapshot("ethanol.png")
```

<details>
<summary> Traceback Observed with MN 4.5.2 </summary>

```
Traceback (most recent call last):
  File "/Users/treddy/github_projects/MDAnalysis-paper-2022/figure_work/repro.py", line 10, in <module>
    traj = mn.Trajectory(u)
           ^^^^^^^^^^^^^^^^
  File "/Users/treddy/python_venvs/py_311_MN/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 50, in __init__
    self.create_object(name=name)
  File "/Users/treddy/python_venvs/py_311_MN/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 505, in create_object
    self._create_object(name=name)
  File "/Users/treddy/python_venvs/py_311_MN/lib/python3.11/site-packages/molecularnodes/entities/trajectory/base.py", line 484, in _create_object
    for att_name, att in self._attributes_2_blender.items():
[...]

PR #992, opened 2025-10-10
Cleanup Trajectory attribute mapping

BradyAJohnston:
Refactors how attributes are computed and stored on the entity. Removes top-level availability of attributes to reduce confusion. For universe attributes one must use `traj.atoms.resnames` and for attributes store on the mesh `tray.named_attribute("res_name")` or `traj["res_name"]`.

Solves the issues raised in #985 about parsing a SMILES. Previously it was getting an error when missing certain attributes but now they just aren't added if not present and keeps moving. 

Example for a SMILES rendering which now works:

```py
import MDAnalysis as mda
import molecularnodes as mn

ethanol_smiles = "CCO"
u = mda.Universe.from_smiles(ethanol_smiles)
canvas = mn.Canvas(
    mn.scene.Cycles(samples=64), transparent=True, resolution=(1000, 1000)
)
traj = mn.Trajectory(u)
traj.add_style(
    mn.StyleSpheres(), selection=f"all"
)
canvas.frame_view(traj.get_view("all"))
canvas.snapshot("ethanol.png")
```


BradyAJohnston:
No reason just missed it - thanks for testing and flagging. Have now added!

PR #999, opened 2025-10-17
Fix dna tests

PardhavMaradani:
PR to verify dna tests

### DIFF
--- a/tests/test_trajectory.py
+++ b/tests/test_trajectory.py
@@ -321,7 +321,13 @@ class TestTrajectory:
 
     def test_get_view(self, universe):
         t1 = mn.Trajectory(universe)
-        # view of resid 1
+        # Note: When a frame is not specified, whatever is the current
+        # universe frame should be used. In some random test failures
+        # with bpy, the frame value was 3 for both cases leading to
+        # an assertion failure. It is very likely a pytest issue.
+        # Explicitly start with the frame reset to 0 for the universe
+        universe.trajectory[0]
+        # view of resid 1 at trajectory frame 0
         v1 = t1.get_view(selection="resid 1")
         # view of resid 1 at trajectory frame 3
         v2 = t1.get_view(selection="resid 1", frame=3)


PardhavMaradani:
The dna test failures and the the martini failures started after PR #992. That PR removed the attribute types, which is a source of the problem for the dna test failures. This PR only attempts to fix the tests and the snapshots and not modify databpy or the code - maybe there needs to be a better fix.

PardhavMaradani:
Hi Brady, I will leave this PR in a draft state as it is not intended for a merge. I see you already have PR #1000 to attempt fixing these test failures the right way. You can close this once they are resolved. This draft only tries to show the reasons for the test failures by fixing the tests and passes all the current tests. Here is quick summary of the issues:

* Windows - test_dna failures:
  * As mentioned previously, this started after PR #992 where explicit types for attributes was removed
  * Saving the attribute using its `str` representation in the snapshot, saves the dtype that is platform specific (`int64` in snapshot vs `int32` as read from attributes in Windows)
  * `databpy` uses an `int` for all blender integer type [attributes](https://docs.blender.org/manual/en/latest/modeling/geometry_nodes/attributes_reference.html#attribute-data-types) and doesn't distinguish between (`Integer`, which is 32-bit, `Byte Color` and `8-Bit Integer`, which are 8-bit and other 16 and 32 bit integer vector types) - not necessarily an immediate issue
* Windows - test_martini failures:
  * The attribute type issue mentioned above applies here too
  * The order of the attributes read from Blender object's `data.attributes` matter for snapshot capture and checking. (need this to be ordered like `list_attributes` in `databpy`) Explicit ordering for anything that goes into snapshots to avoid future issues too
* Intermittent trajectory test_get_view failures:
  * This is a completely random test failure that I think can only be a `pytest` issue
  * Only shows up in `bpy` in both macos and ubuntu, only at times - even in the commits for this PR, the earlier ones don't show it, one of them shows it for macos and an empty commit later, it shows up for ubuntu.
  * The test is extremely simple and doesn't use anything in Blender - no frame change, no ops etc. It only uses an `AtomGroup`'s `bbox` method and the universe trajectory iteration to set the frame. It seems like an already iterated trajectory (at frame 3) is what gets used in the failure cases because the universe is a fixture. The fix for us is to explicitly set the frame (or any other stuff in the future) for our tests and not rely on an assumed state of the universe
* Local `test_dna` failures on ubuntu weirdness
  * This is one of the issues that I can consistently reproduce on my ubuntu
  * `test_dna` run passes the first time and any subsequent run fails with errors like `test_univ_snapshot - KeyError: 'bpy_prop_collection[key]: key "chain_id" not found'` (A version of the same error showed up on a GHA macos [test failure](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/18561675097/job/52912130677?pr=983) too)
  * Just `touching` the `tests/test_dna.py` file and re-running the test works the first time and fails subsequently
  * I could narrow it down to not including the import for `NumpySnapshotExtension` in that file - so something with snapshots and pytest that causes this. Anyway, the import fixes this issue

BradyAJohnston:
Thanks for looking into this so much. Was able to fix and get all of the tests working in #1000

PR #1000, opened 2025-10-17
Fix: Explicit attribute types

BradyAJohnston:
Uses `databpy>=0.4.1` to be more explicit with attribute return types. This affects some snapshots but should stabilise tests across platforms and have better consistency.

Thanks to the additional investigations of test failures from @tylerjereddy and @PardhavMaradani additionally in #999


BradyAJohnston:
Thanks for looking into this @tylerjereddy . You are right that Blender / Geometry Nodes only actually supports `Int32` and not `Int64`. I need to make the changes in `databpy` to be more specific about return types rather than just `int` / `float` to make sure it's matching with what is happening inside of MN. 

This will change some of the current snapshots / tests in MN but those will just have to be updated to be how they "should be". 

As for your point about snapshots - I agree that they are a bit of a pain. They are the easiest way for taking saving what the output of the geometry node tree evaluation should be. I think a lot of their usage could be reduced to only when absolutely needed.

PR #1017, opened 2025-11-07
Suppress MDAnalysis user warnings in tests

PardhavMaradani:
Hi Brady, while looking at an mmcif parsing issue in the MDAnalysis repo recently, I learnt how the user warnings were being suppressed in the tests. I remember you wanting to suppress the over 8k warnings in MN tests too at some point. If you still do, this PR has the required changes. If not, you can close this. Not sure if you wanted to suppress the deprecation/pending deprecation warnings too - a couple of tests currently throw those warnings and this PR ignores them too. Thanks

### DIFF
--- a/tests/test_trajectory.py
+++ b/tests/test_trajectory.py
@@ -45,7 +45,7 @@ class TestTrajectory:
 
     def test_include_bonds(self, universe_with_bonds):
         traj = mn.entities.Trajectory(universe_with_bonds)
-        assert traj.edges.items() != []
+        assert len(traj.data.edges) == 28015
 
     def test_attributes_added(self, universe):
         traj = mn.entities.Trajectory(universe)


BradyAJohnston:
Thanks for looking into this. Some of the warnings we might want to try and address through changing code to remove the warnings (like with some of the deprecations). Unsure if we should just be more selective with suppressing different warnings if we no they won't be an issue.

PardhavMaradani:
Just documenting the current list of warnings here. Unless these are a real issue, maybe we could leave them as is till they are addressed and close this PR.

```
1 warning
tests/test_load.py::test_pdb_no_bonds
biotite/structure/io/pdb/file.py:475: UserWarning: 16 elements were guessed from atom name

1 warning
tests/test_session.py::test_add_trajectory
MDAnalysis/core/groups.py:3536: UserWarning: Empty string to select atoms, empty group returned.

18 warnings
tests/test_dna.py::TestOXDNAReading
MDAnalysis/core/universe.py:1916: UserWarning: there is no reference attributes in this universe to guess types from

1 warning
tests/test_trajectory.py::TestTrajectory::test_include_bonds
DeprecationWarning: BlenderObject.edges is deprecated and will be removed in version 1.0.0. Use bob.data.edges directly for mesh objects.

2 warnings
tests/test_trajectory.py::TestTrajectory::test_selection_from_atomgroup
tests/test_trajectory.py::TestTrajectory::test_add_style
MDAnalysis/core/groups.py:3536: UserWarning: Empty string to select atoms, empty group returned.

8040 warnings (this is the bulk of them)
tests/test_trajectory.py:
MDAnalysis/guesser/default_guesser.py:259: PendingDeprecationWarning: Unknown masses are set to 0.0 for current version, this will be deprecated in version 3.0.0 and replaced by Masse's no_value_label (np.nan)
```

PardhavMaradani:
The latest commit adds more selective filtering of warnings for which there is nothing much we can do in MN.

BradyAJohnston:
Thanks for updating the filters to be more specific, this way we can still see other warnings that pop up and address them (or filter them) as needed. I've just updated the code to address the warning about databpy edges.

PardhavMaradani:
Hi Brady, just fyi that the docs (at https://bradyajohnston.github.io/MolecularNodes/) didn't seem to get updated with the v4.5.5 release. Thanks

BradyAJohnston:
Yep there are some issues with the docs deployment I need to do it manually currently - thanks for catching!

PR #1021, opened 2025-11-11
Fix #1019 selection creation

### LINKED ISSUE #1019: MDA Selections added via GUI duplicate names
Currently every selection added via the <kbd>+</kbd> icon in the GUI result in selections of the same name. Selections when adding via the viewport "add style" aren't duplicated (and are also named differently).

Image

BradyAJohnston:
Fixes selection names on creation via the GUI.


BradyAJohnston:
Better fix in #1023

Issue #989, opened 2025-10-04
Change of Alpha Over inputs


BradyAJohnston:
The input names for **Alpha Over** node have changed: https://projects.blender.org/blender/blender/pulls/146352

Need to change our usage if we reference them directly.

PR #1028, opened 2025-11-22
Fix compositor setup for 5.0

### LINKED ISSUE #988: Change usage of "Fac" -> "Factor"
For 5.0 all of the sockets that previously used abbreviations are being changed to use full words such as "Factor". 

https://projects.blender.org/blender/blender/pulls/140498

So far I've only noticed 1 place where `"Fac"` is used during compositor setup.

BradyAJohnston:
Fixes #988, #989 for compositor creation in 5.0

Ping @PardhavMaradani as it won't let me assign you as reviewer

| Main | PR |
|--------|--------|
| CleanShot 2025-11-22 at 11 49 32 |CleanShot 2025-11-22 at 11 53 59 |

### DIFF
--- a/molecularnodes/blender/__init__.py
+++ b/molecularnodes/blender/__init__.py
@@ -1,4 +1,4 @@
 from . import coll, mesh
-from .utils import path_resolve, set_obj_active
+from .utils import IS_BLENDER_5, path_resolve, set_obj_active
 
-__all__ = ["coll", "mesh", "nodes", "path_resolve", "set_obj_active"]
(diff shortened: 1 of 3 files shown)


PardhavMaradani:
The changes look good to me. Thanks

PardhavMaradani:
Hi Brady, just fyi that there are a few other places (mostly in `molecularnodes/scene` and canvas tests) where we have the `if bpy.app.version >= (5, 0, 0)` checks, which could potentially be changed to use the macro you added. It looks like the changes you have here were added to Blender after we made to those initial changes to support 5.x. Thanks

BradyAJohnston:
Yeah I did think about that after merging - will go through and update.

Issue #1026, opened 2025-11-14
Replicating trajectories over periodic boundaries


jpkrowe:
This feature request came about from a conversation @BradyAJohnston and I had at the MDAnalysis UGM.

In VMD/Ovito etc. you can replicate md trajectories over periodic boundaries to create supercells. This is nice if you want to visualise the larger periodic system and for simulations of crystals/solid state materials.

In VMD/Ovito, the user specifies a grid with how many times the system should be duplicated in each direction.

So with input: 
<img width="300" height="123" alt="Image" src="https://github.com/user-attachments/assets/c27f5ca8-7deb-439c-aafd-3943c8eaafe3" />

This structure:

<img width="249" height="209" alt="Image" src="https://github.com/user-attachments/assets/92186eb1-da25-4c5b-99cc-6d24fa2634fc" />

becomes:
<img width="200" height="250" alt="Image" src="https://github.com/user-attachments/assets/b3d79662-a09e-45f1-94ba-f0051e1a5bc8" />

To generate the coordinates of a single image of the system we need:
```python
shift = np.array([nx, ny, nz])
```
where `nx`, `ny` and `nz` are integers representing how many boxes away the image is from the initial coordinates.

The coordinates can then be generated using (I believe):

```python
new_coords = ts.positions + shift@ts.triclinic_dimensions
```
To get the grid like in vmd/ovito we just need to perform a nested loop over all shift combinations. This is very similar to the tiling a new universe tutorial here: https://userguide.mdanalysis.org/stable/examples/constructing_universe.html

I think it would be good to have features where the user can specify a grid like in other programs. I would also personally find it quite useful to have a bit more control and be able to specify specific shifts - like I could specify (5, -1, 2) and have that specific image appear.

BradyAJohnston:
Thanks for the chat at the UGM, and for kicking off the discussion around this.

Inside of Geometry Nodes, it should be as simple as "Generate Points" -> "Instance on Points". When we are instancing we get the additional performance benefits that only a single copy of the system exists on the GPU and the GPU handles copying & pasting for the other copies. You can't get individual control over instances (by virtue of them being copies of the starting geometry) but we can **Realize Instances** to get individual control. This isn't very problematic with Point Clouds but would be problematic with other heavier styles / geometry.


## Instances

<img width="2032" height="1168" alt="Image" src="https://github.com/user-attachments/assets/a6f7773a-b09a-4f13-bc6c-f27f1c17cbb2" />

[...]

PR #1035, opened 2025-11-24
Adding tools for replicating MD trajectories across periodic boundaries

jpkrowe:
This PR attempts to address Issue #1026. It adds 4 new nodes:
- Periodic Box - this node takes as input the a, b, c lengths and alpha, beta, gamma angles to define a simulation box and outputs the corresponding vectors. This node updates with the trajectory to allow for boxes that change during the simulation. The parameters can also be entered manually for use with non-MD inputs e.g. pdb structures
- Lattice Grid - this node takes the box vectors as input and provides a grid of lattice points. By inputting these points into a instance on points node the system can be replicated at each of these points.
- Periodic Image - this node provides the location of a single lattice point to allow the duplication of the system at a specific point/set of points
- Replicate Trajectory - this node bundles up a Periodic Box, Lattice Grid and Instance on Points node to allow for replication of a trajectory across a grid with one node.


BradyAJohnston:
Don't worry about the current tests failing - it's attempting to add the new nodes from the current `.blend` but they haven't been moved over yet. I'll fix this shortly.

BradyAJohnston:
I've done a little bit of cleanup (haven't pushed it yet): 
CleanShot 2025-11-25 at 09 52 28

I think we want to change the name for "Replicate Trajectory" - as this could also be re-used on crystal structures and the like. Maybe **Instance Periodic** or something similar?

BradyAJohnston:
Additionally I'd love to have some better organisation and annotations inside of **Perioidic Box** node. You can add a frame by selecting some nodes and pressing `F` and then if you use the _Node Arrange_ add-on (https://extensions.blender.org/add-ons/node-arrange/) it should do a nice job of nicely arranging everything and look something like this: 

CleanShot 2025-11-25 at 09 58 21

That's just with me adding some random annotations but I want to try and keep things more organised in node trees (which isn't an easy task).

jpkrowe:
> I've done a little bit of cleanup (haven't pushed it yet): CleanShot 2025-11-25 at 09 52 28
> 
> I think we want to change the name for "Replicate Trajectory" - as this could also be re-used on crystal structures and the like. Maybe **Instance Periodic** or something similar?

I like the changes you have suggested. I think the inputs to "Periodic Image" should be A, B, C to match with "Lattice Grid" and the outputs of "Periodic Box". I think we may also want to use Image A, Image B etc instead of Image X, Image Y. Maybe "Periodic Array" is a better name for the "Replicate Trajectory" node. Do you want me to implement these changes or is it easier if you do it?

BradyAJohnston:
If you want to implement these changes along with tidying up the internals of the `Periodic Box` node that would be great! Then I'll merge it all.

jpkrowe:
I have cleaned up the `Periodic Box` node:

image

and I have implemented those changes to the labels and names:

image

BradyAJohnston:
I'm happy with how this all works and runs. I'm going to hold off on merging as I would like to have some tests for these nodes first. There have been some updates to testing / capturing the outputs of geometry nodes computations that should improve our tests so once I get some test / utility functions for tests I'll add the tests and then we should be good to go!

jpkrowe:
Cool. Sounds good to me!

BradyAJohnston:
Okay I've added some initial tests and also did a bit of a refactor of the update code (to fix a bug and also code style). 

Changes were: 

- break / continue early instead of nesting tabs
- iterate names / values - DRY
- check for a node's node_tree rather than just name / label (was an actual bug)
- assign dimensions / names at top level rather than in each loop

BradyAJohnston (approved):

PR #1036, opened 2025-11-25
Fix: Additional checks in render_pre_handler

BradyAJohnston:
Doesn't really fix issues with crashing during rendering but will reduce the number of users who encounter it. We return early if there aren't any entities or annotations on those entities so unless a user is using annotations directly they shouldn't encounter any issues.

### DIFF
--- a/molecularnodes/annotations/utils.py
+++ b/molecularnodes/annotations/utils.py
@@ -1,10 +1,13 @@
 import inspect
 import types
 import typing
+from typing import TYPE_CHECKING
 import bpy
 from PIL import Image
-from ..entities.trajectory import TrajectoryAnnotationManager
-from ..session import MNSession
+
+if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


PardhavMaradani:
This looks fine to me. I don't think we have documented anywhere on how to setup the compositor when using the regular Blender GUI - the most common workflow. (This would be needed to render annotations for anyone that uses them or in the future to support background colors, etc) For the background mode, this is setup during `Canvas` creation (which also ensures the `Lock Interface` is set). I am not sure where this documentation would go, but that could be an additional step that would help if those code paths hit. Thanks

BradyAJohnston:
Certainly having some written documentation on using the GUI for annotations is needed. I'm going to merge this for now to get the 'corrective' fix out but the majority of users aren't using annotations currently so making some docs is something we need to do going forward.

PR #1037, opened 2025-11-25
Use explicit is None check of entity in UI panel

PardhavMaradani:
For a density entity, the UI panel throws the following error and fails to draw the entire panel:

```
00:50.032  bpy.rna          | ERROR Python script error in MN_PT_Entities.draw
  File "/home/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/ui/panel.py", line 652, in draw
    if not entity:
  File "/home/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/databpy/object.py", line 485, in __len__
    raise TypeError(
TypeError: Object type Volume is not supported. Supported types: Mesh, Curves, PointCloud
```

It looks like the new `__len__` method of `databpy` is getting called for this check which wasn't the case before. Changing this to an explicit is None check as it should have been in the first place. Thanks

### DIFF
--- a/molecularnodes/nodes/style_density_iso_surface.py
+++ b/molecularnodes/nodes/style_density_iso_surface.py
@@ -426,10 +426,12 @@ def style_density_iso_surface_node_group():
     # node Only Contours Switch
     only_contours_switch = style_density_iso_surface.nodes.new("GeometryNodeSwitch")
     only_contours_switch.name = "Only Contours Switch"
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
Accidentally merged the other PR first #1038 which included the changes from here so now this PR is equal so I will close (whoops!)

Issue #1039, opened 2025-11-25
Unable to create annotations for Density entity


PardhavMaradani:
I get the following exception when trying to create a new annotation for a Density entity:

```
01:08.873  bpy.rna          | ERROR Python script error in MN_OT_add_annotation.execute
01:08.874  operator         | ERROR Python: Traceback (most recent call last):
                            |   File "/home/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/ui/ops.py", line 1003, in execute
                            |     MN_OT_Add_Annotation._temp_classes = _register_temp_annotation_add_op(entity)
                            |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            |   File "/home/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/ui/ops.py", line 925, in _register_temp_annotation_add_op
                            |     entity_annotation_type = f"{entity._mn_entity_type}_{cls.annotation_type}"
                            |                                 ^^^^^^^^^^^^^^^^^^^^^^
                            | AttributeError: 'Grids' object has no attribute '_mn_entity_type'. Did you mean: '_entity_type'?

```

This worked in the past and it looks like there is now an `_mn_entity_type` and also an `_entity_type`. I only looked briefly and couldn't figure out the change, hence filing an issue. Thanks

BradyAJohnston:
I was changing up some property access but I seem to have not captured everything.

This diff seems to fix the issue:

```
diff --git a/molecularnodes/ui/ops.py b/molecularnodes/ui/ops.py
index fdc6c4a7..425a04b5 100644
--- a/molecularnodes/ui/ops.py
+++ b/molecularnodes/ui/ops.py
@@ -922,7 +922,7 @@ def _register_temp_annotation_add_op(entity):
     for cls in entity.annotations._classes.values():
         AnnotationTypeInputs = create_annotation_type_inputs(cls)
         register(AnnotationTypeInputs)
-        entity_annotation_type = f"{entity._mn_entity_type}_{cls.annotation_type}"
+        entity_annotation_type = f"{entity._entity_type}_{cls.annotation_type}"
         attributes["__annotations__"][entity_annotation_type] = (
             bpy.props.PointerProperty(type=AnnotationTypeInputs)
         )
[...]

PR #1040, opened 2025-11-26
Fix annotations

PardhavMaradani:
* Use the correct entity type string for accessing annotation inputs
* Support annotations for derived entities like OXDNA and StreamingTrajectory
* Add new tests to add annotation (from UI) for Molecule and Density EntityType

---

Hi Brady, this should fix the issue described in #1039 and a few other related issues.

I have added a couple of new tests (earlier, only add op for Trajectory was in tests) to test for the UI add operator for Molecule and Density entity as well - this would have caught the issue earlier.

Both `OXDNA` and `StreamingTrajectory` that derive from `Trajectory` entity should now be able to re-use the annotations for Trajectory. I have not added `OXDNA` in the poll because the annotations aren't of much use - the topology doesn't have the required attributes like masses (for center-of-mass calculation), names (for selections based on names) etc. I haven't gotten a chance to read about and setup Streaming Trajectories yet. It would be great if you can give this a try if you already have a setup. It would be pretty cool to see if the annotations like com distance etc update in real time. I only added a basic `n_frames` check in the `universe_info` annotation to not error out, but I believe the rest should work as is, though I couldn't verify. Thanks

### DIFF
--- a/molecularnodes/entities/base.py
+++ b/molecularnodes/entities/base.py
@@ -35,6 +35,7 @@ class MolecularEntity(
     metaclass=ABCMeta,
 ):
     update_with_scene = BoolObjectMNProperty("update_with_scene")
+    _entity_type: EntityType = None  # Overridden by derived classes
 
     def __init__(self) -> None:
         super().__init__(obj=None)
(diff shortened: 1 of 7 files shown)


PardhavMaradani:
The Build Extension test failure seems to be a setup issue:

```
/dev/disk6          	GUID_partition_scheme          	
/dev/disk6s1        	Apple_HFS                      	/Volumes/Blender
hdiutil: couldn't eject "disk6" - Resource busy
Error: Process completed with exit code 16.
```

BradyAJohnston (approved):

BradyAJohnston:
Tested with the streaming trajectories and it all seems to work well!

PR #1043, opened 2025-11-27
Fix universe_info annotation for streaming trajectories

PardhavMaradani:
This PR fixes the universe_info annotation that now works with both regular and streaming trajectories.

---

From the API, had to use `mn.entities.StreamingTrajectory` because only `Molecule` and `Trajectory` entities are available in the top level `__init__.py`. All other annotations work fine. Thanks

streaming-trajectory-universeinfo

### DIFF
--- a/molecularnodes/entities/trajectory/annotations.py
+++ b/molecularnodes/entities/trajectory/annotations.py
@@ -1,6 +1,6 @@
-import os
 import typing
 from pathlib import Path
+from typing import TYPE_CHECKING
 from mathutils import Vector
 from MDAnalysis.core.groups import AtomGroup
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
Thanks for looking into this - I had tested the regular annotations but not the universe info stuff. This works well. I haven't moved `StreamingTrajectory` towards the top level yet because it's still quite experimental and I think it should ultimately just be part of the `Trajectory` class with special handling but that is something to deal with later.

CleanShot 2025-11-27 at 08 38 25

BradyAJohnston:
I just pushed some slight tweaks to the string handling for it.

BradyAJohnston (approved):

PR #1047, opened 2025-11-28
Style swap and selection from menu

BradyAJohnston:
Still need to do a general cleanup pass for UI code but I thought I should put up for initial feedback.

To better improve the menu UX this makes the following changes:
* Drops the extra `layout.box()` for the style options to improve UI display
* Expose the selection string if a selection is being used along with error reporting and feedback
* Add operator and menu for swapping a style node to other supported styles (don't have to remove and re-add)

https://github.com/user-attachments/assets/f94e046b-a05e-46d1-91d0-1fe35a443cc7

This does highlight I think the importance that we try and unify the backend to be an `mda.Universe` for molecules as well. It would reduce development burden and also allow someone to utilize the features such as selection strings.

Points to fix up: 

- [x] remove box from annotations
- [x] style label editing
- [ ] style order
- [x] lacking join geometry error
- [x] selection string
- [ ] ~~color inputs~~ Will handle more generally later
- [ ] ~~API changes~~
- [ ] ~~dropdown animateable~~ not really possible / relevant
- [x] handle `Molecule` entity
- [x] handle density

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "PyYAML",
     "imdclient>=0.2.3",
     "ipython>=9.7.0",
+    "pandas<3.0.0",
 ]
 maintainers = [{ name = "Brady Johnston", email = "brady.johnston@me.com" }]
 requires-python = "~=3.11.0"
--- a/uv.lock
+++ b/uv.lock
@@ -544,6 +544,7 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/a4/de/f28ced0a67749cac23fecb02b694f6473f47686dff6afaa211d186e2ef9c/greenlet-3.2.4-cp311-cp311-macosx_11_0_universal2.whl", hash = "sha256:96378df1de302bc38e99c3a9aa311967b7dc80ced1dcc6f171e99842987882a2", size = 272305, upload-time = "2025-08-07T13:15:41.288Z" },
     { url = "https://files.pythonhosted.org/packages/09/16/2c3792cba130000bf2a31c5272999113f4764fd9d874fb257ff588ac779a/greenlet-3.2.4-cp311-cp311-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:1ee8fae0519a337f2329cb78bd7a8e128ec0f881073d43f023c7b8d4831d5246", size = 632472, upload-time = "2025-08-07T13:42:55.044Z" },
     { url = "https://files.pythonhosted.org/packages/ae/8f/95d48d7e3d433e6dae5b1682e4292242a53f22df82e6d3dda81b1701a960/greenlet-3.2.4-cp311-cp311-manylinux2014_ppc64le.manylinux_2_17_ppc64le.whl", hash = "sha256:94abf90142c2a18151632371140b3dba4dee031633fe614cb592dbb6c9e17bc3", size = 644646, upload-time = "2025-08-07T13:45:26.523Z" },
+    { url = "https://files.pythonhosted.org/packages/d5/5e/405965351aef8c76b8ef7ad370e5da58d57ef6068df197548b015464001a/greenlet-3.2.4-cp311-cp311-manylinux2014_s390x.manylinux_2_17_s390x.whl", hash = "sha256:4d1378601b85e2e5171b99be8d2dc85f594c79967599328f95c1dc1a40f1c633", size = 640519, upload-time = "2025-08-07T13:53:13.928Z" },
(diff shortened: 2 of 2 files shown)


PardhavMaradani:
I will try this out and update with my comments before the end of this weekend. Thanks

PardhavMaradani:
Hi Brady, I tried the branch out and this looks good.

Here are some comments in no particular order:

1. If we are removing the box layout, we should remove it from annotations as well to keep the overall tab consistent
2. The added style name is no longer editable (by double clicking on the name and editing inline - see entity name and annotation name as other examples where this works). This worked in the past - I believe this is due to changes in PR #1044. Though styles have no name based lookups today (unlike annotations), it is still useful to be able to rename this for better organization - especially when there are multiple styles. We use the node label for this purpose to avoid changing the node name as it is required to determine if this is a style node elsewhere in the code
3. There is a bug in the style order when there are two styles - they get flipped. 
  To reproduce:
    * Delete all styles
    * Add style `Spheres` with default options
    * Next, add style `Cartoon` with default options
    * Style `Cartoon` now shows up as the first one
  Since styles are accessed based on an index from the API, this can be a problem. (I don't think this bug is due to changes in this PR though)
4. We need to include a `Join Geometry` node in the initial node tree that gets created (maybe in the `nodes.new_tree` method).
  Here is how to reproduce a bug due to this:
    * Add a new `MD` entity with the `Style` option unchecked (i.e., no initial style)
    * Open the n-panel in the 3D viewport - The Styles panel will not show up due to the following errors:

```
...molecularnodes/nodes/nodes.py", line 741, in final_join
    current = node_previous(current)
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/nodes/nodes.py", line 726, in node_previous
    return node.inputs[0].links[0].from_socket.node
           ~~~~~~~~~~~^^^
IndexError: bpy_prop_collection[index]: index 0 out of range, size 0

...molecularnodes/ui/panel.py", line 835, in draw
    style_nodes = get_final_style_nodes(node_group)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/nodes/geometry.py", line 160, in get_final_style_nodes
    links: bpy.types.NodeLinks = final_join(tree).inputs[0].links  # type: ignore
                                 ^^^^^^^^^^^^^^^^
  File "/tmp/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/nodes/nodes.py", line 746, in final_join
    return insert_join_last(tree)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pardhav/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/nodes/nodes.py", line 704, in insert_join_last
    node_output.location += Vector([NODE_SPACING * 2, 0])
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: Writing to ID classes in this context is not allowed: MN_NewTrajectory, NodeTree data-block, error setting NodeGroupOutput.location
```

5. The `Selection` string property only shows up when the style is initially created with a selection. When the style is created from the UI (using the `+` add style button), the default selection is empty, so it doesn't show up after creation and kind of defeats the purpose. We should probably make sure the defaults are set to ensure the selection node (and hence the input property) is always there
6. The only other inputs (from the add new style popup) that isn't covered is the color selection - not sure if you intended to add this or not. I know that these nodes are only created based on the inputs (like the selection case above), but there is definitely value in having those configurable. Anything that is part of the inputs should be ideally configurable later too. A visibility control (similar to the one in entities list and annotations list) would also be really nice to have
7. The new selection and style swap are not available through the API - maybe you intended to add them later? (if not, this would be the first break between API and GUI compatibility)
8. Just an fyi that the style swap dropdown is currently not animatable
9. This panel errors out for the `Molecule` entity because the selections for it are entirely different from those of `MD`

```
  File "/tmp/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/ui/panel.py", line 871, in draw
    panel_selection_node(box, style_node, entity)
  File "/tmp/blender/blender-5.0.0-linux-x64/5.0/python/lib/python3.11/site-packages/molecularnodes/ui/panel.py", line 792, in panel_selection_node
    item: TrajectorySelectionItem = entity.selections.ui_items[attr_name]
                                    ^^^^^^^^^^^^^^^^^
AttributeError: 'Molecule' object has no attribute 'selections'
```
10. There is a single style (picked at the time of creation) for the `Density` entity. Either we don't show the style swap options (as none of them will work) or show the two supported density styles in that list

BradyAJohnston:
Thanks for the review - I think I'll do the panel related changes in a separate smaller PR. The same with bringing back editable style names - it would be useful to be able to name them.

I think we should maybe default to the selection string being `all` for a new selection (at least for trajectories). If there isn't a selection for the node - maybe a button to add one. Will also add some more checks for other entity types.

PR #1048, opened 2025-11-29
Fix Blender __slots__ warnings

PardhavMaradani:
Blender shows a lot of the following warnings when run in debug mode (`--debug-all`):

```
register_class(...): warning, AnnotationInputs: is expected to contain a "__slots__" member to prevent arbitrary assignments.
register_class(...): warning, AnnotationProperties: is expected to contain a "__slots__" member to prevent arbitrary assignments.
```

In addition, there are single warnings for the following:

```
register_class(...): warning, EntityProperties: is expected to contain a "__slots__" member to prevent arbitrary assignments.
register_class(...): warning, MolecularNodesObjectProperties: is expected to contain a "__slots__" member to prevent arbitrary assignments.
register_class(...): warning, MolecularNodesSceneProperties: is expected to contain a "__slots__" member to prevent arbitrary assignments.
register_class(...): warning, TrajectorySelectionItem: is expected to contain a "__slots__" member to prevent arbitrary assignments.
```

This PR adds the `__slots__` member for `AnnotationInputs`, `AnnotationProperties` and `EntityProperties` classes as I added them and am sure there are no dynamic attributes added at runtime for those and adding this seems sensible based on what I read about preventing arbitrary assignments and conserving memory. The last three, I am not sure, but since they are single warnings, maybe we could ignore.

### DIFF
--- a/molecularnodes/annotations/manager.py
+++ b/molecularnodes/annotations/manager.py
@@ -190,8 +190,9 @@ class BaseAnnotationManager(metaclass=ABCMeta):
     @classmethod
     def _update_annotation_props(cls, annotation_class: BaseAnnotation):
         """Update annotation properties attached to Object"""
+        attributes = {"__slots__": []}
         AnnotationProperties = type(
-            "AnnotationProperties", (BaseAnnotationProperties,), {}
+            "AnnotationProperties", (BaseAnnotationProperties,), attributes
         )
         # Add each annotation type inputs as a pointer to a separate property group
         for annotation_type, annotation_class in BaseAnnotationManager._classes.items():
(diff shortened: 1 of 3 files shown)


BradyAJohnston:
I haven't come across this before but thanks for looking into it. If you wanted to add them for the last 3 as well - they should all be static for their properties.

PardhavMaradani:
Hi Brady, I am not sure how to add the `__slots__` attributes for the last three classes. The first two of them have over 20+ properties. We could list all of them, but it will be error prone if someone later adds a property, but misses adding to the `__slots__` list (the property then cannot be set), etc. This cannot be done dynamically as it has to be set at the time of class creation. For the `AnnotationInputs` and `AnnotationProperties` I added, it was straight forward because the classes are created dynamically. The other class (`EntityProperties`) was a really small one.

PardhavMaradani:
It looks like adding an empty `__slots__` list is sufficient. This gets past the Blender warning, works with the attributes defined in the class, gives us all the benefits of `__slots__` usage like preventing arbitrary assignments, potential speedup, better memory management, etc. The API access of the properties also works fine as a lot of these are already part of our tests. Thanks

BradyAJohnston:
Thanks for looking into this - I hadn't come across it before and haven't run Blender in debug before.

PR #1053, opened 2025-12-03
UI: Change style panel for better readability

BradyAJohnston:
Removes the "top level" box that all node inputs were being drawn in. Also adds some types to the code and a check for if an enum should be expanded to match the node group.

| Main | PR |
|--------|--------|
| CleanShot 2025-12-03 at 12 54 48 | CleanShot 2025-12-03 at 12 54 13 |

### DIFF
--- a/molecularnodes/ui/panel.py
+++ b/molecularnodes/ui/panel.py
@@ -820,7 +820,7 @@ class MN_PT_Styles(bpy.types.Panel):
         entities_active_index: int = scene.mn.entities_active_index
         uuid: str = scene.mn.entities[entities_active_index].name
         entity = get_session().get(uuid)
-        if not entity:
+        if entity is None:
             return
         node_group = entity.node_group
         if node_group is None:


molecularnodes/ui/panel.py
```diff
@@ -815,28 +815,31 @@ def poll(cls, context):
...
         scene = context.scene
-        entities_active_index = scene.mn.entities_active_index
-        uuid = scene.mn.entities[entities_active_index].name
-        entity = scene.MNSession.get(uuid)
+        layout = self.layout
+        assert layout is not None
+        entities_active_index: int = scene.mn.entities_active_index
+        uuid: str = scene.mn.entities[entities_active_index].name
+        entity = get_session().get(uuid)
+        if not entity:
```
PardhavMaradani: Maybe we should use the explicit is None check that we fixed in PR #1037 to avoid the len() based comparison in databpy?

BradyAJohnston: yeah good point. That approach overall is better as a style for the codebase as numpy arrays raise an error unless explicitly checking for `is None`.

PardhavMaradani:
Hi Brady, the `is_expanded` check seems broken for 4.5. It works for 5.x+

Here is what I see:
menu-expand-4 5-error

Here is the error:
```
  File "/tmp/blender-4.5.0-linux-x64/4.5/python/lib/python3.11/site-packages/molecularnodes/ui/panel.py", line 907, in draw
    is_expanded: bool = item.id_data.interface.items_tree[
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'bpy_prop_collection[key]: key "Input_2" not found'
```

Here is the difference between 4.5 and 5.0:
```
4.5

>>> sn = bpy.data.objects['NewUniverseObject'].modifiers['MolecularNodes'].node_group.nodes['Style Spheres']
>>> print(sn.node_tree.interface.items_tree.get("Input_2"))
None

5.0

>>> sn = bpy.data.objects['NewUniverseObject'].modifiers['MolecularNodes'].node_group.nodes['Style Spheres']
>>> print(sn.node_tree.interface.items_tree.get("Input_2"))
<bpy_struct, NodeTreeInterfaceSocketMenu("Geometry") at 0x723bb0828f80>
```

Since 4.5 is something we support, this probably needs fixing. Thanks

BradyAJohnston:
Thanks for flagging - I had forgotten that I had to fix this in Blender for 5.0: https://projects.blender.org/blender/blender/pulls/140196

In 4.5 we'll just have to fallback to default of not expanding.

PR #1054, opened 2025-12-03
Add enum input support for annotations

PardhavMaradani:
This PR adds support for annotations to have enum inputs. On the UI, these show up as regular `EnumProperty` dropdowns.

Here is an example:

```py
class EnumInputTest(mn.entities.trajectory.TrajectoryAnnotation):
    annotation_type = "enum_input_test"

    enum_input: list[str] = ["enum value 1", "enum value 2", "enum value 3"]

    def draw(self) -> None:
        pass
```

annotation-enum-input

This will allow greater flexibility for custom annotations in the future. Thanks

### DIFF
--- a/docs/api/annotations.qmd
+++ b/docs/api/annotations.qmd
@@ -439,7 +439,7 @@ class CustomAnnotation(mn.entities.trajectory.TrajectoryAnnotation):
         params.text_size = 20
 
     # optional validate method
-    def validate(self) -> bool:
+    def validate(self, input_name: str = None) -> bool:
         # validate any input params that change (either through API or GUI)
         # return True if validation succeeds else False or raise an Exception
         params = self.interface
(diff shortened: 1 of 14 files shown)


BradyAJohnston:
We should probably follow the enum property convention from Blender: https://docs.blender.org/api/current/bpy.props.html#bpy.props.EnumProperty

It's more verbose - but fits with their convention for the property (which it will ultimately become). 

```py
test_items = [
    ("value_1", "Enum Value 1", "", 1),
    ("value_2", "Enum Value 2", "", 2),
]
```

PardhavMaradani:
One of the things I wanted for annotation inputs was to use regular python annotations to specify their type and not expose any of the `bpy` stuff (The assumption being that not everyone knows Blender Python interface and don't necessarily have to for the most common stuff). Technically, we could have used the `bpy.props` as is or in some form to make it easier to specify a lot more info (like explicit name, description, etc), but that seemed to cross that line. The above way to specify simple string enums (which are regular string inputs too for code purposes) seems a bit too verbose for me.

BradyAJohnston:
I think supporting just a `list[str]` would be OK - but we should also support the more complex Blender-style enum defintion if someone wanted more customisation over enum & descriptions etc.

PardhavMaradani:
> ... but we should also support the more complex Blender-style enum defintion if someone wanted more customisation over enum & descriptions etc.

Hi Brady, I am presuming this is for later and there is nothing actionable from me on this current PR. Please let me know if that is not the case. Thanks

BradyAJohnston:
Given you are adding the functionality for the enum in this PR I'd prefer it also support the Blender-style enum definition. Additionally it should handle a tuple / list as input.

PardhavMaradani:
About blender-style enum definitions, I don't quite like the idea of passing descriptions etc as part of raw values. Note that the value has to also be set after the property is created based on what is passed. Either we have a `bpy.props.` like separate property definitions or reuse `bpy.props`, which is I believe too much for this use case (and adds in an explicit `bpy` knowledge dependency). Note that none of the other inputs have a way to pass along descriptions, ui name, etc today. (They use regular python annotations to specify type) Also, none of the current bundled annotations need this input type today. I only added this to have some additional flexibility if and when we add annotation snippets/templates. Any changes beyond what's in this PR will require a lot more careful thought, without which, this would become needlessly complex. Due to this, I will close this PR for now and we can reopen it when a specific use case for this arises. Thanks

BradyAJohnston:
Given your points about other annotation properties - I think we could accept this and like you say somebody adding an annotation likely won't be making it for other users and wouldn't need descriptions and the like. If you wanted to re-open the PR.

PardhavMaradani:
If not a `bpy.props` like class etc, maybe even a simple dict that can help with a more descriptive way to provide inputs is something we could consider for the future. (Blender's Flamenco uses something similar in JavaScript to create ui layouts in Blender). I will re-open this for the basic support for now.

BradyAJohnston:
I think the flamenco example is a good one and something to take inspiration from.

PR #1055, opened 2025-12-03
UI: Expose node Label property with placeholder

BradyAJohnston:
Re-exposes the style nodes `Label` property, using the nodes `Name` as placeholder text. 

I've gone with `emboss=False` as the emboss is visually too much and also doesn't allow selecting each item without starting the editing of the property. This way the user has to double-click to change the property.

https://github.com/user-attachments/assets/c24b9f93-122a-4d25-98ec-ef329d047d43

Extension of #1053

### DIFF
--- a/molecularnodes/ui/panel.py
+++ b/molecularnodes/ui/panel.py
@@ -826,7 +826,7 @@ class MN_PT_Styles(bpy.types.Panel):
         entities_active_index: int = scene.mn.entities_active_index
         uuid: str = scene.mn.entities[entities_active_index].name
         entity = get_session().get(uuid)
-        if not entity:
+        if entity is None:
             return
         node_group = entity.node_group
         if node_group is None:


PardhavMaradani:
I am not sure why the names are grayed out in the video above. Grayed out ones give the idea that they are not editable - they also seem inconsistent with the list item names in Entities and Annotations.

Here is how the previous code looked like:
```
            col.prop(item, "label", text="", emboss=False)
```

BradyAJohnston:
They are greyed out here because the label is no longer set by default - so it uses the "placeholder" text as a fallback. I did this because if I just display the property - it will display nothing with labels being empty.

The alternative is to start setting the label property again - but this was proving problematic with swapping of nodes.

The greyed out look does distinguish it from a manually created label - but I agree that it makes it look not editable.

PardhavMaradani:
I haven't looked at what the original issue is with labels and swapping of nodes, but it seems like fixing that would be the right way to go to keep this all consistent. Thanks
ps. When we swap the current node with a new node, we should probably just set it's label to that node name (or maybe even carry over the previous node label)

BradyAJohnston:
The previous approach of setting the node label to the node tree's name was required because node groups would display `Style Spheres.001` at the top of the node. This has been changed at some point in Blender and now even if you have multiple nodes with the same node tree and "conflicting" names - it displays just the `Style Spheres`. Because of that reason I removed it in #1044, as we no longer needed to set the label for nicer display.

I think the label should be reserved for a user's custom label. We don't need to change the label as the node tree is swapped out but the same node (`GeometryNodeGroup`) remains inside of the node tree.

The only alternative to the current implementation is for us to check the _current_ node label against possible style node names and if it is one of the defaults then we set the new node label from the incoming node tree.

I think this adds extra handling and complexity for no real gain. The current setup means if a user sets a label it will remain. This can be set when they initially add the style or after they add the style by double clicking (less discoverable) but ultimately isn't not a vital feature currently and they can still discover it via label creation or by just double clicking which is a common way to explore a UI.

BradyAJohnston:
I would prefer that it wasn't greyed out - but the complexity required to add it I don't think justifies it.

PardhavMaradani:
I am not sure I quite follow, but from what I see, the logic for the name currently being displayed on top of a Geometry Node in the GN editor is as follows: if (label exists) use_that_value else use_the_linked_node_tree_name (Eg: `bpy.data.node_groups["Style Spheres"].name`). This is why the `.001` etc no longer show up on the top in the editor, even though the node name has it (because it has to be unique). I don't see why the label cannot be just set to the linked node tree name to solve this issue. (I would have still preferred a unique name here - it is currently not an issue because styles API doesn't have a name based lookup, but this could be a problem later). Thanks

BradyAJohnston:
The problems arise when you swap out the node tree but keep the node (which is what happens when we swap style nodes). The node remains and thus the label remains while the node tree is changed. Do we then change the label or not? If it is a "custom" label from the user or own own labelling code then we want to keep it, if it is a "default" label then we should change it. This requires a check and some code to handle it. We can instead just avoid this all together by default _not_ setting the label and just use the node name or node tree name for display. Then if a user sets the label, we just leave it "as is" when if we swap the style node.

BradyAJohnston:
For this PR at least - this approach of displaying the label and having fallback text is the way to go regardless of the other approach we take with labels. If someone a node ends up without a label then it will just be blank in the list - so I think this PR will be merged and we can potentially change how we label nodes going forward if need be.

PardhavMaradani:
Hi Brady, using a placeholder name for the prop (which is the path for the most common workflows) is problematic. In addition to this being just bad UX (editable grayed out items), it also breaks the filtering and sorting as they currently rely on the label property. (unless of course there are additional changes to fix those, which would be messy)

The changes are not complex, on the contrary, they are pretty straightforward. I will create a new PR that shows the changes. With these changes:
- No grayed out items in our styles ui list
- Style node swaps work exactly as expected - a custom name is always retained and any default name changes over
- This works with even style nodes added manually in the GN editor - a label is always present - no empty list item names
- No complex comparisons against all style names, etc

PR #1056, opened 2025-12-04
Fix node label changes

PardhavMaradani:
The PR fixes the node labels issue as described in this [comment](https://github.com/BradyAJohnston/MolecularNodes/pull/1055#issuecomment-3610598069) of PR #1055 

As mentioned in the past, it would have been ideal to use the node name to ensure they are unique and users have a 1:1 correspondence between the n-panel entries and the names in the GN editor, but keeping as node tree names as you wanted it. These changes work fine with the node swap support as well. Thanks

### DIFF
--- a/molecularnodes/ui/panel.py
+++ b/molecularnodes/ui/panel.py
@@ -1053,7 +1053,8 @@ class MN_PT_Annotations(bpy.types.Panel):
                 if prop_name in ("uuid", "valid_inputs"):
                     continue
                 row = box.row()
-                if hasattr(instance, f"_{prop_name}"):
+                nbattr = f"_custom_{prop_name}"  # non blender property
+                if hasattr(instance, nbattr):
                     # indicate use of non blender property in draw
                     row.label(icon="ERROR")
                     row.alert = True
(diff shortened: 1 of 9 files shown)


molecularnodes/ui/panel.py
```diff
@@ -747,13 +747,7 @@ def draw_item(
...
             col.label(text=seqno)
             col = split.column()
-            col.prop(
-                data=item,
-                property="label",
-                placeholder=item.name,
-                text="",
-                emboss=False,
-            )
+            col.prop(item, "label", text="", emboss=False)
```
BradyAJohnston: I'm happy with the other label changes - but if a user removes the label or adds a node via adding a node group (not via the add operator) then in both cases the `label` will be empty and the menu UI will be blank. Having the fallback property be the name is still useful in these cases (and good that it is greyed out).

BradyAJohnston (changes requested):

molecularnodes/ui/panel.py
PardhavMaradani: Added placeholder

PR #1059, opened 2025-12-05
Add basic support for text editor templates

PardhavMaradani:
This PR adds  support for a templates menu in Blender text editor that can open any of our packaged templates. Templates can be arranged into different folders (currently only two levels to keep it simple) for better organization.

### DIFF
--- a/molecularnodes/annotations/props.py
+++ b/molecularnodes/annotations/props.py
@@ -39,16 +39,19 @@ def create_annotation_type_inputs(
             )
         elif stype is bool:
             prop = BoolProperty(
+                description=attr,
                 default=getattr(annotation_class, attr, False),
                 update=_create_update_callback(update_callback, attr),
             )
         elif stype is int:
             prop = IntProperty(
+                description=attr,
                 default=getattr(annotation_class, attr, 0),
                 update=_create_update_callback(update_callback, attr),
             )
         elif stype is float:
             prop = FloatProperty(
+                description=attr,
                 default=getattr(annotation_class, attr, 0.0),
                 update=_create_update_callback(update_callback, attr),
             )
@@ -74,6 +77,16 @@ def create_annotation_type_inputs(
                 default=getattr(annotation_class, attr, default),
                 update=_create_update_callback(update_callback, attr),
             )
+        elif hasattr(stype, "__name__") and stype.__name__ == "list":
+            if str(stype) == "list[str]":
+                items_list = getattr(annotation_class, attr, None)
+                if isinstance(items_list, list) and items_list:
+                    items = [(item, item, item) for item in items_list]
+                    prop = EnumProperty(description=attr, items=items)
+                else:
+                    continue
+            else:
+                continue
         else:
             continue
         attributes["__annotations__"][attr] = prop
@@ -138,7 +151,12 @@ def create_property_interface(
                     if value not in bpy.data.materials:
                         raise ValueError(f"No material with name '{value}'")
                     value = bpy.data.materials[value]
-            setattr(_prop(), attr, value)
+            if isinstance(value, list) and value:
+                # for list[str] (i.e., EnumProperty types), set default to
+                # the first value of the list
+                setattr(_prop(), attr, value[0])
+            else:
+                setattr(_prop(), attr, value)
             # clear any corresponding non blender property
             if hasattr(instance, nbattr):
                 delattr(instance, nbattr)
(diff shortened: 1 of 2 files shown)


PardhavMaradani:
Hi Brady, can you give this a quick try when you can and let me know your thoughts? Not sure if this is inline with what you were thinking too. I didn't want to expose everything in the filesystem directly, so added an explicit list and also kept it as a simple two level instead of  recursively going down directories for arbitrary submenu levels, etc.

molecularnodes/templates/menu.py
```diff
@@ -0,0 +1,62 @@
...
+            self.layout.menu(_get_submenu_idname(subdir), text=subdir.title())
+
+
+def _draw_templates_menu(self, context):
+    """Draw method for the main menu item"""
+    self.layout.menu("MN_MT_templates")
+
+
+def register_templates_menu():
+    # register submenu classes
```
BradyAJohnston: a lot of these comments are pretty unnecessary, maybe have an overview comment of a line or two explaining what is going on but each line of code should mostly explain itself

molecularnodes/templates/menu.py
```diff
@@ -0,0 +1,62 @@
...
+    return type(subdir, (bpy.types.Menu,), attributes)
+
+
+class MN_MT_templates(bpy.types.Menu):
+    bl_idname = "MN_MT_templates"
+    bl_label = "Molecular Nodes"
+
+    def draw(self, context):
+        for subdir in _subdirs:
+            self.layout.menu(_get_submenu_idname(subdir), text=subdir.title())
```
BradyAJohnston: Rather than having a second layer of nested menus - we should probably just have a single submenu with all of them listed (under MN) (unless the number of templates grows to be _huge_) - we can separate them with menu UI separators for visual categories.

BradyAJohnston (commented):
I like the initial approach - I think we only need the one layer of menus though.

PardhavMaradani:
Hi Brady, are you sure you want all these templates just flat under `Text Editor > Templates > Molecular Nodes`? Note that there is no filtering of these and `path_menu` just sorts them alphabetically with no additional tooltip info. I went down this path thinking that you have different categories of scripts that be bundled (like basic api examples, additional material setup (from `nodes/material.py`), analysis code, custom annotations, maybe simple rendering setup scripts etc) and at least have a basic two level setup to ensure we are covered for the future with some organization to start with.

If you confirm you only want this flat, I'll move the current two scripts to a single `scripts` folder in `molecularnodes/templates` and change the menu setup accordingly - it would be a single `path_menu` call and wrap this up. Please let me know. Thanks

BradyAJohnston:
I'm meaning something along these lines: 

```py
from pathlib import Path
import bpy


class MN_MT_templates(bpy.types.Menu):
    bl_idname = "MN_MT_templates"
    bl_label = "Molecular Nodes"

    def draw(self, context):
        layout: bpy.types.UILayout = self.layout # type: ignore
        assert layout is not None

        for i, subdir in enumerate(Path(__file__).parent.glob("*")):
            if not subdir.is_dir() or subdir.name.startswith("_"):
                continue
            if i > 0:
                layout.separator()
            layout.label(text=subdir.name.title())
            self.path_menu(
                searchpaths=[str(subdir)],
                operator="text.open",
                props_default={"internal": True},
            )


def _draw_templates_menu(self, context):
    """Draw method for the main menu item"""
    self.layout.menu("MN_MT_templates")


def register_templates_menu():
    bpy.utils.register_class(MN_MT_templates)
    bpy.types.TEXT_MT_templates.append(_draw_templates_menu)


def unregister_templates_menu():
    bpy.types.TEXT_MT_templates.remove(_draw_templates_menu)
    bpy.utils.unregister_class(MN_MT_templates)

```

Which gives you this menu: 

image


It doesn't have to be exactly like this. Not having the labels is probably best (or adding better label handling for proper capitals) but this way they are still partially separated by category but we don't fall into the trap of menus in menus in menus.

PardhavMaradani:
I added a simple folder map that allows us to specify explicit labels and folders (for better control).
It now looks as follows:
image

Hope this is ok.

molecularnodes/templates/menu.py
PardhavMaradani: Done

BradyAJohnston:
Yep - looks and works great!

BradyAJohnston:
I removed WIP label as you had marked it ready for review.

PR #1060, opened 2025-12-05
Fix: Add more water resnames for trajectory `is_solvent`

### LINKED ISSUE #1051: Water model support for the "Is solvent" selector node
At the moment the "Is solvent" node does not support TIP3, and I presume other water models. 

I would find it useful to include TIP3/4 and maybe even SCP/E in the node. Otherwise a general "Is water" selector node could be useful.

BradyAJohnston:
Fixes #1051 by adding more water residue names for different water models. 

Need feedback on if these are actually correct / if there should be any more.

### DIFF
--- a/docs/api/annotations.qmd
+++ b/docs/api/annotations.qmd
@@ -495,6 +495,48 @@ Custom annotations get automatically registered with the [](`~mn.entities.trajec
 manager = mn.entities.trajectory.TrajectoryAnnotationManager
 manager.unregister_type("custom_annotation")
 ```
+### Input types
+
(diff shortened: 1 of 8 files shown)


PardhavMaradani:
When adding a trajectory from the API as follows:
```py
# Add universe to Blender
u = mda.Universe(PSF, DCD)
```

There is a new print I see in the notebook cell:
```
e=NoDataError('This Universe does not contain chainID information')
e=NoDataError('This Universe does not contain chainID information')
```

This is because of the `print` in `molecularnodes/entities/trajectory/base.py` due to changes here:
```py
...
    def _store_default_attributes(self) -> None:
        """Store default attributes"""
        ...
            except (mda.NoDataError, AttributeError) as e:
                print(f"{e=}")
                logger.debug(f"Skipping attribute '{name}': {e}")
        ...
```
Hi Brady, not sure if this is intentional, just bringing it to your attention. Thanks

BradyAJohnston:
Thanks for flagging - I thought I had specifically removed that one but must have left it in another branch. Removing in #1069

PR #1064, opened 2025-12-12
Docs: Add table of supported annotation input types

PardhavMaradani:
This PR adds a simple table to the docs that lists the currently supported annotation input types and their corresponding representation in Blender GUI. This was missing item in the docs:

This looks as follows in the Custom Annotations section of the Annotations API doc:

image

### DIFF
--- a/docs/api/annotations.qmd
+++ b/docs/api/annotations.qmd
@@ -518,9 +518,7 @@ Text, lines and meshes drawn in custom annotations can be controlled by the same
 
 Here is an example within a custom annotation that draws two 3D circles, the first with a hardcoded color value and the second with a custom input color:
 
-```{python}
-#| eval: false
-
+```python
     ...
     # custom line color input
     line_color_input: tuple[float, float, float, float] = (1, 1, 1, 1)


PardhavMaradani:
Added a section about param overrides as well:

image

docs/api/annotations.qmd
```diff
@@ -495,6 +495,50 @@ Custom annotations get automatically registered with the [](`~mn.entities.trajec
...
+|`tuple[float, float, float, float]`|A vector input in 4 dimensions. Input names that contain the word "color" will become Color selections|
+|`list[str]`|An Enumeration drowpdown|
+
+### Param overrides
+
+Text, lines and meshes drawn in custom annotations can be controlled by the same params as described in the [Common Annotations Params](#common-annotation-params) section above. Custom annotations can also override these parameters to either hardcode specific values or allow additional customization though separate inputs. This can be done by passing the `overrides` keyword param to any of the drawing utilities and specify a dictionary of the parameters to override along with their values.
+
+Here is an example within a custom annotation that draws two 3D circles, the first with a hardcoded color value and the second with a custom input color:
+
+```{python}
```
BradyAJohnston: If you just change this this to 

```
```python
```

quarto will just treat it as a markdown code block and you won't need the 
`eval: false`.

PardhavMaradani: Didn't think there was a difference between `python` and  `{python}` - fixed now

Issue #872, opened 2025-06-02
Secondary Structure from MD Trajectory Import


BradyAJohnston:
Previously secondary structure wasn't available when importing via MDAnalysis, but was available using the `DSSP` node. 

It has since become possible to calculate via MDAnalysis: https://github.com/MDAnalysis/mdanalysis/pull/4304

It would be useful to have a toggle to calculate secondary structure during import for cartoon style.

PR #1067, opened 2025-12-16
Add Helix Analysis example template

PardhavMaradani:
* Fix missing update callback for Annotation Enum inputs
* Minor fix to run RMSD analysis only initially and on selection change

---

This PR adds an example template for [Helix analysis](https://userguide.mdanalysis.org/stable/examples/analysis/structure/helanal.html) from the MDAnalysis user guide. There is additional stuff we can do like drawing the local axes, global axis, etc, but this PR is limited to just displaying the various charts. It is also the first example of using the enum annotation input.

Here is an example of how this looks:

https://github.com/user-attachments/assets/4cedb4e6-54fd-40d7-bdc3-34312c025cce

### DIFF
--- a/docs/api/annotations.qmd
+++ b/docs/api/annotations.qmd
@@ -466,6 +466,18 @@ class CustomAnnotation(mn.entities.trajectory.TrajectoryAnnotation):
             self.draw_text_3d(atom.position, atom.name)
 ```
 
+#### The [](`~molecularnodes.annotations.base.BaseAnnotation.defaults`) method
+
+This method is called only once when the annotation instance is created. Common annotation param defaults and other initializations that need to be done once can be setup in this method.
+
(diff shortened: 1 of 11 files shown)


PardhavMaradani:
Added options to show global and local axes:

helix-local-global-axis

BradyAJohnston:
Thanks this is looking good. Somewhat related but not relevant to this current PR is computing secondary structure and updating it on the trajectory during playback. #872 It's something I have been meaning to get around to but would complement well with this kind of analysis and annotation.

PardhavMaradani:
> Somewhat related but not relevant to this current PR is computing secondary structure and updating it on the trajectory during playback. https://github.com/BradyAJohnston/MolecularNodes/issues/872 It's something I have been meaning to get around to but would complement well with this kind of analysis and annotation.

I will try and take a look. Thanks

PR #1071, opened 2025-12-18
Fix trajectory averaging outside valid frame range

### LINKED ISSUE #761: Trajectory averaging outside of frame range
With trajectory averaging on, if an average of `3`, then averaging still happens even when outside of final frame range by up to 3 frames as the averaging function can still fetch "back" 3 frames into the actual trajectory. This should not be the case and if outside of the frame range we should instead just return the final (or first) frame.

ref: https://github.com/yuxuanzhuang/ggmolvis/pull/22

Prasun-13:
This PR fixes an issue where trajectory averaging could access
frames outside the valid range near the start or end of a trajectory.

The frame indices returned during averaging are now clamped to
[0, n_frames - 1], preventing out-of-range access.

Fixes #761

### DIFF
--- a/molecularnodes/entities/trajectory/helpers.py
+++ b/molecularnodes/entities/trajectory/helpers.py
@@ -255,7 +255,6 @@ class FrameManager:
 
         return frames
 
-
     def update_position_cache(self, frame: int, cache_ahead: bool = True) -> None:
         """Update the position cache for the current frame.
 


BradyAJohnston:
Hi @Prasun-13 thanks for looking into this. Can you please address the linting issues raised by `ruff` so the tests can run.

Prasun-13:
Thanks for pointing that out!  
I’m looking into the ruff linting issues now and will push an update shortly.

Prasun-13:
Thanks for pointing that out!  
I’ve addressed the ruff linting and formatting issues and pushed the updated changes.  
Please let me know if anything else needs adjusting.

BradyAJohnston (approved):

PR #1072, opened 2025-12-18
Add DSSP support for trajectories

### LINKED ISSUE #872: Secondary Structure from MD Trajectory Import

PardhavMaradani:
Hi Brady, this PR adds [DSSP](https://docs.mdanalysis.org/stable/documentation_pages/analysis/dssp.html) support for trajectories that you mentioned in this [comment](https://github.com/BradyAJohnston/MolecularNodes/pull/1067#issuecomment-3664921575) and pointed to an existing issue #872. Instead of an annotation, I think it would be better to have this as a direct import option for trajectories (both though GUI and API) - it might be helpful later too when everything moves to an MDA backend. This should fix #872 

Here are some before and after examples:

Before|After
:---|:---
psf_dcd_before|psf_dcd_after
md_ppr_before|md_ppr_after

For the API, there is a new `use_dssp` param and for the GUI, there is a new checkbox as follows:
use_dssp

The DSSP [example](https://docs.mdanalysis.org/stable/documentation_pages/analysis/dssp.html#calculating-average-secondary-structure-of-a-trajectory) from the user guide shows how an average value can be used as well. Using this (which is the default) leads to a more smoother animation as these don't change per frame. This is a runtime option that can be changed from the API using `dssp_type` attribute in the trajectory or as below from the GUI (The GUI option only shows when there is valid dssp run info):
dssp_type

I added a couple of tests (with and without dssp) for basic validations.

One issue I ran into that you probably know more than me why it happens: From the GUI, the sec structure only show up after a frame change. It is the same from the API if the style is created without any selection (eg: `t = mn.entities.Trajectory(u, use_dssp=True).add_style("cartoon")` requires a frame change to see the results, but for `t = mn.entities.Trajectory(u, use_dssp=True).add_style("cartoon", selection="all")`, this shows up right away in the viewport.


molecularnodes/ui/ops.py
```diff
@@ -734,6 +734,11 @@ class MN_OT_Import_Trajectory(bpy.types.Operator):
         description="Add nodes to the scene to load the trajectory",
         default=True,
     )
+    use_dssp: BoolProperty(  # type: ignore
+        name="Use DSSP",
+        description="Assign secondary structure using the DSSP algorithm",
+        default=False,
```
BradyAJohnston: I think we can probably have the `default=True` as most users would want secondary structure for their visualisations.

molecularnodes/entities/trajectory/base.py
```diff
@@ -144,6 +151,9 @@ def __init__(
         self._updating_in_progress = False
         self.annotations = TrajectoryAnnotationManager(self)
         self.frame_manager = FrameManager(self)
+        self._using_dssp = False
```
BradyAJohnston: I think we should use an object-level property for this similar to average / dssp type etc.

This way someone can enable / disable computation of SS after import. For the UI we can expose the tick box. Someone can then import with SS, turn off the computation and "re-use" the SS from the first frame for the rest of the animation if they wish.

molecularnodes/entities/trajectory/base.py
```diff
@@ -230,6 +240,33 @@ def _elements(self) -> np.ndarray:
...
+    def _setup_dssp(self) -> None:
+        try:
+            # From: https://docs.mdanalysis.org/stable/documentation_pages/analysis/dssp.html
+            self._dssp_run = DSSP(self.universe).run()
+            self._dssp_mean = translate(
+                self._dssp_run.results.dssp_ndarray.mean(axis=0)
+            )
+            self.calculations["sec_struct"] = self._calculate_sec_struct
+            self._using_dssp = True
+        except Exception:
```
BradyAJohnston: Try and be a bit more specific with which exceptions we are wanting to handle / what we are expecting from a failed selection.

molecularnodes/ui/panel.py
```diff
@@ -387,6 +390,10 @@ def panel_md_properties(layout, context):
     row.enabled = traj._is_orthorhombic
     col.prop(obj.mn, "interpolate")
 
+    if traj._using_dssp:
```
BradyAJohnston: This is where we could expose the `use_dssp` object-level property for the UI.

PardhavMaradani:
> Only real change is I think we should allow for someone to enable or disable the SS computation after import. 

I did think about having this as a dynamic option like you suggested, but wasn't sure how intensive the analysis run could be, etc - so thought it'd be best to leave it at initial import. If we are making this true by default, one other option is to include a "None" option to the `dssp_type` enum (i.e., none, average, per_frame) - that way, it can be easily turned off (ie., not show up) and on. Currently, the actual analysis run is only done once and just the results get used in per-frame changes. This would keep things far simpler. What do you think about this option? Thanks

BradyAJohnston:
Oh I didn't realise it was all being run on import. The "ideal" would be that it runs on frame change (not sure how possible this is) - so that we are only ever computing what is currently relevant (and could support streaming trajectories also).

Depends on how feasible this is and what kind of overhead is associated with calling the analysis each frame instead of once on initial import.

I think your suggestion of adding the `None` to the enum is probably better than us having another boolean property _and_ the enum property.

PardhavMaradani:
> The "ideal" would be that it runs on frame change (not sure how possible this is) - so that we are only ever computing what is currently relevant (and could support streaming trajectories also)

The dssp run can be done for specific frames - i.e., on frame change for us, but then, we'll not have the concept of averaging because that requires a run across the whole trajectory. It also introduces an analysis run every frame change and if we make this the default, it could make scrubbing the timeline etc a bit laggy. I also tried out for streaming trajectories, but couldn't get this to work yet (even though there is frame slicing support) - seems like this is common for most other analysis stuff too that depend on `n_frames`.

> adding the None to the enum is probably better than us having another boolean property and the enum property.

I wasn't thinking of removing the original `use_dssp` boolean - that would still determine whether the dssp analysis is run on import (and default to True as you suggested). The enum is the runtime option that has an additional None to not show the sec struct if needed. This same enum could also be brought into the n-panel as well. Is that ok?

PardhavMaradani:
Hi Brady, the above commit is the closest we can get to "ideal" case.

I also figured out why the sec structures were not showing up until a frame change when importing a trajectory from the GUI or from the API without an explicit selection string. (the issue I described at the end of the initial comment). The reason for this is that an `update_entities` is needed after the node tree is setup after adding a style for the calculations to run and take effect. In the call through API with explicit selection string (`t = mn.entities.Trajectory(u).add_style("cartoon", selection="all")`), the selection phrase causes the addition of a new selection in the global selections list, which calls `update_entities` as part of the property update callback.

Here is a summary of the changes:
- The DSSP analysis object is now setup by default for trajectories. For trajectories that don't have enough info (like some examples in tests that didn't have names and resids) this fails gracefully and they continue to work as before
- There is a single enum called `dssp` that determines how the results show up. The values are `per-frame`, `average` and `none`. The default now is `per-frame`, which runs the analysis (the actual `.run(...)`) per frame
- When the user picks `average`, a full `.run()` is performed at that time
- When the results of a full run are available (above point), these essentially serve as cached data, so the `per-frame` setting just uses the values from these instead of running the per-frame analysis again
- Because of the issue mentioned above, I changed the default from `None` to "all" for style selection param. This is a slightly disruptive change. The effect of this is that a new selection is created and a corresponding named attribute node is added to the node tree. This required changes to the tests which were checking for the exact number of nodes after a style branch is created. One good thing about this is that this will make it easier to expose the selection for each of the styles (as we discussed in the style swap PR - this was one of the issue I pointed out there).
- The `sec_struct` calculation had to be skipped from picking because it causes a recursive loop. I didn't see any other way to address this

molecularnodes/ui/ops.py
PardhavMaradani: This doesn't exist anymore

molecularnodes/entities/trajectory/base.py
PardhavMaradani: The single `dssp` enum now controls what gets displayed.

molecularnodes/entities/trajectory/base.py
PardhavMaradani: The DSSP Analysis docs specify that raises a `ValueError`, but in tests, I noticed that it also raised attribute errors (for topologies that didn't have resids or names), no data errors and others. We still want to proceed gracefully, so I added a log mesage to at least have some visibility.

PardhavMaradani:
Forgot to add a few rough performance measurements I made earlier in my setup:
- Previously, tests in `test_trajectory.py` took around 16 seconds
- With the current changes and the default of `per-frame`, the same tests take 40 seconds
- With the current changes and dssp option `none`, the same tests take 25s
  - The additional time can be accounted for the `DSSP` setup that happens during trajectory init
- With the current changes and dssp option `full`, the same tests take 35 seconds
  - Though this might seem counter intuitive, I think this can be attributed to several frame changes in the tests
- I am not sure if the `Peptide DSSP`  option in the Cartoon style is accurate (it shows a very small selection), but with that turned on for the trajectory animation, in my setup, I see an fps of around 1.8, where as with the MDA one, it is the full 24 fps (for the regular `PSF`, `DCD` universe)

PardhavMaradani:
The current [DSSP Analysis Class](https://docs.mdanalysis.org/2.9.0/documentation_pages/analysis/dssp.html#analysis-classes) unfortunately doesn't work with streaming trajectories. There is however a standalone [assign](https://docs.mdanalysis.org/2.9.0/documentation_pages/analysis/dssp.html#MDAnalysis.analysis.dssp.dssp.assign) function that can be used to get the sec structs, but the changes are a bit more involved. They seem to work fine for some simpler cases, but requires more testing, etc. I think it is best to handle those as a separate PR for the `StreamingTrajectory` entity later. Thanks

BradyAJohnston:
> I am not sure if the Peptide DSSP option in the Cartoon style is accurate 

You are right that this has broken. I'm not sure when but it seems like it has been broken for a while & I haven't noticed it. I've got a  version that is working again (in the branch you have seen) that is about 15x faster than what was previous running (and not working). 

Given potential performance hits that you have detailed, it does make me think that maybe instead we should have SS computation disabled by default. Whether we still compute and store `sec_struct` on the first frame might still be a good idea, but then have re-computing the SS an option that a user has to enable. That way if someone isn't using it (just using sphere style) it won't affect performance for them. 

I've seen you PR about the current frame iterator which should help us here. Additionally I've opened a PR that will help speed up SS computation : https://github.com/MDAnalysis/mdanalysis/pull/5182

PardhavMaradani:
Hi Brady, I changed the default to none. Thus, this will not change any existing behavior and even the `DSSP` class is not initialized. However, this can be changed at runtime and this will work seamlessly. Thanks

PardhavMaradani:
Here is the regular psf/dcd viewport animation using the mda dssp computed average values:

https://github.com/user-attachments/assets/cf26be56-4bc2-426a-8f8f-f40b42194aad

BradyAJohnston:
Looks like something strange is happening with the cartoon style - likely an issue with the node and not this PR. I'll have a look into it.

PardhavMaradani:
> Looks like something strange is happening with the cartoon style - likely an issue with the node

Hi Brady, just fyi that I am using the `node_data_file.blend` from [here](https://github.com/BradyAJohnston/MolecularNodes/blob/e9d7e9650dbd7e0075406daaf497dabf00c0a041/molecularnodes/assets/node_data_file.blend). On a related note, enabling the `Peptide DSSP option` in the style shows nothing at all - might be because this is still work-in-progress. I don't know much about this to be of additional help. Thanks

```py
import MDAnalysis as mda
import molecularnodes as mn
from MDAnalysis.tests.datafiles import DCD, PSF

u = mda.Universe(PSF, DCD)
canvas = mn.Canvas()
```
```py
```

BradyAJohnston:
Ah OK - looking at it just from this branch it all works fine so that's something to worry about for that branch: 
image

molecularnodes/entities/trajectory/base.py
```diff
@@ -230,6 +235,62 @@ def _elements(self) -> np.ndarray:
...
+            mask = np.isin(universe.residues.resids, self._dssp_run.results.resids)
+            self._dssp_resindices = universe.residues.resindices[mask]
+        # create attribute data
+        if self.dssp == "average":
+            dssp_chars = self._dssp_mean
+        else:
+            index = 0 if self._dssp_mean is None else frame
+            dssp_chars = self._dssp_run.results.dssp[index]
+        dssp_ints = self._dssp_vmap(dssp_chars)
+        attribute_data = np.zeros(len(universe.atoms), dtype=int)
```
BradyAJohnston: This array has basically already been created with `no_sec_struct` - we could just initialize the `sec_struct` array like you currently do and then potentially update it and return here.

molecularnodes/entities/trajectory/base.py
```diff
@@ -230,6 +235,62 @@ def _elements(self) -> np.ndarray:
...
+            try:
+                self._DSSP = DSSP(universe)
+            except Exception as e:
+                self._DSSP = None
+                logger.warning(f"Failed to setup DSSP: {e}")
+            self._dssp_init = True
+        if self.dssp == "none" or self._DSSP is None:
+            return no_sec_struct
+        frame = universe.trajectory.frame
+        if self._dssp_mean is None:
```
BradyAJohnston: I think an ideal mean calculation would allow user input to change the window size - rather than just doing the entire trajectory. If we have a property for window size we could just use it here to mask +/- 3 frames and get the mean value. We should expect some trajectories to change conformation quite a lot, so smoothing of the calculation (through a mean value from multiple frames) while still allowing for a change in the computed secondary structure would be ideal.

molecularnodes/entities/trajectory/base.py
```diff
@@ -230,6 +235,62 @@ def _elements(self) -> np.ndarray:
...
+        else:
+            index = 0 if self._dssp_mean is None else frame
+            dssp_chars = self._dssp_run.results.dssp[index]
+        dssp_ints = self._dssp_vmap(dssp_chars)
+        attribute_data = np.zeros(len(universe.atoms), dtype=int)
+        attribute_data[self._dssp_resindices] = dssp_ints
+        return attribute_data[universe.atoms.resindices]
+
+    def _setup_dssp(self) -> None:
+        self._DSSP = None
```
BradyAJohnston: Should we have a small helper class to accessing all of the dssp-related stuff? Might help clean it up over tracking multiple properties via prefix.

PardhavMaradani:
> and allow a user to change the window over which we average SS computed frames

Hi Brady, given the requirements are evolving, could we agree upon how this will eventually look (from both the GUI and API perspective) before I go ahead and make any further changes? Here is what I have in mind that could potentially address all the issues discussed so far:

- Have a separate `DSSP` panel in the n-panel
- Have an `Initialize` button (operator) in the panel
  - This only shows up as long as `DSSP` is not initialized
  - No other options are shown along with it - they show up automatically after initialization
  - This would be equivalent to disabling `DSSP` by default and allowing users to enable it if they need cartoon style, etc
  - Both the GUI operator and API will use a common `initialize` method (`<trajectory>.dssp.initialize()`) - using a helper class for all `DSSP` stuff
- Once intitialized, show a drop down with the following display options:
  - None
    - This option will not show any SS
  - Per Frame
    - This option will invoke a dssp run for the current frame unless a full run across the trajectory (see 'Trajectory Average' below) is available
    - Per frame SS are shown
  - Average of +/- N Frames
    - This option will show another integer property selection below the drop down (min 1)
    - A dssp run for the current frame +/- n frames is run for every frame change (sliding window clamped at frame beginning and end)
    - No caching of the results
    - Does not show up for streamed trajectories
  - Trajectory Average
    - This option will invoke a full dssp run and cache the results
    - The initial idea for the mean came from the user guide [example](https://docs.mdanalysis.org/2.9.0/documentation_pages/analysis/dssp.html#calculating-average-secondary-structure-of-a-trajectory), which was run on the whole trajectory
    - We could potentially add a threshold option as described [here](https://docs.mdanalysis.org/2.9.0/documentation_pages/analysis/dssp.html#find-parts-of-the-protein-that-maintain-their-secondary-structure-during-simulation) in the future
    - Does not show up for streamed trajectories
- An `Apply` button below the above options
  - This is disabled if none of the options change
  - An explicit click of this is required after changing the drop down or int selection values above
    - For streamed trajectories, this might seem like an unnecessary step, but this will keep it consistent
  - This invokes the operator which calls the common `display(...)` method used by the API as well
  - This is where the the full run or the initial average +/- n frames run etc happen - similar to the `Fetch` button for PDB download, this stays clicked till the blocking op is complete (this is the best we can get - adding another popup for confirmation etc, will make it unnecessarily complex)
    - We will obviously not do a full run if already available, etc
- Since the `sec_struct` calculation is not pickled, the same path from initialization again will work for the save and restore case

Your thoughts? Thanks

molecularnodes/entities/trajectory/base.py
PardhavMaradani: Calculated `no_sec_struct` separately and once to keep it inline with how the nodes compute today - 3 for protein, 0 for all else. The one in above code was wrong.

molecularnodes/entities/trajectory/base.py
PardhavMaradani: Added a sliding window support

molecularnodes/entities/trajectory/base.py
PardhavMaradani: Added a new DSSP Manager

PardhavMaradani:
Hi Brady, I went ahead and made changes as per the previous comment. A few minor changes from there - mostly for better usability, but the rest are the same. You could give this a try. Here is a video that shows the new options:

https://github.com/user-attachments/assets/c9edefb8-a92e-4737-ada7-2ce85d3ee1ba

BradyAJohnston:
Thanks for continuing to work on this - have been away over the Christmas break. 

## UI Tweaks

After testing it a bit I don't think we need the `Apply` / `Cancel` for the sliding window. Given the computation over the smaller window should be relatively quick - when the changes the value I think we should just use a callback function to update what is happening for the computation (unless there is a reason that might get messy). Obviously we have the issue that if a user was to input a very large number, then it would be potentially slow to update. I think we have a `soft_max` of maybe `10` so that via clicking & dragging in the UI they max out at 10, but can always manually input a higher value and press enter (at which point a slow down wouldn't be un-expected).

image



The `Threshold` input should be a `Factor` slider given it is a value between 0 and 1.

image


## Computation Tweaks

This is something maybe we think about going forward and not necessarily for this PR (which I think is otherwise close to being done) - but even with the window averaging I am still seeing "popping" of secondary structure where if I have a window average of `10` there might be a single frame where a bit of SS disappears and then re-appears on the next frame. The ideal usage of the sliding frame average would to remove these single frame changes to remove visual noise from the overall playback. It's not immediately obvious to me why this might currently be happening, but it might be something we try and 'fill in' better in the future.

PardhavMaradani:
> I think we have a soft_max of maybe 10 so that via clicking & dragging in the UI they max out at 10, ...

The soft_max will only limit the UI slider range initially. Once it is manually set (to say 1000), that would be the max slider value from then on and any slider change after that would just slow down everything. Are you sure you want me to change this to a soft_max and remove the apply and cancel for it?

> The Threshold input should be a Factor slider given it is a value between 0 and 1.

I used the same term as was in the [documentation](https://docs.mdanalysis.org/stable/documentation_pages/analysis/dssp.html#find-parts-of-the-protein-that-maintain-their-secondary-structure-during-simulation) and computing it the same way. So, could you confirm that I should rename this to Factor?

> The ideal usage of the sliding frame average would to remove these single frame changes to remove visual noise from the overall playback. It's not immediately obvious to me why this might currently be happening, but it might be something we try and 'fill in' better in the future.

Currently, we just pass in all the frames in the sliding window to the dssp run and just do a mean. Maybe this will get addressed if we add in a threshold / factor for this too? (similar to the full trajectory mean)

BradyAJohnston:
> I used the same term as was in the [documentation](https://docs.mdanalysis.org/stable/documentation_pages/analysis/dssp.html#find-parts-of-the-protein-that-maintain-their-secondary-structure-during-simulation) and computing it the same way. So, could you confirm that I should rename this to Factor?

I meant to leave the name as is, but you can set the property [subtype](https://docs.blender.org/api/current/bpy.props.html#bpy.props.FloatProperty) to be [`FACTOR`](https://docs.blender.org/api/current/bpy_types_enum_items/property_subtype_number_items.html#rna-enum-property-subtype-number-items) so that it will be a "filled in" slider from 0 to 1.

> The soft_max will only limit the UI slider range initially. Once it is manually set (to say 1000), that would be the max slider value from then on and any slider change after that would just slow down everything. Are you sure you want me to change this to a soft_max and remove the apply and cancel for it?

I think the easier UX benefit for tweaking values in the lower ranges would be worth it. If someone is setting to such large values then they should expect it to slow down.

> Currently, we just pass in all the frames in the sliding window to the dssp run and just do a mean. Maybe this will get addressed if we add in a threshold / factor for this too? (similar to the full trajectory mean)

If you wanted to play around with it or leave it for later on I don't mind.

PardhavMaradani:
I made the above changes and also added a threshold comparison option for sliding window average as follows:
image
I however don't know if it helps with the use case you had  - we can remove/hide it if it doesn't help. Thanks

molecularnodes/ui/props.py
```diff
@@ -77,6 +80,110 @@ class EntityProperties(bpy.types.PropertyGroup):
...
+                "Sliding Window Average",
+                "Average secondary structures of a sliding window of frames",
+            ),
+            (
+                "trajectory-average",
+                "Trajectory Average",
+                "Average secondary structures across all frames",
+            ),
+        ),
+        default="none",
```
BradyAJohnston: After doing some testing - I think the default should be `per-frame`. If the user is initializing the DSSP module then we should start with at least some value computed. The `none` option should still be there for someone to disable.

PardhavMaradani: Agreed

PR #1073, opened 2025-12-22
Fix DSSP node

BradyAJohnston:
Fixes and speeds up DSSP node. Still not 100% equivalent to DSSP algorithm but unable to do so without list support in GN.

- **initial fixing of DSSP node**
- **better refactor of dssP**
- **fixed helix detection**
- **more cleanup of sheets DSSP**
- **tweaking of sheets**
- **fixed dssp node**


BradyAJohnston:
Need to also fix SS sheet display when of length 1: 
image

PR #1080, opened 2026-01-07
Swap X and Z axes of mrcfile-loaded maps

### LINKED ISSUE #993: Misalignment of PDB and MRC files when imported into Blender
We’re trying to visualize both density and atomic structures in Blender using your add-on, specifically, an MRC (density map) and a PDB (atomic model) file. Both files were previously aligned in ChimeraX, and their positions match correctly there.
 
However, when importing them into Blender via the add-on, the two files are placed differently, their origins don’t match, so the PDB and MRC are placed in completely different locations (see attached files). When trying to correct this manually, we even noticed size differences, making it nearly impossible to align them manually and get a scientifically correct image.
 
The model and density originate from pdb 4gb2. The density was converted from .mtz to .mrc using the clipper plugin in chimeraX (just in case that you cannot recreate the misalignment / wrong scale with a standard .mrc)

Image

Image

PlethoraChutney:
Closes #993 

`mrcfile` loads the data into a numpy array which has the Z axis as the first dimension, but Grid wants the Z axis as the last dimension. This commit swaps the two when returning a map loaded using the `mrcfile` fallback. Tested using `EMD-62270` and pdb `9KDD` as described in the original issue.

### DIFF
--- a/tests/test_density.py
+++ b/tests/test_density.py
@@ -126,7 +126,7 @@ def test_fallback_reading():
 
 def test_fallback_transforms():
     def evaluated_obj(file):
-        density = mn.entities.density.load(file)
+        density = mn.entities.density.load(file, overwrite=True)
         return mn.blender.mesh.evaluate_using_mesh(density.object)
 
     space_group_one = evaluated_obj(data_dir / "62270-small_sg1.mrc")


BradyAJohnston:
Thanks for looking into this and finding a simple fix! Definitely works on my machine as well. I would like to have a test for this going forward - but I can't really download and re-save using ChimeraX during a test. We could add another test file is probably the way to go.

PlethoraChutney:
At some point I'll try to figure out why the ChimeraX maps fail the Grid method -- I can't see any obvious differences between the files on disk...

BradyAJohnston:
Yeah it would be good to know but as long as it's working for us that's OK I suppose. Could you add two example files, one map that has been processed and the other straight from the EMDB? Ideally if they were small maps from somewhere that would be ideal. 

I think this should work for a test:
```py
def test_fallback_transforms():
    def evaluated_obj(file):
        density = mn.entities.density.load(file)
        return mn.blender.mesh.evaluate_using_mesh(density.object)

    obj_default = evaluated_obj(file_default)
    obj_chimerax = evaluated_obj(file_chimerax)

    pos_default = databpy.Attribute(obj_default, "position")
    pos_chimerax = databpy.Attribute(obj_chimerax, "position")

    assert np.array_equal(pos_default.as_array(), pos_chimerax.as_array())
    ```

PlethoraChutney:
Will do RE: example files. Would you mind assigning this to me so I don't lose track of it?

BradyAJohnston:
Appreciate it! If the tests end up giving you grief just upload some good example files and I'll put something together test-wise as testing volume objects is messy currently.

PlethoraChutney:
OK, figured it out. It seems like ChimeraX (and CryoSPARC, and others) both write `0` to the ISPG field of the MRC header. This marks them as image stacks, not volumes, according to the [MRC spec](https://www.ccpem.ac.uk/mrc-format/mrc2014/#:~:text=ISPG-,space%20group%20number,-6). It seems like basically everyone ignores this header field except `MDAnalysis`, which raises a `ValueError`. `mrcfile` ignores this flag and loads it anyway, so we're all good -- just need to swap the axes b/c of differing conventions.

Figuring out what causes a file to go to down one path or the other made it easy for me to generate very small MRC files which exhibit this bug (hard to do when I couldn't use ChimeraX!). Those are the two new test data files (`sg0` for "space group 0", which should fail the Grid path, and `sg1` for "space group 1", which should work with `MDAnalysis`).

Regarding tests: I had to modify yours a bit (maybe the `databpy` interface changed?) and they pass, but so does the old code, at least when I commented out my fix. I could use your help checking that I've got everything right!

PlethoraChutney:
Oh, I also changed the `test_fallback_reading()` test to use the local `_sg0.mrc` file (which fails the Grid I/O) instead of downloading one from EMDB to save a bit of time during testing.

BradyAJohnston:
Another classic example of everyone just silently going against spec and not writing it down properly anywhere. Thanks so much for looking into this! I've just formatted with `ruff` and pushed those changes but otherwise this all looks good!

BradyAJohnston:
I am also struggling to get this test to fail without the change as well. Testing with volumes is just weird at the moment. Will keep poking around and once I can get that to reliably fail & then pass I'll merge.

BradyAJohnston:
Okay just got it working. As the `.vdb` intermediate is created on load I needed to change `load(file, overwrite=True)` so that is properly overrides the created `.vdb` file. 

I can confirm the test now fails without your change and passes with it!

PR #1093, opened 2026-01-30
Fix MN module path for templates

PardhavMaradani:
Blender adds a separate prefix for extension modules since 4.2+. When MN is installed through the GUI, MN module path becomes `bl_ext.blender_org.molecularnodes`. If templates or other scripts don't use the correct module, entities will not get registered with `MNSession` and any custom annotations will also not get registered to the correct module (and hence not show up in the GUI).

This PR adds a check in the packaged templates that work with both the GUI and API cases.

### DIFF
--- a/molecularnodes/annotations/manager.py
+++ b/molecularnodes/annotations/manager.py
@@ -65,7 +65,7 @@ class BaseAnnotationManager(metaclass=ABCMeta):
     annotation classes, instances and interfaces
 
     Entities that need annotation support have to derive from this base class and
-    set the class attribute _entity_type to the entity tpye, the class attribute
+    set the class attribute _entity_type to the entity type, the class attribute
     '_classes' and instance attribute '_interfaces' to empty dictionaries. Derived
     classes will have to pass the entity instance as part of its '__init__' as well.
 
@@ -164,7 +164,7 @@ class BaseAnnotationManager(metaclass=ABCMeta):
 
         # Add a dynamic wrapper to ensure custom signature is retained
         def dynamic_wrapper(cls, annotation_class, **kwargs):
-            return cls._create_annotation_instance(annotation_class, **kwargs)
+            return cls._create_annotation_instance(annotation_class, None, **kwargs)
 
         method = functools.partialmethod(dynamic_wrapper, annotation_class)
         parameters = [
@@ -287,11 +287,6 @@ class BaseAnnotationManager(metaclass=ABCMeta):
     def visible(self, value: bool) -> None:
         """Visibility of all annotations - setter"""
         self._entity.object.mn.annotations_visible = value
-        if value:
-            self._draw_handler_add()
-        else:
-            self._draw_handler_remove()
-        self._update_annotation_object()
         viewport_tag_redraw()
 
     @classmethod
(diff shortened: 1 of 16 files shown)


BradyAJohnston:
The problem with hard-coding the prefix is that it will change depending on _how_ the extension is installed. Most users would likely come across it from the extensions platform but if it is installed via another method (or they rename the repo) then it will change.

This is an issue that we run into by treating the Extension and the "python package" as the same entity. This is supposed to add the extension to path: https://github.com/BradyAJohnston/MolecularNodes/blob/e6dc55b146dc219a450137b778a92f213d522c2d/molecularnodes/utils.py#L11-L13

But it might not be working properly?

PardhavMaradani:
> But it might not be working properly?

The above method does add the extension module path to sys.path, but that doesn't help with the `bl_ext.blender_org.` prefix that gets added internally for these modules.

I agree with the problems with the hard-coded prefix. I've now modified to check this dynamically and verified that it works fine. It looks a bit weird to use a common method from our utils to import the extensions module, but I didn't find any other way and wanted to avoid repeating the code in all templates. It also seemed like something that can be useful from the API too. Thanks

BradyAJohnston:
I'm still a little bit confused by this one. If we are already able to do `import molecularnodes as mn` then how is returning the add-on module any different? The module has already been added to path so we are returning the same module / folder in both? Can you give me examples of the failure cases where it is currently going wrong?

PardhavMaradani:
Hello Brady, for the same single path of our extension, there are two corresponding modules in Blender.

Here is how you can quickly verify after opening the Python Console of a clean install with MN extension enabled:

```
>>> import molecularnodes as mn1
>>> mn1
<module 'molecularnodes' from '/home/pardhav/blender/clean/blender-5.0.1-linux-x64/portable/extensions/blender_org/molecularnodes/__init__.py'>

>>> import bl_ext.blender_org.molecularnodes as mn2
>>> mn2
<module 'bl_ext.blender_org.molecularnodes' from '/home/pardhav/blender/clean/blender-5.0.1-linux-x64/portable/extensions/blender_org/molecularnodes/__init__.py'>
```

Note that it is the same path, but two different modules, one with a Blender prefix.

You can also verify this from `sys.modules`:

```
>>> import sys
>>> for m in sys.modules:
...     if m.endswith("molecularnodes"):
...         print(m, sys.modules[m])
...         
bl_ext.blender_org.molecularnodes <module 'bl_ext.blender_org.molecularnodes' from '/home/pardhav/blender/clean/blender-5.0.1-linux-x64/portable/extensions/blender_org/molecularnodes/__init__.py'>
molecularnodes <module 'molecularnodes' from '/home/pardhav/blender/clean/blender-5.0.1-linux-x64/portable/extensions/blender_org/molecularnodes/__init__.py'>

>>> 
```

So, for everything within Blender, it is the `bl_ext.blender_org.molecularnodes` module (for the above default setup) that matters as can be seen here:

```
>>> type(bpy.context.scene.MNSession)
<class 'bl_ext.blender_org.molecularnodes.session.MNSession'>
```

So, any scripts that do a regular import `import molecularnodes as mn` will be working with a different module. The `fetch_from_pdb.py` template for example will not register the molecule entity with the session shown above. The trajectory template examples similarly will not have the annotation types show up because they didn't register to the correct module. Hope this helps with reproducing the problem.

Here is a relevant snippet from Blender's `scripts/addons_core/bl_pkg/readme.rst` file:

```
...
Other Modules
-------------

Some functionality that's relevant to the extensions system is implemented in other modules,
as it relates to how extensions are loaded by Blender.

*Paths are relative to Blender's source tree.*

``./scripts/modules/_bpy_internal/extensions/junction_module.py``
   This stand-alone module allows extensions to appear as if they are all loaded
   from a single *package* independent of their file-system location.

   This is done so extensions don't pollute the module name-space
   (avoiding naming collisions with packages downloaded from ``https://pypi.org``).

   So each extension's add-on ID follows this format: ``bl_ext.{repository_id}.{extension_id}``.
...
```

PardhavMaradani:
Note that if we don't add the `add_current_module_to_path()` during register, then we cannot do a `import from molecularnodes as mn` from the Python Console as well and would need to know the exact module prefix. Thanks

BradyAJohnston:
Ah OK thank you for the additional explanation I get what is going on now. 

Should we maybe stop adding the add-on to `sys.path` then - given the potential confusion that it creates? If MN is installed in a python environment then you would do `import molecularnodes` but when installed as an add-on we require more specific imports.

PardhavMaradani:
> Should we maybe stop adding the add-on to sys.path then - given the potential confusion that it creates?

I am not sure, Brady. The Blender module prefix can be variable as you pointed out earlier - it also annoyingly doesn't tab-complete even in Blender's Python Console. Not having this added means that all templates we add need to include code at the top to determine the correct module (as opposed to what's in the current PR - `mn = mn.utils.load_extension_module()` which works universally). Even for the API case, let us say from a notebook, we don't necessarily require a separate `molecularnodes` package installation and can use the same extension module if it is installed and enabled from Blender's GUI.

PardhavMaradani:
A few more data points that might help decide which path to take:

- For a pure API use (ie., background mode), there is only the `molecularnodes` package from pypi, so there are no issues
  - Note that any blend files (and corresponding MNSession files) saved cannot be restored in a GUI context, this is because pickled data will include the module details (like `molecularnodes.session.MNSession`, where as in a pure GUI context, it needs to be `bl_ext.blender_org.molecularnodes.session.MNSession`). This can also be verified with the `python3 -m pickletools file.blend.MNSession` command. This is one of the problems with pickling
- For an API use with a notebook and GUI mode, we don't need to install the `molecularnodes` package from pypi. We can use the extension module
  - This will require the API code to explicitly load the MN module as `import bl_ext.blender_org.molecularnodes as mn` or iterate through sys.modules to load the correct module if we don't want hard-coded stuff
  - If we left the add-on path in sys.path like we have today, `mn = mn.utils.load_extension_module()` would be possible to ensure we are working with the correct module
  - If we are using the extension module in the API, there will be full compatibility between files saved and re-loaded between API <-> GUI in both directions
- For a pure GUI case, the packaged templates are essentially API calls and hence need to load the correct module. The same two cases in the previous point apply here

My recommendation after a bit more thinking would be to stop adding the add-on to `sys.path`, leave the `utils.load_extension_module()` for convenience (for power users who customize repo or module name) and change any API imports to address the most common scenario and not make the import itself very complex:

For templates (this issue):
```
import bl_ext.blender_org.molecularnodes as mn
```

Any API use (to cover both GUI mode and background mode):
```
try:
    import bl_ext.blender_org.molecularnodes as mn
except ModuleNotFoundError:
    import molecularnodes as mn
```

PardhavMaradani:
Hi Brady, I updated the PR as per the last comment. I can change it if you think there is a different approach we should go with. Thanks

BradyAJohnston (approved):

PR #1099, opened 2026-02-05
Fix density ISO surface style slicing offset

PardhavMaradani:
This PR fixes the slicing offset used by the density ISO surface style to account for the cell-centered nature of the volume. The issue shows up clearly when the invert option is used to import the density entity - the tiny half delta offset error leads to those surfaces not showing up. Additionally, the contour thickness prop is now sub type factor (since it is from 0 to 1) and the slicing props are sub type percentage (since they go from 0 to 100).

Before:
density-slicing-offset-fix-before

After:
density-slicing-offset-fix-after

### DIFF
--- a/molecularnodes/nodes/arrange.py
+++ b/molecularnodes/nodes/arrange.py
@@ -207,9 +207,11 @@ def calculate_node_dimensions(
     unconnected_vector_input_count = len(
         list(
             filter(
-                lambda input_socket: input_socket.enabled
-                and input_socket.type == "VECTOR"
-                and socket_input_connection_count[input_socket] == 0,
+                lambda input_socket: (
+                    input_socket.enabled
+                    and input_socket.type == "VECTOR"
+                    and socket_input_connection_count[input_socket] == 0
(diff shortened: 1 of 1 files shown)


PardhavMaradani:
@BradyAJohnston , the ruff failure is in a file that wasn't even touched in this PR. Maybe it's version changed? I didn't see the error locally when I ran the check. I'll try and take a look. Thanks

PardhavMaradani:
> @BradyAJohnston , the ruff failure is in a file that wasn't even touched in this PR. Maybe it's version changed? 

`ruff` 0.15.0 was [released](https://pypi.org/project/ruff/#history) recently and this error is due to that change. I am able to reproduce this locally after updating the ruff version. I fixed it as part of this PR to get past this and have the tests run. Thanks

BradyAJohnston:
Thanks for fixing this and the ruff issue. If I remember correctly Blender might actually be switching to the corner values in 5.1 - but that's something to be tested / addressed for later versions.

Issue #944, opened 2025-08-27
Compatibility Request- GROMACS Martini (v3) Coarse-Grained Molecular Dynamics Structures (.gro or .pdb)


kbwarren17:
Hi! My feature request is currently not related to an active problem per-se, however, I would like to investigate whether or not coarse-grained structures could feasibly be made compatible with the blender MN package. At the moment, VMD is essentially the only visualization software that is both able to read GROMACS output topologies, and then successfully re-map the bonding networks to display the MD simulation outputs with visualized bonding networks (via help from the MartiniGlass package found here: https://github.com/Martini-Force-Field-Initiative/MartiniGlass). Unfortunately, visualization within VMD cannot match the level of detail and freedom offered by Blender and the MN package. Do note that the coarse-grained .pdb or .gro files do not explicitly contain bead connectivity information (and from my understanding -- cannot), and as such, these systems cannot be modelled properly in Blender through MN as no bonds can be drawn -- it's just a mess of beads).

Ideally, we could try to implement some way for the MN package to read-in coarse-grained .gro/.pdb files, while simultaneously reading-in the pertinent .top file that itself calls on the .itp files that stores the inter-bead bonding information. Note that the information contained within the .itp files will have to be processed  in some form as-done in the MartiniGlass package (see link above) to reconstruct the bonding network. For example, the MartiniGlass package will read your default .itps that contain the bonding information, then generate new .top files that need to be loaded into VMD (with your coarse-grained topology already loaded) to re-assemble inter-bead bonding networks for your system's molecules (which then facilitates showing the system in licorice-like representations and CPK).

[...]

PR #1109, opened 2026-02-23
Added mn_is_elastic edge attribute for elastic networks

### LINKED ISSUE #1061: Topology node for elastic networks
Need to add a topology node for detecting if an edge is is part of an elastic network or if it is a "topology" edge. 

Current best is this - but this would fail for 'elastic' network edges within a single residue (only really the case for MN simulations). The below will work for course-grained simulations.

Image

Initially brought up in #944

spyke7:
Hello!
In this PR, basically I tried to solve the issue if an edge is is part of an elastic network or not. 
Related issue - 
fixes #1061 

Changes made in this PR - 
- created `_compute_is_elastic_edge` func, that returns bonds whose length are basically greater than 2.5 Angstrom
- inside `_create_object` func, implemented a check for this particular thing, and thus added an edge attribute `mn_is_elastic` 
- Added tests inside `test_trajectory.py`  
 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Firstly, talking about the 2.5Angstrom cutoff. Most of the bonds length are below 2 angstrom, and maximum length of a covalent bond can go upto like 2 angstrom. So, i.e, why created this threshold of 2.5 angstrom. Anything above it is basically flagged as elastic bond. This is done inside `_compute_is_elastic_edge`.

In this PR, I have just tried to put an attribute that solves the issue. But not till made the geometry node as asked in the issue, but will definitely try to make it, and so to generate to PR to apply fixes and edits.

Applied this on a sample data. I have tried to find suitable data having bonds length above 2.5 angstrom, but was not able to find (or sort) it perfectly. (if it is there plz tell me)
So, downloaded the dna strand from here - [dna_strand from cgmartini](https://cgmartini.nl/docs/downloads/example-applications/other-systems.html#dna-strand)
Put those inside data inside tests.

Here is the some info about the dna_strand data - 
script - 
```
import MDAnalysis as mda
import numpy as np
u = mda.Universe(
    "dna.itp",
    "dna.pdb",
    topology_format="ITP"
)

print("Atoms:", len(u.atoms))

bonds = u.atoms.bonds
indices = bonds.indices
positions = u.atoms.positions

pos_a = positions[indices[:, 0]]
pos_b = positions[indices[:, 1]]
lengths = np.linalg.norm(pos_b - pos_a, axis=1)

print(f"Covalent: {(lengths < 2.5).sum()}")
print(f"Elastic: {(lengths > 2.5).sum()}")
print(f"Min: {lengths.min()} A")
print(f"Max: {lengths.max()} A")
```
output - 
```
Atoms: 130
Covalent: 40
Elastic: 747
Min: 2.2441799640655518 A
Max: 6.4992852210998535 A
```
As you can see, the maximum bond length goes above 6 angstrom.

inside blender scripts and simulations - 
```
import bpy
import MDAnalysis as mda
import molecularnodes as mn
mn.register()

u = mda.Universe(
    "dna.itp",
    "dna.pdb",
    topology_format="ITP"
)

traj = mn.entities.Trajectory(u, name="DNAElastic")
obj = bpy.data.objects["DNAElastic"]
```

And this is the output-  
Screenshot (197)

Inside geomtery nodes - 
Screenshot (198)

This is the geometry nodes part, of how I removed the edges which are elastic. 
Added named attribute node, and chose mn_is_elastic , connected it with Delete Geometry node..
Screenshot (199)

@BradyAJohnston can you pls have a look at this, and tell me any updates that can be made, and on further to adding a topology node


spyke7:
I just came to know about Martini pentapeptide system already present in the data, could have used that in the blender sim....

BradyAJohnston:
Hi @spyke7 thanks for looking into this! Apologies for delay in getting back to you, lots of travel and work at the moment. I haven't looked in detail at the code yet, but rather than creating a new attribute, I think we can use the `bond_type` attribute which is already being created and stored.

https://github.com/BradyAJohnston/MolecularNodes/blob/aff023a66e01d22b80d5c3edc36a9cfe91a67118/molecularnodes/entities/utilities.py#L175-L185

We can decide on an integer to use that will just represent if the bond is an elastic one (nothing comes to mind, open to suggestions) and ensure that we are properly storing it. 

I think looking at this currently that the `Trajectory` might not be storing bond types as it is, so this might need to be improved to better reflect the bond types possible.

spyke7:
> We can decide on an integer to use that will just represent if the bond is an elastic one (nothing comes to mind, open to suggestions) and ensure that we are properly storing it.
> 

Simply, we can do like integer 8 for elastic bonds, and returning 8 instead of boolean in `_compute_is_elastic_edge`.

BradyAJohnston:
I think `8` is good of a value as any. I think it should all go into `_compute_edge_type()` function, that would include ` array.bonds.as_array()[:, 2], ` and the new comparisons you are doing. We maybe move it out of the `create_object()` function and instead into the greater molecule / trajectory initialisation.

spyke7:
Screenshot (214)
 is this looking good?
 I have just checked for elastic edges inside `_compute_edge_type()`

BradyAJohnston:
Yep that's looking great! I might also add a new topology node that outputs bond type, but I can handle that later (adding new nodes is currently a bit messy).

BradyAJohnston:
Although is that currently doing the inverse of the selection?

spyke7:
> Although is that currently doing the inverse of the selection?

Ig not. It's deleting edges greater than 2.5 Angstrom length 

Before using delete geometry node for bond type 8 - 

Screenshot (215)

After connecting the delete geometry node - 

Screenshot (216)


Pls see this

molecularnodes/entities/trajectory/base.py
```diff
@@ -456,6 +457,21 @@ def _create_object(self, name: str = "NewUniverseObject") -> None:
...
+            bond_types = _compute_edge_type(
+                bonds_array=np.column_stack(
+                    [
+                        self.atoms.bonds.indices,
+                        np.zeros(len(self.atoms.bonds), dtype=int),
+                    ]
+                ),
+                positions=self.atoms.positions,
+            )
+            self.object.data.attributes.new(name="bond_type", type="INT", domain="EDGE")
```
BradyAJohnston: This can use the `self.store_named_attribute(data=data, name=name, domain="EDGE")` that follows what is done here: 

https://github.com/BradyAJohnston/MolecularNodes/pull/1109/changes#diff-821c651ceba9d0c5ec7afd6cfb4f761d816410d4dfc2e582d77013c117bbf27dR222

molecularnodes/entities/trajectory/base.py
```diff
@@ -456,6 +457,21 @@ def _create_object(self, name: str = "NewUniverseObject") -> None:
             edges=self.atoms.bonds.indices if hasattr(self.atoms, "bonds") else None,
         )
 
+        if hasattr(self.atoms, "bonds") and len(self.atoms.bonds) > 0:
+            bond_types = _compute_edge_type(
```
BradyAJohnston, line 468: Can this also take into account the `bond.type` that you mentioned here? (haven't actually looked into it) https://github.com/BradyAJohnston/MolecularNodes/issues/1061#issuecomment-3938332694

If we can potentially look first at the topology and get some info from there that would be ideal - but otherwise falling back and just using a distance cutoff.

BradyAJohnston (changes requested):
Looking good for approach - just some clarification for potentially using info from the topology (not sure if actually available or not)>

molecularnodes/entities/trajectory/base.py
spyke7: ok this is done

molecularnodes/entities/trajectory/base.py
spyke7: uh, bond.length() gives the length I calculated using numpy, so I will remove it with this. I checked in jupyter that the results were same.

Regarding bond.type, I am figuring which one to be as elastic 
I ran a code where length if greater than 2.5 is elastic and else covalent and based on it stored the type here is the output - 
```
Covalent bonds :   3953
Elastic bonds  :  31272
Length range   : 1.81 – 10.79 Å

Elastic bond type-pair breakdown:
     Nda – Nda       6505
      P5 – P5        5679
      P5 – Nda       3675
     Nda – P5        3116
      P5 – P4        1667
      P4 – P5        1415
      N0 – Nda        945
     Nda – P4         945
      P4 – Nda        840
      P5 – N0         635
     Nda – N0         628
      N0 – P5         535
      P5 – Nd         380
     SC4 – SP1        360
      P4 – P4         305
     Nda – SC5        270
      P5 – C1         258
      P5 – SC5        250
      P5 – Qa         208
     Nda – C1         198
      N0 – Qd         190
     Nda – SC4        170
     Nda – Qa         155
      P5 – Na         145
      P4 – N0         130
      N0 – P4         125
      Nd – Na         120
     SC4 – SC5        120
      N0 – N0         110
      Na – Nda         95
      P4 – Nd          95
      P5 – SC4         90
     Nda – C5          82
      C3 – Qd          80
      Na – P5          75
      Nd – N0          60
     SC5 – SC5         60
      N0 – Na          50
      Nd – Nd          50
      Na – Na          40
     Nda – Na          40
     Nda – Nd          40
      P4 – Na          40
      P5 – C5          35
      Nd – Nda         30
     SNd – SC5         28
      Qd – P5          25
     Nda – C3          21
      Na – C1          20
      Na – P4          20
      Nd – SC4         20
      Nd – SC5         20
      P5 – C3          14
      N0 – Nd          10
      Na – N0          10
      Nd – C1          10
      Qd – C5           9
      Nd – Qa           8
      Nd – P4           7
      Qd – P4           5
      Qa – N0           4

```

spyke7: if you could tell that bond.type for elastic for any type of universe revolves around this output above, then it is easy to just store this in a list and check for every atom.bonds.type and classify them as `8` or elastic

BradyAJohnston: I'm not an expert in these simulations so I wouldn't be sure what is considered elastic or not. Maybe we can track down those definitions somewhere? Might be that we have an option on import to also adjust the distance threshold for considering as an elastic bond on import as this might change for different CG simulations.

spyke7: Ig taking user input to determine elastic bonds, should be best with a default of 2.5A. Where should it be placed? like while importing or in the MN side panel like as an annotation?
But I would like to ask MDA maintainers regarding elastic bonds before proceeding....., (will ask in discord)

BradyAJohnston: It could go into the import menu / operator - but lets try and see if anyone has any tips on topology for us. The distance check can also be done in the form or a selection node where someone can adjust the cutoff in the node tree as well.

spyke7: Not got the required thing

As I looked into [vermouth martini elastic network](https://vermouth-martinize.readthedocs.io/en/latest/tutorials/elastic_networks.html), the user can change there the -el and -eu, i.e, lower and upper bounds. So, in mda universe we cannot specifically check for a particular tag/bond type. So Ig calculating distance with a default value of 2.5A, and after that user can always tweak to specify elastic bonds.
Or else, without setting a default value, we ask user for the value that is required for classifying elastic bonds

spyke7: https://docs.mdanalysis.org/stable/documentation_pages/topology/TPRParser.html#module-MDAnalysis.topology.TPRParser

BradyAJohnston: From the notes in the parser:

> regular bonds (type 1), G96 bonds (type 2), Morse (type 3),
  cubic bonds (type 4), connections (type 5), harmonic potentials (type 6),
  FENE bonds (type 7), restraint potentials (type 10),
  tabulated potential with exclusion/connection (type 8),
  tabulated potential without exclusion/connection (type 9), constraints with
  exclusion/connection (type 1), constraints without exclusion/connection (type
  2), SETTLE constraints

I'm not sure if we try to support these directly or just do a single distance based cutoff that the user can adjust on import and leave it at that. Might be worth just doing that for now and leaving it as is.

spyke7: Ok then,
Can you pls tell me exactly where to provide this field for setting the cutoff length, like in the import part where user selects trajectory or imports pdb files
or in the side panel?
then I can send  you the screenshot regarding it and the proper space where this cutoff field is there. 
after this I can properly set up the geometry nodes part where user can change this value of cutoff. 

Ig all the things remain same, but just with a new field for user cutoff and changing it!!

(not having a bond type to certainly distinguish elastic bonds really hectic 🥲)

BradyAJohnston:
I had forgotten about the details discussed in this thread: https://github.com/BradyAJohnston/MolecularNodes/issues/944#issuecomment-3407152276

We should still be able to detect bond types for some topologies

spyke7:
> I had forgotten about the details discussed in this thread: [#944 (comment)](https://github.com/BradyAJohnston/MolecularNodes/issues/944#issuecomment-3407152276)
> 
> We should still be able to detect bond types for some topologies

Sorry for late reply, 
I have looked into it, and for topology_format="ITP", we can easily segregate btw elastic bonds (i.e having bond.type = 6) without the need for threshold value

The thing is still tricky for TPR. For now, I will implmene the ITP part only, and check for different data files, and keep the threshold as it is (ignore it for ITP).
Btw then the dna strand file will not have any elastic bonds as all are 1 in the itp file. 
And can you tell me more about the user input part, where we take the threshold value from user? Or we don't take input and keep the value None, the user can tweak it inside Geometric Nodes ?

Thanks

spyke7:
From this [comment](https://github.com/BradyAJohnston/MolecularNodes/issues/944#issuecomment-3407152276)

@BradyAJohnston can you please provide me the exact .itp file and the trajectory file which has 6 inside it, so that I can check it once

spyke7:
The changes are - 
- checks for `atoms.bonds._bondtypes` and if it has integer value then we can segregate them
- for type = 6, classified as elastic 

**Example** - 
*I could not find any specific itp file with funct 6 inside it, so I just took the dna sample and replaced some of the bonds having 1 with 6*

Before connecting delete node - 
Screenshot (264)

After connecting the delete node - 
Screenshot (265)

As you can see, from the top some bonds are not present.

Further, I'm thinking to expose the threshold to the user, which can be edited inside the geometry nodes only, so that it not needs to be set during import. What are your thoughts on it @BradyAJohnston

BradyAJohnston:
Values from `0` up until `7` already have values assigned from `biotite`: 

https://github.com/BradyAJohnston/MolecularNodes/blob/fe537e5b126c5cca94855914b0cc1c952daa7706/molecularnodes/entities/utilities.py#L175-L185

Maybe instead we just start from `10` for values from topology files that we can't immediately map back to the biotite values. That way at least `16` is similar to `6` but won't be confused with the `6` meaning aromatic double.

spyke7:
> Maybe instead we just start from `10` for values from topology files that we can't immediately map back to the biotite values. That way at least `16` is similar to `6` but won't be confused with the `6` meaning aromatic double.

Ok
So all the elastic bonds should be then classified with 16 instead of 8 right? As in the compute edge elastic I am just checking for 6 and then setting it up to 8
(But still I have to check for 6 inside the  function)

Edit: - Ig having proper docstring and docs should remove the confusion

spyke7:
Hello 👋
Please once check the things, and should I expose the threshold value as a Geometric node?

If you have time, then you can suggest me regarding any more updates or not.
Thank you

PR #1111, opened 2026-03-01
Add AI Policy

BradyAJohnston:
Adds an AI_POLICY.md

This is mostly a reposting of the [MDAnalysis AI Policy](https://github.com/MDAnalysis/mdanalysis/blob/develop/AI_POLICY.md) with some select changes to reflect that MN is not a full organisation and currently just a single repo.

### DIFF
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -1,6 +1,6 @@
 # 🤖 MolecularNodes AI-generated contributions policy 🤖
 
-* Version: **1.1 (2026-02-17)**
+* Version: **1.0 (2026-03-02)**
 * License: **[CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/)** (Copied as a template from [MDAnalysis AI Policy](https://github.com/MDAnalysis/mdanalysis/blob/develop/AI_POLICY.md))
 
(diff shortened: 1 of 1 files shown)


AI_POLICY.md
```diff
@@ -0,0 +1,83 @@
+# 🤖 MDAnalysis AI-generated contributions policy 🤖
```
IAlibay:
```suggestion
# 🤖 MolecularNodes AI-generated contributions policy 🤖
```
?

AI_POLICY.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Version: **1.1 (2026-02-17)**
```
IAlibay: Do you want to reset the version number? Not sure what the right way to do it so keeping it as-is in sync with MDA would probably be ok too!

AI_POLICY.md
BradyAJohnston: I changed most of them, miss the title haha

AI_POLICY.md
BradyAJohnston: Will reset as it might deviate from MDA going forward.

PR #1116, opened 2026-03-19
Add SMILES import support via MDAnalysis

fallofpheonix:
## Summary

This PR adds direct SMILES import support to Molecular Nodes through `MDAnalysis.Universe.from_smiles`.

The implementation uses the existing trajectory/session pipeline instead of introducing a separate RDKit-to-Biotite conversion path.

## Implementation

- Adds Blender operator `mn.import_smiles`
- Adds SMILES UI inputs in the import panel
- Calls `MDAnalysis.Universe.from_smiles(smiles)` to construct the molecule
- Imports the result through `get_session().add_trajectory(...)`
- Supports the existing style selection flow
- Makes the imported object active when one is created

## Error Handling

- Fails cleanly when the installed MDAnalysis build does not provide `Universe.from_smiles`
- Rejects empty SMILES input
- Surfaces SMILES parse failures from MDAnalysis to the user
- Falls back to `SMILES` when no object name is provided

## Compatibility

- No new RDKit dependency is introduced by this PR
- Reuses the existing Molecular Nodes session/object creation path
- Keeps the addon functional on environments without SMILES-capable MDAnalysis by reporting a clear operator error

## Testing

Automated coverage includes:

- successful import path into `session.add_trajectory`
- missing `Universe.from_smiles` support
- empty SMILES input
- default name fallback
- parse failure propagation

## Notes

A follow-up test fix was included to align the negative-path assertions with Blender operator behavior: error-reporting operators raise `RuntimeError` through `bpy.ops`, rather than returning `{"CANCELLED"}` directly to the test.

### DIFF
--- a/molecularnodes/ui/ops.py
+++ b/molecularnodes/ui/ops.py
@@ -807,6 +807,60 @@ class MN_OT_Import_Trajectory(bpy.types.Operator):
         return {"FINISHED"}
 
 
+class MN_OT_Import_SMILES(bpy.types.Operator):
+    bl_idname = "mn.import_smiles"
+    bl_label = "Import SMILES"
+    bl_description = "Import a molecule from a SMILES string"
+    bl_options = {"REGISTER", "UNDO"}
+
+    smiles: StringProperty(  # type: ignore
+        name="SMILES",
+        description="SMILES string to import",
+        default="",
+    )
+    name: StringProperty(  # type: ignore
+        name="Name",
+        description="Name of the molecule on import",
+        default="SMILES",
+        maxlen=0,
+    )
+    style: EnumProperty(  # type: ignore
+        name="Style",
+        description="Default style for importing",
+        items=STYLE_ITEMS,
+        default="spheres",
+    )
+    setup_nodes: BoolProperty(  # type: ignore
+        name="Setup Nodes",
+        description="Add nodes to the scene to load the structure",
+        default=True,
+    )
+
+    def execute(self, context):
+        if not hasattr(mda.Universe, "from_smiles"):
+            self.report(
+                {"ERROR"},
+                "MDAnalysis does not support SMILES in this version "
+                "(missing Universe.from_smiles).",
+            )
+            return {"CANCELLED"}
+
+        smiles = self.smiles.strip()
+        if not smiles:
+            self.report({"ERROR"}, "SMILES string is empty")
+            return {"CANCELLED"}
+
+        universe = mda.Universe.from_smiles(smiles)
+        style = self.style if self.setup_nodes else None
+        traj = get_session().add_trajectory(universe, name=self.name, style=style)
+
+        if hasattr(traj, "object") and traj.object is not None:
+            context.view_layer.objects.active = traj.object
+
+        self.report({"INFO"}, f"Imported SMILES '{self.name}'")
+        return {"FINISHED"}
+
+
 class MN_OT_Import_OxDNA_Trajectory(TrajectoryImportOperator):
     """
     Blender operator for importing oxDNA trajectories.
@@ -1173,6 +1227,7 @@ CLASSES = [
     MN_OT_Import_Fetch,
     MN_OT_Import_OxDNA_Trajectory,
     MN_OT_Import_Trajectory,
+    MN_OT_Import_SMILES,
     MN_OT_Reload_Trajectory,
     MN_OT_Import_Map,
     MN_OT_Import_Star_File,
--- a/molecularnodes/ui/ops.py
+++ b/molecularnodes/ui/ops.py
@@ -35,6 +35,10 @@ from .props import SURFACE_STYLE_ITEMS
 from .style import STYLE_ITEMS
 
 
+# Allow monkeypatching in tests.
+hasattr = hasattr
+
+
 def _add_node(node_name, context, show_options=False, material="default"):
     """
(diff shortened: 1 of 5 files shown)


molecularnodes/entities/molecule/base.py
```diff
@@ -138,6 +138,135 @@ def load(
...
+        if bonds:
+            bond_list.add_bond_list(np.array(bonds, dtype=int))
+
+        array.bonds = bond_list
+
+        # Standardize annotations (mass, vdw, etc.)
+        array = ReaderBase.set_standard_annotations(array)
+
+        mol = cls(array)
+        mol.create_object(name=name)
```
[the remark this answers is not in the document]
fallofpheonix: Good catch. The SMILES import path was bypassing the reader-backed metadata setup, so chain_ids and entity_ids were not being populated on the Blender object.

I’ve updated the implementation to populate these object-level properties for SMILES imports, aligning the behavior with the standard load() pipeline so chain/entity-based menus and workflows are available.

fallofpheonix:
[CI / Test with bpy (macos-latest) (pull_request)](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/23318674059/job/67824829700?pr=1116)
CI / Test with bpy (macos-latest) (pull_request)Failing after 45s
[CI / Test with bpy (ubuntu-latest) (pull_request)](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/23318674059/job/67824829692?pr=1116)
CI / Test with bpy (ubuntu-latest) (pull_request)Failing after 21s
[CI / Test with bpy (windows-latest) (pull_request)](https://github.com/BradyAJohnston/MolecularNodes/actions/runs/23318674059/job/67824829708?pr=1116)
CI / Test with bpy (windows-latest) (pull_request)Failing after 33s

CI is failing due to docs dependency requiring bpy>=5.2.0a0, which appears unavailable (404 from dailybpy).

Since CI installs all dependency groups via `uv sync --all-extras`, this affects all test jobs.

Should we pin docs bpy to a released version (e.g. >=5.1.x) or exclude docs dependencies from CI?

BradyAJohnston:
I have fixed the issues with the CI - it is now some of your tests that are not passing.

fallofpheonix:
I'll look into it and update you

molecularnodes/ui/ops.py
```diff
@@ -807,6 +810,65 @@ def execute(self, context):
...
+        try:
+            universe = mda.Universe.from_smiles(smiles)
+        except Exception as e:
+            self.report({"ERROR"}, f"Failed to parse SMILES: {e}")
+            return {"CANCELLED"}
+        style = self.style if self.setup_nodes else None
+        traj = get_session().add_trajectory(universe, name=name, style=style)
+
+        if hasattr(traj, "object") and traj.object is not None:
+            context.view_layer.objects.active = traj.object
```
BradyAJohnston, line 852: We should instead implement it as `Trajectory.from_smiles()` similar to `Trajectory.load()`. We don't have to manually add it to the session then. Look at the other MDA import operators.

molecularnodes/ui/ops.py
```diff
@@ -807,6 +810,65 @@ def execute(self, context):
...
+    )
+
+    def execute(self, context):
+        if not hasattr(mda.Universe, "from_smiles"):
+            self.report(
+                {"ERROR"},
+                "MDAnalysis does not support SMILES in this version "
+                "(missing Universe.from_smiles).",
+            )
+            return {"CANCELLED"}
```
BradyAJohnston: We are specific with the support versions of MDAnalysis so this won't be an issue. The only time there will be an issue is if using as a python package and they don't have RDKit installed.

molecularnodes/ui/ops.py
```diff
@@ -34,6 +34,9 @@
 from .props import SURFACE_STYLE_ITEMS
 from .style import STYLE_ITEMS
 
+# Allow monkeypatching in tests.
+hasattr = hasattr
```
BradyAJohnston: We don't want to do monkeypatching in tests.

tests/test_smiles.py
```diff
@@ -0,0 +1,114 @@
...
+    assert calls["name"] == "SMILES"
+
+
+def test_import_smiles_parse_failure(monkeypatch):
+    def mock_fail(_smiles):
+        raise ValueError("invalid smiles")
+
+    monkeypatch.setattr(mda.Universe, "from_smiles", mock_fail, raising=False)
+    with pytest.raises(RuntimeError, match="Failed to parse SMILES: invalid smiles"):
+        bpy.ops.mn.import_smiles(smiles="bad", name="Broken")
```
BradyAJohnston: We don't need to mock for our test - nowhere else do we use monkeypatch or dummy versions of the objects. Test the actual operations and get back the _actual_ results and assert on those.

BradyAJohnston (changes requested):
The tests need to be real tests and the import method should from a `Trajectory.from_smiles()` classmethod.

fallofpheonix:
I've updated the branch to align with that feedback: the SMILES path now goes through `Trajectory.from_smiles()` instead of constructing a `Universe` directly in the operator, the monkeypatch/dummy-object tests were removed in favor of real operator/API execution, and the temporary MDAnalysis version-check path was dropped. I'm keeping an eye on the current CI run for this head and will follow up if anything else fails.

PR #1118, opened 2026-03-20
Fix multi-file trajectory path handling during save/load

Aditya-ha11:
Fixes issue where only a single trajectory file path was handled during save/load relativization.

### Changes
- Added support for multiple trajectory files using `.filenames`
- Applied path conversion to all trajectory files instead of just one
- Maintained backward compatibility for single-file trajectories
- Ensured correct handling when reloading via `load_new`

### Notes
- No change to GUI behavior
- Improves API completeness for multi-file trajectories

### DIFF
--- a/molecularnodes/session.py
+++ b/molecularnodes/session.py
@@ -26,26 +26,48 @@ def trim(dictionary: dict):
 
 def _make_trajectory_paths_relative(trajectories: Dict[str, Trajectory]) -> None:
     for key, traj in trajectories.items():
+        trajectory = traj.universe.trajectory
+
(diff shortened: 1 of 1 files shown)


Aditya-ha11:
The failing CI checks appear to be due to an external dependency issue with `bpy` (404 during download), rather than changes in this PR. The error occurs during installation and is consistent across macOS, Ubuntu, and Windows.

Happy to help investigate further if needed.

BradyAJohnston:
You need tests to check if you are actually fixing what you claim to be fixing.

PR #1125, opened 2026-04-04
Fix: Use world-scale box dimensions when PBC correction is applied to averaged frames.

Golui:
This PR addresses a problem with PBC correction being applied to trajectory frames that were being averaged (specifically, the problem is not present for interpolated frames).

The box dimensions were not scaled by the `world_scale`, making them too large by a factor of 100. This means that the PBC correction was pretty much never applied.

Fixes #1124


BradyAJohnston (approved):

Golui:
Thanks for the quick approval and merge!

BradyAJohnston:
Thanks for finding and providing the fix!

PR #1146, opened 2026-07-21
Small molecule CAS (amongst others) importer

mancheskiL:
Built out small molecule import pipeline. Originally meant to use CAS numbers, but as the pipeline ultimately makes a API requests, and the API used actually accepts CAS numbers, SMILES, common names, scientific names, and more, this addition organically became a bit more versatile.

Additions
-----------
CAS specific Scene panel (realizing this "CAS" annotation should probably be adjusted as I write this...)
New .download pipeline (can actually probably be refactored into the original .download pipeline)
New .cas_fetch pipeline (broke out from the original fetch more so for simplicity)
New CAS_Import operator
Some new props to setup the operator and styling (in my opinion, it seems like a ball and stick default better showcases small molecules as opposed to the space filling sphere, but this is subjective of course.

FINAL NOTE: As I'm literally writing this PR, I'm realizing the "CAS" naming convention should probably be adjusted to better match the fact that this new importer can accept more than just CAS. And, there are almost certainly some methods that can be refactored or merged with previously existing functions. At this point, I chose to leave everything as I built it, if not at least so it's clear how the pipeline works, and then I can later refactor things back down so it's a little more streamlined if desirable (most obvious to me at this point is the .download function.

### DIFF
--- a/molecularnodes/blender_manifest.toml
+++ b/molecularnodes/blender_manifest.toml
@@ -24,7 +24,6 @@ copyright = [
 ]
 
 wheels = [
-	"./wheels/asttokens-3.0.1-py3-none-any.whl",
 	"./wheels/biotite-1.6.0-cp313-cp313-macosx_11_0_arm64.whl",
 	"./wheels/biotite-1.6.0-cp313-cp313-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl",
 	"./wheels/biotite-1.6.0-cp313-cp313-win_amd64.whl",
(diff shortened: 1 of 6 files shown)


BradyAJohnston:
I think this is initially looking pretty reasonable! Deciding on naming is always the hardest part - and I have no easy suggestions initially. Could maybe just be "small molecule" and then we have sub-options for either SMILES or from CAS number etc. 

I think I'll want to land the changes in #1144 which will change a bit of how this is implemented, but feel free to continue iterating on this for now.

Overall the idea seems good!

mancheskiL:
Great! I'm glad it's looking like a strong start at least. 

Naming - I think something more along the lines of "small molecule importer" and it's derivatives is pretty reasonable considering how it's evolved a bit. 

Sub-piplines - I can definitely see the idea for 2 sub-pipelines (SMILES deriving from MDA and then the database API method). I can look more into this.

Like I said, I do think the .download methods can probably be merged now that I've got something working. I'll see how the MDA SMILES pipeline goes, and then determine any other refactors after.

BradyAJohnston:
I've now landed #1144 which introduces some conflicts for this branch. 

The majority of things that have changes are the that both `Molecule` and `Trajectory` have been unified under `Molecule` (shouldn't actually affect you, can actually just help to directly use the Universe.from_smiles). 

The major change will be in the UI. I'm currently experimenting with removing the majority of the panel, and it is instead just a popup operator with the required settings. So for this PR you can remove the scene settings and the panel stuff, just exposing the options on the operator when it is invoked.

mancheskiL:
Understood! I've been on a bit of hiatus on this, but have been meaning to get back to it, largely with focus on the FromSmiles stuff. I'm happy to hear about the unification, I did find the Molecule and Trajectory stuff a bit hard to wrap my head around.

I'll check it out and make adjustments in the near future!

BradyAJohnston:
The new updated design and workflow is still in the testing phase so don't hesitate to tag me for questions around stuff as well!

Issue #1147, opened 2026-07-21
Add support for particle metadata from CryoSPARC


PlethoraChutney:
I'd like to add support for importing particle metadata from CryoSPARC .cs files. I'm happy to take on long-term maintenance of this feature. Assuming you're amenable, I thought we could coordinate the approach here.

# What to add

## Minimum viable product

I think the lightest thing that would be useful is the ability to read in particles' poses, micrograph positions, and UIDs from a `.cs` file. These are really just [numpy structured arrays](https://numpy.org/doc/stable/user/basics.rec.html). CryoSPARC stores particle poses as axis-angle rotations, but these could be changed to whatever convention we prefer using [`scipy.spatial.transform.Rotation`](https://docs.scipy.org/doc//scipy/reference/generated/scipy.spatial.transform.Rotation.html#scipy.spatial.transform.Rotation). Both of these are already transitive dependencies from MDAnalysis (which we may want to make explicit).

## Nice to have

I think it would also be nice to read in classification information (e.g., class index, per-class pose, class probability) and variability information (e.g., latent coordinates from 3DVA and 3D Flex). Not sure how viable it is for this feature to just target Blender v5.2+ so that we can use lists, there are certainly other ways to do it.

These are just additional columns from the dataset, but it would mean that at least temporarily there is more information available from a CryoSPARC metadata import than others. Which brings us to:

# Approach

Molecular Nodes can already import STAR files from either RELION or cisTEM. Currently STAR and `.cs` files are the only cryo-EM metadata formats I'm aware of, but I'm not up to date on what the cryo-ET folks are up to. So we could just do the same thing and have a third ensemble class (CryoSPARC) and be done with it.

However, it might be nice to keep the existing STAR loading machinery but have it provide the metadata to a generic "metadata" ensemble class, then write the CryoSPARC code to use the same metadata class. This way all the other MN / Blender machinery only ever has to interface with one thing. More work up front but easier to maintain and coordinate in the long run IMO.

Let me know what you think!

BradyAJohnston:
Happy to have support for this - I'm unfamiliar with the metadata that is associated with it. Happy to have you head up efforts for better support & import of relevant data. 

Do you have some examples of what the data would look like? Would it a be a single 'table' of data? We can import a "data object" that we then source the data from (can't set lists as attributes via the API yet), but we currently create a collection of data objects to import multiple conformations of a structure.

[...]

PR #1148, opened 2026-07-28
WIP: Add CryoSPARC Support

PlethoraChutney:
This PR is working toward #1147

This is a WIP for the particle metadata data object (holding off on building nodes until #1144 is done). You should be able to load `tests/data/cryosparc/J123_particles_exported.cs` and see how I'm doing things. I had a lot of fun learning more python typing for this!

A few notes:

 - I changed `entities.ensemble.base.Ensemble.create_object()` to returning `bpy.types.Object | databpy.BlenderObject`. I don't see anywhere that explicitly expects/requires `Object` but wanted to flag it.
 - CryoSPARC assigns 21 digit UIDs to everything. This is somewhat inconvenient because that's too big for `np.int32`. For now I coerce them to `int32` and warn the user if there are collisions.
 - All of the supported fields are kept with the same data type as they are in CryoSPARC, with the exception of UIDs, above; rotations (`alignments{2,3}D/pose`), which are converted to quaternions; and defocus values, which are converted to microns
 - I'm currently not applying the world scale to the particle X/Y coordinates, since they are relative to the micrograph shape (i.e., reals in [0,1]), so scaling them doesn't make sense

And a few questions:

- How would you feel adding [CryoSPARC Tools](https://tools.cryosparc.com/intro.html) as a dependency? If we do this, users can load datasets, images, and volumes directly from their CryoSPARC instance all without leaving Blender. As-is, users will have to export and download `.cs` files.
- A bit off topic: would it be possible to support `String` attributes in databpy? Happy to submit a PR/work on it over there if so.

Lemme know what you think!


BradyAJohnston:
> How would you feel adding [CryoSPARC Tools](https://tools.cryosparc.com/intro.html) as a dependency?

If it doesn't balloon the final `.zip` too much I think it would be OK to add. The extensions platform frustratingly doesn't let us download packages afterwards otherwise we could have optional additional modules that people could install. I think this would be useful enough though that it's worth trying to add.

> would it be possible to support String attributes in databpy?

We are unfortunately _still waiting_ on string attribute support in GN: https://projects.blender.org/blender/blender/pulls/154618

We can do a work around with an enum / category of string where we store integers and create a custom node for lookups for selections etc (which is what I do with MN currently), but it would be impossible to actually do string attributes properly otherwise.

---

Otherwise with a brief look I think it's looking good. I should hopefully have the #1144 landed in the next couple of days which will change a couple of things for how you have structured it, but overall should remain mostly the same.

PlethoraChutney:
> If it doesn't balloon the final .zip too much I think it would be OK to add.

I'll take a look! Didn't realize we couldn't install things post-hoc anymore.

> We are unfortunately still waiting on string attribute support in GN

Ah man I thought I remembered there being strings but just remembered that when I needed a string I'd actually just store an index into a massive string I copy-pasted into GN myself lol.

BradyAJohnston:
We _can_ install things post-hoc, but to be allowed to download install from the default extensions platform we have to abide by their rules and not download any required files after the fact. 

More and more I am thinking of just hosting our own extension platform that plays nicer with scientific workflows, but for now this is the approach.

PlethoraChutney:
Adding `cryosparc-tools` brings the `.zip` size from (160, 123, 132) MB (linux, mac, windows) to 165, 126, 136. It also means we *could* drop `mrcfile`, since cryosparc-tools ships with its own MRC handling code, but doing so doesn't reduce the `.zip` size appreciably.

If that size increase is reasonable to you I'll go ahead and implement the `cryosparc-tools` stuff I'm imagining!

BradyAJohnston:
The plan was to instead use [`GridDataFormats`](https://github.com/MDAnalysis/GridDataFormats) over `mrcfile` (which just uses it internally) for more general data handling related so simulations and also densities, but I'm happy to ship both and use them for their specific use-cases. It's already a massive `.zip` so increasing by a few mb isn't a problem. I'm not paying for the bandwidth so it's not my issue haha.

BradyAJohnston:
I've now landed #1144 - the biggest things that affect this PR is the unifying of the import operators that sit behind a single button & popup (instead of a persistent panel), and the `world_scale` increasing `10x` so that `1 nm == 1 m` whereas previously `1 nm == 100 mm`. Everything is just "10x bigger", but this makes _so much_ math easier to handle around the place. 

I've gone in and merged / updated things I think getting everything. Hopefully I haven't messed anything up too much!

The biggest thing that I am excited about, is that we node control geometry / shader / compositor nodes using [`nodebpy`](https://bradyajohnston.github.io/nodebpy/) so we can actually script the creation of nodes but that doesn't affect this PR directly yet.

PlethoraChutney:
Sick! And congrats on merging #1144, looks like a lot of work haha.

BradyAJohnston:
It has been a hellish couple of months working on the refactor - the original refactor stopped so I could spend 6 months building `nodebpy` but now I finally (mostly) finished that I could do this and I'm sure there is a bunch of stuff that has been missed but the majority of it is _FINALLYYY_ done it feels great.

BradyAJohnston:
(I will also stop pushing to your branch - sorry!)

PlethoraChutney:
haha no worries at all

Issue #1131, opened 2026-05-22
oxDNA not loading properly in 4.5.12


Krystof-Knapp:
**Describe the bug**
The import of oxDNA files seems to be corrupted. Base representations are not correctly oriented.

**To Reproduce**
Import oxDNA through the MolecularNodes importer

**Screenshots**
In Blender 4.2, MolecularNodes 4.2.5
<img width="885" height="615" alt="Image" src="https://github.com/user-attachments/assets/cf99c9aa-e715-49af-8fcc-616a10c7efd2" />

In Blender 5.1, MolecularNodes 4.5.12
<img width="1032" height="406" alt="Image" src="https://github.com/user-attachments/assets/2361b2d0-7015-4471-8c76-618ca0780092" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Hardware: x86 PC
 - Blender Version: 5.1
- MolecularNodes Version: 4.5.12

**Additional context**

BradyAJohnston:
Thanks for the bug report @Krystof-Knapp! I can confirm I can reproduce the issue. Will look into it!

If you've got any other style-related requests around oxDNA as well please do let me know. I'm not a domain expert so I welcome requests related to it.

BradyAJohnston:
I've found the issue. Currently the rotation / position for the bases is updated whenever the timestep changes, but not when first created. If you change the frame in Blender a single time it should update - but I should have an easy fix for this to ensure it's created initially as well.

Krystof-Knapp:
Hi, thank you for your swift answer! Indeed, changing the timestep fixes the orientation. That is an easy fix :). I wasn't aware that it could affect that. 
However, I found out that it doesn't fix the orientation every time. In my workflow, I usually convert the model to mesh later through ("realize instances" in geometry node editor and then "object" "convert to mesh"). Interestingly, if I have multiple oxDNA molecules in the project, and some of them were already converted to meshes (one is enough), changing the timestep doesn't fix the orientation anymore. It is not a major issue, as one can, of course, import objects from other projects, but I found it interesting. 

I am pretty happy with the current styles for oxDNA in MolecularNodes, but I will definitely let you know if I get some improvement ideas. Thanks again for your time.

BradyAJohnston:
Thanks for the extra details. I should have a fix for the import uploaded soon.

BradyAJohnston:
Became an issue again; with the 5.2 refactor I think I undid the changes

[...]

Issue #672, opened 2024-11-26
Trajectory averaging of values


BradyAJohnston:
Now we have position averaging available for trajectories (#671), values that are sample with each frame (b_factor, energy etc) could also be sampled and averaged. Currently don't have in place per-frame sampling, so that would need to be setup first.

PR #1189, opened 2026-09-07
Fix oxDNA Import and Style Node

dhollis6:
Fixes issues in the oxDNA import, re-using the frame handling from the base `Molecule`. Also adds functionality for subfame and averaging of other attributes to support base rotations.


BradyAJohnston:
Hey there! Happy to help get this fixed up and passing to improve the oxDNA import as I'm certianly no expert in the domain. 

Are you able to give some more explanations about what is being changed and why? The `set_frame()` I am familiar with from a previous issue #1131

dhollis6:
Hey @BradyAJohnston ! Thank you, and sorry I'm a bit disorganized. This is my first git pull request and I'm learning as I go, so any feedback is welcome. I'll explain the changes here and you can let me know your thoughts.

All changes are to compensate for the fundamental difference between the oxdna model and atomistic models. As you know, each oxdna "atom" (really several dozen atoms, coarsely modelled) is defined by both a position coordinate and a pair of mutable rotation vectors. So these vectors must be handled like position coordinates for the purpose of visualization.

1. Add _create_object() inside OXDNA to override the parent class function. This allows us to update the oxdna vectors upon instantiation of the Blender object. Otherwise the parent only updates the positions. Like you said it was present earlier and I'm not sure when it was removed.

2. Give OXDNA objects their own instance of FrameManager. Create caching and updating functions for oxdna vectors and velocities (analogous to those for positions) inside FrameManager. Recycle the PositionCache class, which can be used for any vector values, to maintain these caches. This ensures that when handlers.py>update_entities() calls set_frame(), the oxdna attributes are updated with all processing applied (interpolation, averaging). The MDAnalysis Timestep object is now read in FrameManager instead of OXDNA to keep a consistent pattern of input/output between classes.

3. Use different iteration variables in nested for loops in _read_topo_new in OXDNAParser. Before, they were the same, and since iteration variables aren't restricted to the scope of the loop all chain IDs were being set to 0. Added a check for is_circular tag, since it's the only way to tell if a strand's ends are ligated in the new topology format. I notice github is throwing some errors related to topology comparisons/tests....maybe this is the culprit? Old topology reader works great.

I've attached some new oxdna nodes here that you might find valuable. A few months back you mentioned it was difficult to add new nodes...not sure if this is still the case. If you give me a hint I might be able to add them myself.

- oxDNA Style Ball and Stick. Generates the standard oxDNA model seen in Thomas Ouldridge's original thesis and most papers since. Defaults to base-agnostic color scheme seen in many papers. Lots of built-in customization options. Some could be removed. I think this could make a good default style.

- oxDNA Realize Vectors & oxDNA Recover Vectors. These nodes generate vectors as physical geometry, and read that geometry back as stored attributes, respectively. These are necessary when animating oxdna using armatures, since armatures only act on geometry and there's no node support for them yet. Deform the coordinates, render the geometry afterwards... no shearing or stretching!! Already works great for atomistic models.

No hurry to respond--the hours I can spend on this are few and far between, anyways. Thanks for making such an awesome program. I told my friend, this is the future!

[Short node demo](https://youtu.be/A7yjxcAZSG4?si=j3zU5bjs2U4n6EMd)
Attached oxdna files have different sequences but otherwise same topology.

Short node demo:
[oxdna_custom_nodes_1_MN_v520.blend.txt](https://github.com/user-attachments/files/31906171/oxdna_custom_nodes_1_MN_v520.blend.txt)
[fully_ligated_weird.dat.txt](https://github.com/user-attachments/files/31906374/fully_ligated_weird.dat.txt)
[fully_ligated_weird_new.top.txt](https://github.com/user-attachments/files/31906375/fully_ligated_weird_new.top.txt)
[fully_ligated_weird_old.top.txt](https://github.com/user-attachments/files/31906376/fully_ligated_weird_old.top.txt)

(Edit: inline video thumbnail)

BradyAJohnston:
Thanks for the detailed write-up and being interested in contributing!

From an initial look, it all seems pretty good and reasonable.

## Averaging of Attributes

The averaging of attributes is something that could also be applied more generally to the `Molecule` class as well, as `b_factor` is a per-trajectory-frame attribute that we could potentially be averaging. I have an initial issue raised here: #672

That could be a separate PR completely but happy to keep things as-is for this PR for now.

## Tests

You can run the tests to update the old snapshots using `uv run pytest --snapshot-update`. The example data that you attached, you could add that under `tests/data/oxdna` for more example datasets that we could use, and create some new tests for the them to ensure we are testing against the `is_circular` etc. 

After adding the tests, if you use snapshots again you can create them with `uv run pytest --snapshot-update`.

## New Nodes

I'm _very close_ to being able to accept node contributions in the form of code instead of `.blend` files via [`nodebpy`](https://github.com/BradyAJohnston/nodebpy), but it's still not quite there yet. 

Once you have tests updated and working I can manually go in and merge the new style nodes. The 'classic' style like you mentioned is one that I have wanted to get around to making for a long time, so thanks for doing that!

BradyAJohnston:
Additionally if you wanted to contribute some written documentation for using with `oxDNA` that would be great (but not a requirement) as well! I'm not a domain expert so I haven't written any yet - but they would go `docs/tutorials/oxdna.qmd` with some written & screenshots etc. If you wanted to!

dhollis6:
Copy that, I'll try to get testing done this week! I don't think I qualify as a domain expert--just picked up a few things in undergrad some years ago--but I am interested in writing some documentation. I can't guarantee it'll be anytime soon, though. I'll keep you updated!

BradyAJohnston:
I haven't really used oxDNA short of creating the initial parser, so any more experience with it than me makes you an expert in comparison! Docs could be separate to this PR, as we could get the fix implemented first and then docs later on.

dhollis6:
OK, posting this preemptively since it looks like tests should succeed. Here are the changes since my last post:

1. Fix strand directions in oxdna test files and update tests accordingly. Topology and configuration files alike can be written in either 3'-->5' direction (old format) or 5'-->3' (new format). The test data contained a trajectory which was written in the new format but assumed to be format-agnostic, and this previously caused the tests to fail on this branch. Unfortunately, the format of configuration (trajectory) files cannot be easily determined without using analysis scripts or visualizing the strand directions in a program like oxView (or now, MN!)
2. Add ligated minicircle test data and tests to verify reading of fully ligated strands.
3. Move Cando2oxdna test files into oxdna test folder. Method of file creation should not matter.
4. Rename test files for clarity.
5. Define pytest fixtures using a single function which draws from a dictionary of file names. This improves readability and reduces duplication.
6. Replace most hard-coded comparisons with snapshot comparisons to increase adaptability.
7. Update test_dna snapshots.

Everything should hopefully be ready to go. Let me know if you'd like me to change anything. Otherwise I'll unfortunately have to step away from this project for now to focus on other things. But it's been really fun and I've learned a ton. I'll look forward to when you have the oxdna node(s) implemented. And I'm excited to watch your talk at the blender conference when it's posted online! Break a leg!!

BradyAJohnston:
Thanks for working on this! I'll review and make some tweaks / changes, but overall I'm happy with this and should be able to merge it soon. 

Thanks so much for the contribution!

BradyAJohnston:
I've gone through a cleaned up a bit and tweaked some things - looking great!

Apologies if you have been swamped with notifications from this branch. I've been using it as a test for the first time using `nodebpy` to generate a `.py` code version of the node groups. Eventually everything will _only_ be the `.py` code but for now I'm keeping `.blend` and `.py` side-by-side.

I've exposed the 3 nodes that you created after cleaning up some of the internals. I think the two oxdna vectors nodes I think we could potentially do more cleanly and in the background, but that's a design decision for the future - I'm happy with them shipping for now.

dhollis6:
Okay sweet! Obviously you have a better understanding of the big picture than myself so I figured you'd want to make changes. I'm just happy that the oxdna implementation will be functional at baseline. I think a lot of people will find this useful.

I was monkeying around with the oxdna style node and integrated the ribbon features from the old node along with toggles for more customization... only adds a couple sockets. This could be added later, or not at all...

[oxdna_custom_nodes_2_MN_v520.blend.txt](https://github.com/user-attachments/files/32087502/oxdna_custom_nodes_2_MN_v520.blend.txt)

oxdna_custom_nodes_2

BradyAJohnston:
I can update the style node to include a few more options. 

The tests are currently failing due to a bug in GN that I managed to discover: https://projects.blender.org/blender/blender/issues/163764

I'll have to tweak the tests to get around it for now.

dhollis6:
Okay gotcha. Also the ribbon backbone will have to use Mesh to Curve instead of Points to Curve so edges are converted to curve spline data. I corrected this in the attached file.

[oxdna_custom_nodes_2_fix_MN_v520.blend.txt](https://github.com/user-attachments/files/32091009/oxdna_custom_nodes_2_fix_MN_v520.blend.txt)

PR #1212, opened 2026-09-14
WIP: ENH: Select String node (chain, residue ID and residue name selection strings)

BradyAJohnston:
Adds a **Select String** node: type `A, B:10-20, LYS` and get a boolean selection, the string-typed counterpart of the per-entity `Select Chain …` groups and of `Select Res ID String`.

## Syntax

Comma-separated terms, ORed together:

| term | selects |
|---|---|
| `B` | chain B |
| `42`, `10-20`, `-5` | residue ID, inclusive range, negative ID |
| `LYS`, `MSE` | residue name (aliases sharing a `res_name` number, e.g. `MET/MSE`, all match) |
| `B:10-20`, `2:HIS` | residues within a chain |
| `1:` | a chain whose ID looks like a number (a bare number is always a residue ID) |
| `:A` | the residue name `A` when a chain `A` also exists |

## How the names are resolved

Chain IDs and residue names live on the geometry as integers, so the node carries the lookup tables in a collapsed **Context** panel:

- **Residue Names** ships as the node default, generated from the `res_name_num` table in `assets/data.py` (list position = number, `/` joins aliases, empty entries are unused numbers). A test asserts it stays in sync with `data.py`.
- **Chain IDs** is filled in by the node add handler (`nodes/_handlers.py`) from `obj.mn.chain_ids`, the same ordered list the `Select Chain …` groups are built from, when the node is dropped into an entity's tree. Python users pass `chain_ids=",".join(mol.props.chain_ids)`.

A `String List Index` helper group (inlined into the module by dump because only one asset uses it) turns a name into its list position, or -1.

| Node | Tree |
|---|---|
| Select String node as seen from another tree | Select String internals: split term, chain, residue ID, residue name and combine frames |
| String List Index node as seen from another tree | String List Index internals: a repeat zone matching each entry against the item |

1cd3 cartoon in grey with the selection "B, F:1-60, 2:HIS" coloured orange

## Tests

- `tests/test_nodes.py::test_select_string` checks 19 selection strings against numpy on 1cd3 (which has chains named `1`–`4` as well as letters), including the `Inverted` output.
- `tests/test_nodes.py::test_select_string_residue_names_match_data` guards the baked residue table.
- `tests/test_node_on_add.py` covers the handler for appended and linked adds, an already-filled value, and a tree with no entity.
- `uv run -m nodebpy.assets check` round-trips.
- nodebpy is bumped to 520.26.0 (same pin as #1210) for the two-image plots above.

## Found on the way

The first build selected nothing for any chain or residue name although the same source evaluated correctly in-session. The asset build splits the Group Input node per socket and re-creates its links, which reversed the order of the two links into a Join Strings multi-input, turning the search key `/B/` into `B//`. The helper now trims the item first so no group-input link lands on a multi-input. Reported with a standalone reproduction as BradyAJohnston/nodebpy#153.

Agent-authored (Claude Code) following the `mn-nodes` skill (#1226); the skill file itself lives in #1226.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014r4R2rt1afRN7umWzaWH8L


BradyAJohnston:
Was an interesting experiment but closing as if I was to create this it would likely go in a different direction and waiting for string attributes might just be the best option

Issue #1203, opened 2026-09-11
oxDNA Import and Style Node features


dhollis6:
Thanks Brady for cleaning up the oxdna style node in #1189, this looks much cleaner and more efficient than what I had made. Here I'm just documenting some nit-picky features I'd still like to see that didn't make it into the final PR. This is low priority and would maybe be easier once `nodebpy` is fully implemented.

1. `Ball radius` socket should not affect backbone radius. When `Curve` menu option is selected, `Ball radius` should be hidden.
2. Base positions should be set to the endpoints of the base stems.
3. Add `Stem Geometry` socket to enable, for example, visualizing bases as flat rectangles.
4. `Strand` color sockets appear to override `Color` attribute for only one strand (in the case of a linear duplex). They should override `Color` for all strands.
5. Backbone geometry (curves and sticks alike) should have filled faces.
6. Backbone geometry should extend slightly past base stems at terminated ends, so base stem doesn't appear floating/partially detached. This should depend on `Stem scale`.
7. `Ribbon` node label is maybe misleading. Something like `Classic` or simply `oxDNA Style` might be better.

~~And an unrelated bug:~~ (Moved to #1206)

~~8. `Average` input field in right-hand trajectory settings panel should cap out at 4. Setting to 5 freezes the trajectory.~~

(Edit: clarity)

(Edit: separate bug into its own issue)

dhollis6:
I'm happy to make these changes myself, but I'll probably wait until the node editing workflow is more established. Most of these features were present in the .blend file nodes attached to the original PR, but I completely understand if changes needed to be made for efficiency or readability. This is my first time working with geometry nodes so I still have a lot to learn about best practices and methods.

BradyAJohnston:
Thanks for looking it over and adding some more suggestions. 

I have just completed the transition over to `nodebpy` (#1204) base node groups, so if you'd like to give it a try I think this would be a perfect test case for trialling accepting a PR of node group changes.


The workflow will be as follows:

## 1. Build the `nodes.blend` asset file.

```bash
uv run -m nodebpy.assets build
```

## 2. Make changees

Open the `nodes.blend` file, make changes to the node group that you would like - or if they are relatively simple just make them to the `.py` files instead. Probably best for now to just edit the `.blend`.

## 3. Dump the `.blend` back into `.py`

```bash
uv run -m nodebpy.assets dump
```

## 4. Check in the `.py` changes

[...]

PR #1213, opened 2026-09-15
ENH: oxDNA style node quality-of-life improvements and bug fixes

dhollis6:
Test the `nodebpy` PR workflow and implement the oxDNA style node enhancements proposed in issue #1203.


BradyAJohnston:
You have no idea how ridiculously happy it makes me to see this PR come about - and to see the nice clean diff that makes this possible!

I've cleaned up the line endings and added the `.gitattributes` which should fix it going forward as well.

dhollis6:
That makes me happy to hear! This is a really powerful workflow and I'm sure it has applications well beyond what I'm familiar with. Really cool.

Thanks for fixing that EOL issue. I have to step away for a couple days but I'll try to have this PR done by the end of the week.

dhollis6:
@BradyAJohnston I noticed that the oxdna style as you implemented it has several differences to the mainstream style seen elsewhere. E.g. the base stems are thinner by default, and set at a noticeably different angle relative to the base vectors... Was this intentional? I thought it'd be best to have the default style align with what people are already familiar with, but I wanted to ask and make sure before I make any changes.

BradyAJohnston:
I was doing some more general cleanup and some things changed unintentionally it seems - I defer to you about what the defaults should be!

If you could include some screenshots of the adjusted style vs from a different program as well that would be great.

dhollis6:
Here's a visual side-by-side. Left is oxView, which is I think the most common visualizer and reflects the style you see in Ouldridge's original thesis. Center is my original PR before review, and right is the current implementation. I was going to try and return it to the center style (besides the colors). Regardless, I'll stick to the techniques you used--`Curves to Mesh` is way more efficient than whatever I was doing with cone instances and repeat loops, lol.

oxdna_style_compare

dhollis6:
And I'll share screenshots of the node changes and final style once I've made more progress. Still need to fix strand direction and recreate test snapshots.

BradyAJohnston:
OK yep I see what you mean, I also managed to have the backbone in the wrong direction lmao. Looking forward to the changes

dhollis6:
A lot of changes have been made both visually and under the hood. I'll list the major ones here.

# Fixes

- The backbone offset calculation uses the cross product of the normal and base vectors. Previously these operands were flipped, causing the major and minor grooves to be mismatched. oxDNA offset calculations: https://lorenzo-rovigatti.github.io/oxDNA/configurations.html

- `Mesh to Curve` sets the start of the curve to the point with the lowest index. Therefore, using this node to generate the backbone assumes the atoms are indexed in 5'->3' order. But structures imported using the old oxdna file formats will be indexed in 3'->5' order, and the backbone direction will be flipped. So we need a way to tell the order in which the atoms are indexed. Currently, the index order is not accessible inside Blender. To work around this, I've added a node group which compares the oxdna normal vector of the current atom to the position of the next atom (index + 1). Normals generally point in the 3'->5' direction in a relaxed oxdna structure due to the persistence length. But this is an extrapolation and may not hold true for unnatural, highly strained, or very small systems. So eventually we should make the atom index order accessible. I can open an issue after this PR is reviewed.

- Disconnect modulo node from selection input socket of strand `Set Color`. Previously, the `chain_id` modulo result was fed into the selection socket, meaning that only strands with `chain_id mod 4 > 0` would be colored.





# Features (subjective & open to change)

- Node renamed to `Classic` and input labels renamed for clarity

- `Ball Radius` and `Backbone radius` are now decoupled.

- Make stem geometry point toward base center (i.e. stacking interaction site). Previously, the stems were rotated 20 degrees out from the base vectors such that they pointed toward the helix center. While this is technically correct, it meant the stems did not pass through the center of the bases, which are rendered at the stacking sites and directly along the base vectors. It's a matter of taste, but this change creates symmetry and adds consistency with other oxdna visualizers.

- Bases are locked to ends of stems, so changing stem length moves bases. By default, bases are located at the stacking interaction site.

- Add `Stem Geometry` input with similar handling to `Base Geometry`.

- When using `Ribbon` backbone shape, make ribbon extend past final nucleotide in strand by `End Overhang`.

- Change default geometry sizes to better align with mainstream oxDNA visualizations.

- Change all input units to Angstroms for consistency and expected behavior.

- Update render test and recreate snapshots.




This PR implements all of the features mentioned in #1203.

Edit: clarity

dhollis6:
Cross-checked with oxView model. Ligated and nicked minicircles are accurately represented too.

I currently have the default `Taper` value set to 0, since I never found the taper to be very useful in the past, and I figured MN will mostly be used for pretty renders rather than detail-oriented stuff. It also helps to distinguish it a bit from oxView. Just my two cents.

oxview_compare

dhollis6:
Increased default geometry scales to further distinguish from oxView. Now it's closer to the renders seen in Ouldridge publications for oxDNA 1 and 2. oxView is a popular software and while it's important to maintain continuity across platforms I didn't want to come across as copying the style verbatim.

molecularnodes/nodes/geometry/oxdna_style_classic.py
```diff
@@ -31,19 +33,48 @@
...
+from .edge_info import EdgeInfo
 from .fallback_geometry import FallbackGeometry
 from .integer_distance import IntegerDistance
 from .oxdna_vectors import OxDNAVectors
 from .set_color import SetColor
+from .world_to_angstrom import WorldToAngstrom
 
 
-class OxDNAStyleRibbon(AssetGeometryGroup):
+class OxDNAAreIDs53(CustomGeometryGroup):
```
BradyAJohnston, line 49: I think this could be renamed to `OxDNA Direction` if we want to be a public node, with boolean outputs `5->3` and `3->5` which are just inverse of each other. 

If you don't imagine it being useful outside of this limited context, we could rename it to an internal node group with a `.` prefix `.OxDNA Direction`. 

Given it's only used once, cleaner to not require the output have a `not()` and instead just use the `3->5` output.

molecularnodes/nodes/geometry/oxdna_style_classic.py
```diff
@@ -343,130 +395,172 @@ def _build_group(self, tree: TreeBuilder[GeometryNodeTree]) -> None:
...
+                with g.Frame("Add overhang to strand ends"):
+                    vector_math_3 = (
+                        EdgeInfo(vertex_index=g.Index()).o.edge_vector.normalize()
+                        * -1.0
+                        * AngstromToWorld(angstrom=end_overhang)
+                    )
+                    set_position_1 = g.SetPosition(
+                        geometry=set_position,
+                        selection=g.Compare.integer.equal(
+                            g.EdgesOfVertex(vertex_index=g.Index()).o.total, 1
```
BradyAJohnston, line 480: This can  just be `g.EdgesOfVertex().o.total` as the default value is already `Index`.

molecularnodes/nodes/geometry/oxdna_style_classic.py
```diff
@@ -343,130 +395,172 @@ def _build_group(self, tree: TreeBuilder[GeometryNodeTree]) -> None:
...
+                    geometry=set_position
+                ) >> g.InstanceOnPoints(
+                    instance=g.IcoSphere(subdivisions=quality),
+                    scale=AngstromToWorld(angstrom=ball_radius),
+                )
             with g.Frame("Backbone Stick"):
                 angstrom_to_world = AngstromToWorld(angstrom=backbone_radius)
+                with g.Frame("Add overhang to strand ends"):
+                    vector_math_3 = (
+                        EdgeInfo(vertex_index=g.Index()).o.edge_vector.normalize()
```
BradyAJohnston, line 473: This can just be `EdgeInfo().o.edge_vector.normalize()` as the default is already `Index`.

molecularnodes/nodes/geometry/oxdna_style_classic.py
```diff
@@ -265,18 +310,20 @@ def __init__(
                 "Backbone Shape": backbone_shape,
                 "Backbone Radius": backbone_radius,
                 "Ball Radius": ball_radius,
-                "Arrow Taper": arrow_taper,
+                "5'→3' Taper": _5_3_taper,
```
BradyAJohnston, line 313: Better to just name this `Taper` or `Backbone Taper`

BradyAJohnston: (partly because currently the auto-generated signature is `_5_3_taper` and it would be better to just have `taper`

BradyAJohnston (changes requested):
Thanks so much again for this. It's actually so nice to be able to review a PR for nodes in code form - I still can't get over how well this is working. 

Just a couple of minor tweaks that I have highlighted in comments and one additional request. 

Previously I have been pretty slack with adding descriptions to the socket inputs and outputs. Now that we can write the descriptions in the form of code it's much less hassle to add a `description="some description info etc"` to the input and output sockets. If you could add those to fill out this node it would be perfect!

EDIT: Nothing should change the actual output of the node, so these could be able to be done entirely in python without the need to build / dump as well if you wanted!

dhollis6:
OK that should be everything! Feel free to request any additional changes.

BradyAJohnston (approved):
Thanks so much again for all the work, especially for adding the extra descriptions to the sockets!

BradyAJohnston:
No pressure of course, but if you did want to write up any written documentation for oxDNA import that would be a great addition also!

PR #1226, opened 2026-09-18
Add agent skills: mn-nodes and mn-render (draft, not for merge)

BradyAJohnston:
> **Draft, not for merging yet.** Opened to make the skills visible and discuss whether and how agent skills should live in the repo.

## What this adds

A tool-agnostic top-level `skills/` directory holding two [agent skills](https://docs.anthropic.com/en/docs/claude-code/skills), each a single `SKILL.md`, plus symlinks under `.claude/skills/` so Claude Code discovers them (the same pattern the `great-docs` branch uses for its own skills). They load on demand when a task matches; nothing changes at runtime or in normal sessions.

### `skills/mn-nodes` – writing, building and testing node groups

Captures what was learned building the symmetry nodes (#1210) and Select String (#1212):

- **Mental model**: Python modules are the source of truth, `nodes.blend` is a build product, and `dump` regenerates modules from the built tree (so `#` comments inside `_build_group` are lost).
- **Pipeline**: `uv run -m nodebpy.assets build | dump | ensure | check` and `docs/generate.py`.
- **Anatomy of a node module**, catalog IDs, field inputs, units, `_shared/` helpers; in-tree comments via a Frame plus an orphan String node.
- **nodebpy idioms** and where to find the API; numeric tests with independently computed numpy expectations; golden-image tests.
- **PR workflow**: use-case renders plus node-tree plots attached with `gh pr create --attach`; dumping third-party `.blend` files to nodebpy source; pitfalls (multi-input link order, `dump` inlining single-use helpers, `Compare` wanting sockets).

### `skills/mn-render` – rendering from a Python script

Distilled from `docs/api/*` (canvas, rendering, materials, styles, trajectories, mdanalysis), `tests/test_canvas.py`, `tests/test_render_images.py` and the `scene/` source:

- **Environment**: `uv run`, `BLENDER_USER_EXTENSIONS` before `import bpy`, Canvas before entities, compositor on CPU.
- **Canvas settings** table (engines and their constructor arguments, resolution, transparency, background, colour management, passes, timeline, camera, compositor) and what `clear()` / `load_preset()` / `reset()` each keep or drop.
- **Loading** with `fetch`, `load` and `Molecule(universe)`; there is no `mn.Trajectory` any more.
- **Styles, materials, colour**: `add_style` rules (callable styles, selections, `assembly=`), preset material parameters and the shared-datablock trap of string materials, colour validation, spheres on EEVEE.
- **Framing** with `look_at`, viewpoints and margin, lens ordering, clip distance.
- **Trajectories** (subframes, interpolate, average, frame range arithmetic) and the three animation routes: `animation()`, `record()` with `FrameRecorder`, or `snapshot(frame=)`.
- **Deterministic renders and golden tests**: the fixture settings, tolerances, snapshot locations and how to update them.

## Relation to other PRs

This branch is the only copy of the skills. #1210 carried an earlier version of `mn-nodes` (a strict subset); that copy has been removed from the `symmetry-nodes` branch and its description now points here. #1212, #1224 and #1225 were authored with the untracked skill and their descriptions reference this PR. The `great-docs` branch (#1166) has its own `.claude/skills/*` symlinks into `.agents/` for the great-docs tooling, unrelated to these.

## Found while writing `mn-render`

- **Bug**: `Canvas.look_at(points_array, viewpoint=...)` fits against the stale camera basis (#1227, with a proposed one-line fix). Entity targets are unaffected.
- **Docs drift** (not fixed here): `docs/api/styles.qmd` shows an `add_style` signature without `assembly=` and `name=`; `mn.material.Transparent` and `"Flat Outline"` have no reference pages; `docs/api/canvas.qmd` uses the private `compositor._annotations_image()` in an example where `compositor.add_annotations()` is the public path; `docs/api/_density.qmd` is underscore-prefixed so its canvas walkthrough is never rendered; two typos in `docs/api/index.qmd` ("You _must_ a new canvas") and `docs/api/canvas.qmd` ("camer").

## Open questions before merging

- Should skills be committed at all, or should the content move into `CONTRIBUTING.md` / `docs/dev/` and be referenced from much thinner skills?
- Some sections are tied to one person's setup (a sibling `../nodebpy` checkout, scratchpad paths, a survey report that only exists in a session directory). Those need to be generalised or removed.
- Whether `.claude/skills` symlinks are the right discovery mechanism or each tool should get its own pointer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BradyAJohnston:
Superseded by #1245

